PR #2, opened 2015-03-25
Enable multiple target platform definition through a DSL

donat:
Defined in the BuildDefinition plug-in.


etiennestuder:
Integrated.

PR #7, opened 2015-04-01
Implement launching methods and classes from the editor

donat:
Add CMD+OPTION+X G key binding
Use property testers to determine when to show the launch shortcut


etiennestuder:
Replaced by newer PR.

PR #10, opened 2015-04-01
[463693] Invalid cast to StructuredSelection

waynebeaton:
Cast to an IStructuredSelection because the selected object's class may
not be a StructuredSelection subclass.

https://bugs.eclipse.org/bugs/show_bug.cgi?id=463693


waynebeaton:
I'll resubmit with the Signed-off-by line.

PR #17, opened 2015-04-06
Use conventionMapping for assembleTargetPlatformTask

donat:


etiennestuder:
Obsolete since we now use .target files as input.

PR #24, opened 2015-04-09
Implement changes proposed in the 'Revisit current Task View' story

donat:
Don't show project tasks by default
Move filter actions to the task view's menu
Rename 'Include private tasks' to 'Show All Tasks'
Cleanup i18n resources


etiennestuder:
Reviewed and integrated.

PR #31, opened 2015-04-16
Sign the update site with eclipse.org certificate if it is uploaded to the public repository

donat:


etiennestuder:
Extracted some properties and integrated.

PR #32, opened 2015-04-21
Cleanup package exports

donat:
Use plugins only for production dependencies.


etiennestuder:
Integrated. Please close the BugZilla issue, too, and reference the commit in the resolution comment.

PR #34, opened 2015-04-23
Optimize p2 repository

donat:
p2.index files for the update sites
update sites contain compressed artifacts too


etiennestuder:
Reviewed and integrated.

PR #35, opened 2015-04-24
Bug 465355 - Provide option to convert an Eclipse project to Gradle

vogella:
This commits adds the convert story to the Project stories.

Signed-off-by: Lars Vogel <Lars.Vogel@vogella.com>


etiennestuder:
Thanks, Lars. We will look into this PR as soon as we have a clear strategy on partial imports.

PR #39, opened 2015-04-24
Externalize strings

DenisZygann:
Externalize Strings in org.eclipse.buildship.core plugin.


etiennestuder:
Thanks for the PR. We plan to integrate it soon.

etiennestuder:
We will address full String externalization later.

PR #42, opened 2015-04-28
Change color of run configuration arguments shown in the Gradle Console

donat:
Create dedicated output stream in GradleConsole where configuration information is printed


etiennestuder:
Reviewed and integrated.

PR #43, opened 2015-04-28
Reveal console view when build starts

donat:


etiennestuder:
Reviewed, refactored and integrated.

PR #47, opened 2015-05-04
Implement project nesting for Gradle projects in the Package Explorer

donat:


donat:
Closed because of the implementation is insufficient:
- More testing required
- Missing features: the option to enable-disable the hierarchical presentation
- We can’t represent the hierarchy in the package explorer

PR #51, opened 2015-05-13
Don't recreate project descriptors if they already exist

donat:
If they exist we only add the project nature
This makes it possible to import Buildship into Buildship


etiennestuder:
Review, slightly refactored, and integrated.

PR #69, opened 2015-05-22
Execution view context menu

SimonScholz:
-add switch action to the very right
-add context menu: navigate to source file (only offer on operations for test classes (JvmTestOperationDescriptor)), separator, expand, collapse
-add double-click and pressing ENTER to navigate to source file
-use OperationDescriptor#getName in case of TestOperationDescriptor
-add action to navigate to corresponding console page from the execution page


etiennestuder:
Refactored, extended, and integrated.

PR #77, opened 2015-05-28
Add 'Show stacktrace' action to the context menu of the executions view

donat:
Add UI widgets to the StacktraceDialog
Buttons work on StacktraceDialog
Show test failure action appears in the context menu only if a failed test is selected
Externalize strings
Rename show context menu item to 'Show failures'
Rename action to ShowFailureAction
Show failure for all nodes having containing FailureResult
Copy only the stacktrace text to the clipboard
Disable copy button when there is nothing to copy
Add scrolls to the stacktrace text


etiennestuder:
Fine-tuned the UI and then integrated.

PR #95, opened 2015-06-03
Add validation decoration on the project location text of the project…

donat:
… import wizard


etiennestuder:
We will not decorate for now to be consistent with all our error reporting.

PR #100, opened 2015-06-05
Project conversion

SimonScholz:
Adds menu entry to the configure menu
Creates and manipulates gradle files for an existing project


fbricon:
Instead of hard coding java support in a StringBuffer (you should definitely use StringBuilder instead BTW), is it possible to build up a Gradle project model instead, and then have some gradle facilities serialize that model as a gradle file?

That would allow injecting conversion participants that could add plugins configuration (I'm thinking Java EE projects) or dependencies.

The m2e conversion mechanism I wrote handles that conversion scenario pretty well. https://github.com/eclipse/m2e-core/blob/ced5ba9121233c507fef1909e6a1f24ac67a5c63/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/conversion/ProjectConversionManager.java#L136

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/project/internal/GradleInitializer.java
```diff
@@ -0,0 +1,146 @@
...
+        }
+
+        return gradleInitTempDir;
+    }
+
+    private static String getBuildFileContents(List<String> gradlePlugins, List<String> repositories, List<SourceSet> sourceSets, List<String> fileSystemClassPaths)
+            throws IOException {
+        StringBuffer buffer = new StringBuffer();
+        addGradlePlugins(buffer, gradlePlugins);
+        buffer.append(System.lineSeparator());
```
fbricon: System.lineSeparator() is not part of Java 1.6. It fails to compile

fbricon:
the gradlew* files are not set as executable, so they can't be used unless one `chmod +x` them

fbricon:
For a simple maven-like structure, the conversion generates this build.gradle file : 
```groovy
apply plugin: 'java'

repositories {
jcenter()
}

sourceSets {
main {
java{
srcDir 'java'
}
}
}
```

- The java source set should be skipped, as it corresponds to the default value.
- Eclipse compiler settings (source/target levels) should definitely be locked in build.gradle.
- To a lesser degree, project encoding should be added as well.
- Indentation could be added here. This is pretty annoying since there's no Buildship formatter for gradle files.

SimonScholz:
@fbricon Thank you very much for the review. 
Of course this hard coded build file generation will be removed later on and the conversion will be moved to gradle itself.
Actually it is intended to let gradle handle the majority of work, so that we have as less Eclipse specific code as possible. Therefore I tried to pass certain non Eclipse parameters, which are of concern for a build.gradle file as first "draft".
I am currently considering to implement a first version of a Gradle model, which can be passed to the GradleInitializer class in order to configure a project according to given parameters in the Gradle model.
But having an extension point in order to enhance this gradle model, which will be passed to the GradleInitializer is a great thing, which I will add tomorrow. 
Does this make sense to you? Again thank you for your review.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/gradle/model/GradleModelFactory.java
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * This factory creates Gradle model specific objects.
+ */
+public class GradleModelFactory {
+
+	private GradleModelFactory() {
+	}
+
+	public static Dependency createDependeny(final String dependencyString) {
```
Ian-Stewart-Binks: 'Dependency' is missing a 'c' (Dependeny).

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/wizard/project/ProjectPreviewWizardPage.java
```diff
@@ -237,21 +243,10 @@ private void updateGradleVersionLabel(Label target, Property<GradleDistributionW
         }
 
         // if the length of the text is changed and the version warning is visible then we have to
-        // adjust their horizontal alignment
+        // adjust their horizontal alignemnt
```
Ian-Stewart-Binks: Typo in `alignemnt`.

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/wizard/project/GradleProjectConvertWizard.java
```diff
@@ -0,0 +1,82 @@
...
+
+        setDialogSettings(dialogSettings);
+
+        this.setController(new ProjectImportWizardController(this));
+        // instantiate the pages and pass the configuration object that serves as the data model of
+        // the wizard
+        ProjectImportConfiguration configuration = this.getController().getConfiguration();
+        this.gradleOptionsPage = new GradleOptionsWizardPage(configuration, publishedGradleVersions);
+        this.gradleOptionsPage.setTitle("Convert to Gralde");
+        this.gradleOptionsPage.setMessage("Settings for the conversion to Gralde");
```
Ian-Stewart-Binks: 'Gralde' misspelt on this line and above.

Ian-Stewart-Binks:
`settings.gradle` looks like it hasn't been updated with an `org.eclipse.buildship.jdt` entry.

vogella:
Any chance that this feature can made available? Would be handy to avoid adding the nature and build.gradle file manually?

etiennestuder:
Gradle has built-in functionality to convert projects to Gradle, e.g. Maven projects. We should leverage this functionality rather than re-implementing this functionality.

PR #113, opened 2015-06-09
Update the duration every second in the ExecutionPage

SimonScholz:
Change-Id: I01b84a4b5255a8dc0b9243a408f9bb88e34c0c4d
Signed-off-by: Simon Scholz <simon.scholz@vogella.com>


etiennestuder:
Integrated.

PR #128, opened 2015-06-30
Bug 471005 - Importing existing project results in Java Model Exception

Ian-Stewart-Binks:
When the build.gradle file of a project contains multiple source sets that specify overlapping source directories, each specified source directory is added to that projects source directories, regardless of whether or not it's a duplicate.

When the project is imported, these source directories are each added to the classpath. This happens in `GradleClasspathContainerInitializer#updateSourceFoldersInClasspath()`. If there are duplicate source folders, then this bit of code tries to add duplicate entries to the classpath.

The list of source folders passed to `IJavaProject#setRawClasspath()` can't contain any duplicates, and as such, this is where the error lies.

The fix is to remove duplicates when the source folders are collected in buildship. I am hesitant to not add duplicate entries into a projects source folders in the first place, because I am unfamiliar with the other uses of source folders.

### DIFF
--- a/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/GradleClasspathContainerInitializer.java
+++ b/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/GradleClasspathContainerInitializer.java
@@ -24,6 +24,7 @@ import com.google.common.base.Preconditions;
 import com.google.common.base.Predicate;
 import com.google.common.collect.FluentIterable;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
 
 import com.gradleware.tooling.toolingmodel.OmniEclipseGradleBuild;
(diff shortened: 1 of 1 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=471005

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/GradleClasspathContainerInitializer.java
```diff
@@ -171,6 +172,12 @@ public IClasspathEntry apply(OmniEclipseSourceDirectory directory) {
                 // @formatter:on
             }
         }).toList();
+
+        return removeDuplicatesFromSourceFolders(sourceFolders);
+    }
+
+    private List<IClasspathEntry> removeDuplicatesFromSourceFolders(List<IClasspathEntry> sourceFolders) {
+        return ImmutableSet.copyOf(sourceFolders).asList();
```
maxandersen: Does ImmutableSet preserve ordering ? normally classpath lists has an order thus if a set gets involved we'll need to make sure it guarantees order.

Ian-Stewart-Binks: I've answered this question in our conversion, but I'll post the reply here for posterity:

ImmutableSet.copyOf() keeps the order - http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/collect/ImmutableSet.html

etiennestuder:
Hi Ian. Thanks for your investigations and the PR. We have integrated. In the future, please also provide a test case exposing the problem and demonstrating the fix. @donat will write the test this time such that you have a template for the future.

PR #130, opened 2015-07-03
Bug 471711

Ian-Stewart-Binks:
_Import existing Gradle project from the workspace folder fails with 'ResourceException: Invalid project description'_

When importing a project that is located in the default location (workspace), the project description location URI attribute must be set to `null` (see `IProjectDescription.getLocationURI()` JavaDoc). Otherwise, LocationValidator thinks that the project being imported overlaps with another project in the workspace and throws the error that we see. When importing a project, the method `DefaultWorkspaceOperations.findProjectInFolder()` is used to retrieve the project description of a given project, but does not account for this subtlety. Before this PR, if a project were located in the default location, the method would add the absolute path as the location, and not `null`.

This issue was addressed in the `DefaultWorkspaceOperations.createProject()` method, but was not addressed in the `includeProject` method. This pull request applies the same fix but for the `includeProject` method.


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=471711

etiennestuder:
Thanks for your investigations and the fix, Ian.

PR #131, opened 2015-07-09
Bug 468928 - Expose info logging only if tracing is enabled in the platform configuration

Ian-Stewart-Binks:
*Expose info logging only if tracing is enabled in the platform configuration*

Tracing can be enabled in two ways:

* In the _Run Configurations..._ > _Tracing_ tab

* On the command line, by using the `-debug` flag. A _.options_ file must be located in the _eclipse_ directory, or a _.options_ file must be specified after the `-debug` flag.


Ian-Stewart-Binks:
@donat _org.eclipse.buildship.ui/.options_ deleted, and only info statements are affected by the tracing.

PR #132, opened 2015-07-15
Bug 470071 - Includes and excludes are not applied to source folders

Ian-Stewart-Binks:
Temporary fix for [Bug 470071](https://bugs.eclipse.org/bugs/show_bug.cgi?id=470071)

If the user modifies the entries of the _.classpath_ file that correspond to source folders in the Gradle model, those modifications are ignored when the project's classpath is updated by Buildship (e.g., via `Gradle > Refresh Gradle Project`).

(I noted that this is a temporary fix, as it has been noted that the Tooling API will expose source folder filters.)

### DIFF
--- a/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/SourceFolderUpdater.java
+++ b/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/SourceFolderUpdater.java
@@ -79,26 +79,21 @@ public final class SourceFolderUpdater {
                 FileUtils.ensureFolderHierarchyExists(sourceDirectory);
                 final IPackageFragmentRoot root = SourceFolderUpdater.this.project.getPackageFragmentRoot(sourceDirectory);
                 IClasspathAttribute fromGradleModel = JavaCore.newClasspathAttribute(CLASSPATH_ATTRIBUTE_FROM_GRADLE_MODEL, "true");
-                
-                IPath[] includes = new IPath[] {};
-                IPath[] excludes = new IPath[] {};
 
-                // preserve the includes/excludes defined by the user. 
+                // preserve the includes/excludes defined by the user.
                 // this should eventually be provided by the Tooling API,
                 // see Bug 470071
                 Optional<IClasspathEntry> userDefinedClasspathEntry = FluentIterable.from(rawClasspath).firstMatch(new Predicate<IClasspathEntry>() {
-                    
+
                     @Override
                     public boolean apply(IClasspathEntry entry) {
-                        return root.getPath().toString().equals(entry.getPath().toString());
+                        return root.getPath().equals(entry.getPath());
                     }
                 });
 
-                if (userDefinedClasspathEntry.isPresent()) {
-                    includes = userDefinedClasspathEntry.get().getInclusionPatterns();
-                    excludes = userDefinedClasspathEntry.get().getExclusionPatterns();
-                }
-                
+                IPath[] includes = userDefinedClasspathEntry.isPresent() ? userDefinedClasspathEntry.get().getInclusionPatterns() : new IPath[] {};
+                IPath[] excludes = userDefinedClasspathEntry.isPresent() ? userDefinedClasspathEntry.get().getExclusionPatterns() : new IPath[] {};
+
                 // @formatter:off
                 return JavaCore.newSourceEntry(root.getPath(),
                         includes,                                    // use manually defined inclusion patterns, include all if none exist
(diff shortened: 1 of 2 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=470071

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/SourceFolderUpdater.java
```diff
@@ -68,19 +69,40 @@ private void updateClasspath(IProgressMonitor monitor) throws JavaModelException
...
             @Override
             public IClasspathEntry apply(OmniEclipseSourceDirectory directory) {
                 IFolder sourceDirectory = SourceFolderUpdater.this.project.getProject().getFolder(Path.fromOSString(directory.getPath()));
                 FileUtils.ensureFolderHierarchyExists(sourceDirectory);
-                IPackageFragmentRoot root = SourceFolderUpdater.this.project.getPackageFragmentRoot(sourceDirectory);
+                final IPackageFragmentRoot root = SourceFolderUpdater.this.project.getPackageFragmentRoot(sourceDirectory);
                 IClasspathAttribute fromGradleModel = JavaCore.newClasspathAttribute(CLASSPATH_ATTRIBUTE_FROM_GRADLE_MODEL, "true");
+                
+                IPath[] includes = new IPath[] {};
+                IPath[] excludes = new IPath[] {};
```
donat: It can be made a bit more compact: 
IPath[] includes = userDefinedClasspathEntry.isPresent() ? userDefinedClasspathEntry.get().getInclusionPatterns() : new IPath[] {};
                IPath[] excludes = userDefinedClasspathEntry.isPresent() ? userDefinedClasspathEntry.get().getExclusionPatterns() : new IPath[] {};

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/SourceFolderUpdater.java
```diff
@@ -68,19 +69,40 @@ private void updateClasspath(IProgressMonitor monitor) throws JavaModelException
...
+
+                // preserve the includes/excludes defined by the user. 
+                // this should eventually be provided by the Tooling API,
+                // see Bug 470071
+                Optional<IClasspathEntry> userDefinedClasspathEntry = FluentIterable.from(rawClasspath).firstMatch(new Predicate<IClasspathEntry>() {
+                    
+                    @Override
+                    public boolean apply(IClasspathEntry entry) {
+                        return root.getPath().toString().equals(entry.getPath().toString());
```
donat: No need to compare the strings; can be replaced with:  
return root.getPath().equals(entry.getPath());

fbricon:
@donat : so the `exclude` attribute (is there an `include` one?) defined in https://docs.gradle.org/current/dsl/org.gradle.api.tasks.SourceSet.html is not exposed through the tooling API yet? Is that what Ian is referring to in his 1st comment?

From my understanding of the origin bug report, the user wants to have his changes survive gradle configuration refresh. My take is build.gradle should be the truth and the sourceset  `exclude`s should be modified by the user instead.

Now having both (initializing inc/exc patterns from gradle then surviving user changes in eclipse) is doable I think, but wayy more complex. I'm not certain this is something buildship should even try to do.

etiennestuder:
I would not count on the Tooling API expose includes/excludes anytime soon (unless we get a PR for it). But, for now, what we can do is make the user-configured excludes/includes survive a refresh.

fbricon:
Ok so I take it we need (as part of a Java configurator) invoke an init.gradle script, that'd expose a custom model containing these patterns, along the java source/target version and all other attributes that we might need. WDYT @etiennestuder?

etiennestuder:
Yes. That will be 'quicker' than waiting for it to be in the Tooling API. This model would be part of a third-party extension, though (not of Buildship itself) since in Buildship we want to exclusively use the Tooling API and enhance its model as needed.

As mentioned to Ian, before this configurator story is tackled, we will need a spec for it that explains what the goal is and how is intended to be achieved in detail (maybe there is also another pattern than what is used for m2e). That document can be shared with the community and serves as the base for discussions until it gets to the matureness where implementation can start.

The reason I mention this is that I really want to avoid any implementation without everybody involved being on the same page regarding the functional goals and the implementation.

fbricon:
Fair enough, I'll get @Ian-Stewart-Binks to write some doc, so we have something to discuss on our next meeting.

donat:
Merged.

PR #133, opened 2015-07-19
Bug 472193 - UI Tests Crash When Run From Eclipse

Ian-Stewart-Binks:
The SWTBot tests crash. Certain code in the  `UiPlugin`'s `start` and `stop` methods needs to be run on the UI thread, which is something that is [not guaranteed](http://help.eclipse.org/mars/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Freference%2Fapi%2Forg%2Feclipse%2Fui%2Fplugin%2FAbstractUIPlugin.html) for those methods (note the `startup` method in the reference).

This PR runs the code that requires the UI thread in `start`/`stop` on the UI thread.

### DIFF
--- a/org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/UiPlugin.java
+++ b/org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/UiPlugin.java
@@ -22,6 +22,7 @@ import org.osgi.framework.ServiceRegistration;
 import org.eclipse.debug.core.DebugPlugin;
 import org.eclipse.jface.resource.ImageRegistry;
 import org.eclipse.ui.ISelectionService;
+import org.eclipse.ui.PlatformUI;
 import org.eclipse.ui.plugin.AbstractUIPlugin;
 import org.eclipse.swt.widgets.Display;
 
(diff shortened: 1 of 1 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=472193

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/UiPlugin.java
```diff
@@ -127,17 +128,30 @@ private void registerListeners() {
...
 
     @SuppressWarnings({"cast", "RedundantCast"})
     private void unregisterListeners() {
-        // if the selection service is disposed, then the listeners are already removed
-        // (see the javadoc on ISelectionService.addSelectionListener(ISelectionListener listener))
-        ISelectionService selectionService = ((ISelectionService) getWorkbench().getService(ISelectionService.class));
-        if (selectionService != null) {
-            selectionService.removeSelectionListener(this.contextActivatingSelectionListener);
-        }
+        Display.getDefault().asyncExec(new Runnable() {
```
donat: The Display object should be obtained through the IWorkbench interface and should be tested if it was already disposed. IMHO it should look like this:

Display display = PlatformUI.getWorkbench().getDisplay();
if (display != null && !display.isDisposed()) {
    display.asyncExec(...)
}

Ian-Stewart-Binks:
@donat Good idea, and I've made the change. I can squash the commit if you'd like. I decided not to, as [the last time I did](https://github.com/eclipse/buildship/pull/131), your comments were deleted.

donat:
If the code is corrected I don't mind loosing my comment. It won't be forgotten until the time of the merge anyway.

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/UiPlugin.java
```diff
@@ -127,16 +129,35 @@ private void registerListeners() {
...
-            selectionService.removeSelectionListener(this.contextActivatingSelectionListener);
+        Display display = PlatformUI.getWorkbench().getDisplay();
+        if (display != null && !display.isDisposed()) {
+            display.asyncExec(new Runnable() {
+
+                @Override
+                public void run() {
+                     // if the selection service is disposed, then the listeners are already removed
+                    // (see the javadoc on ISelectionService.addSelectionListener(ISelectionListener listener))
+                    ISelectionService selectionService = ((ISelectionService) getWorkbench().getService(ISelectionService.class));
```
fbricon: why is this calling `getWorkbench().getService(...)` but the previous block is calling `getWorkbench().getActiveWorkbenchWindow().getService(...)`?

Ian-Stewart-Binks: Taking me some time to find the difference online. @SimonScholz, it appears as though you wrote the initial code, do you have more insight?

SimonScholz: Thanks for pointing at this. Actually every WorkbenchWindow or MWindow(for e4 Apps) has its own ISelectionService. This means that the ISelectionService should be obtained from the workbenchwindow's getService(...) method rather than from the workbench itself, like I've done it in the registerListeners() method of the UIPlugin class.

SimonScholz:
I just added this PR: https://bugs.eclipse.org/bugs/show_bug.cgi?id=473389, which also resolves the issues with the different ways how the ISelectionService is obtained

Ian-Stewart-Binks:
@SimonScholz Nice!

PR #135, opened 2015-07-22
Bug472223

SimonScholz:
* Added a warning in case the import project location is inside the workspace location
* In case the rootProject.name property of the settings.gradle file is different to the folder name, the folder name will now be changed to the rootProject.name in case the folder is inside the workspace location

### DIFF
--- a/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/projectimport/ProjectImportJob.java
+++ b/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/projectimport/ProjectImportJob.java
@@ -7,40 +7,13 @@
  *
  * Contributors:
  *     Etienne Studer & Donát Csikós (Gradle Inc.) - initial API and implementation and initial documentation
+ *     Simon Scholz <simon.scholz@vogella.com> - Bug 472223
  */
 
 package org.eclipse.buildship.core.projectimport;
 
 import java.io.File;
 import java.util.List;
-
-import org.gradle.tooling.ProgressListener;
-
-import com.google.common.base.Function;
-import com.google.common.base.Optional;
-import com.google.common.base.Preconditions;
-import com.google.common.collect.FluentIterable;
-import com.google.common.collect.ImmutableList;
-
-import com.gradleware.tooling.toolingmodel.OmniEclipseGradleBuild;
-import com.gradleware.tooling.toolingmodel.OmniEclipseProject;
-import com.gradleware.tooling.toolingmodel.OmniGradleProject;
-import com.gradleware.tooling.toolingmodel.repository.FetchStrategy;
-import com.gradleware.tooling.toolingmodel.repository.FixedRequestAttributes;
-import com.gradleware.tooling.toolingmodel.repository.ModelRepository;
-import com.gradleware.tooling.toolingmodel.repository.TransientRequestAttributes;
-import com.gradleware.tooling.toolingmodel.util.Maybe;
-
-import org.eclipse.core.resources.IProject;
-import org.eclipse.core.resources.IProjectDescription;
-import org.eclipse.core.resources.IWorkspaceRoot;
-import org.eclipse.core.resources.ResourcesPlugin;
-import org.eclipse.core.runtime.IPath;
-import org.eclipse.core.runtime.IProgressMonitor;
-import org.eclipse.core.runtime.SubProgressMonitor;
-import org.eclipse.core.runtime.jobs.IJobManager;
-import org.eclipse.core.runtime.jobs.Job;
-import org.eclipse.jdt.launching.JavaRuntime;
 
 import org.eclipse.buildship.core.CorePlugin;
 import org.eclipse.buildship.core.configuration.GradleProjectNature;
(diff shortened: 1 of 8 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=472223

etiennestuder:
Why do we check the rootProject.name property and not the eclipseRootProject.name? We always use the OmniEclipseProject.name to define the folder name.

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/wizard/project/GradleProjectWizardPage.java
```diff
@@ -99,6 +99,19 @@ private void bindToConfiguration() {
...
+                // TODO currently a
+                // com.gradleware.tooling.toolingutils.binding.ValidationListener
+                // only receives error messages, therefore the warning message
+                // is implemented here, but it should also be done by a
+                // com.gradleware.tooling.toolingutils.binding.Validator<T>
+                if (NONE == getMessageType()) {
+                    // in case the current message type is NONE, a warning may
+                    // be shown
+                    if (WorkspaceUtils.isInWorkspaceFolder(projectDir)) {
+                        setMessage("It is recommended to locate your Gradle projects outside the workspace location",
```
donat: I'd only state that for certain configurations the root project folder have to be renamed.

donat: Externalize message.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/projectimport/ProjectImportJob.java
```diff
@@ -21,45 +43,24 @@
...
-import org.eclipse.buildship.core.workspace.WorkspaceOperations;
-
 /**
  * Imports a Gradle project into Eclipse using the project import coordinates given by a
  * {@code ProjectImportConfiguration} instance.
  */
 public final class ProjectImportJob extends ToolingApiWorkspaceJob {
 
-    private final FixedRequestAttributes fixedAttributes;
+    private FixedRequestAttributes fixedAttributes;
```
donat: Prefer to be immutable. Store the reference of the `ProjectImportConfiguration` and then pass around the `fixedAttributes`.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/projectimport/ProjectImportJob.java
```diff
@@ -102,6 +110,21 @@ public void runToolingApiJobInWorkspace(IProgressMonitor monitor) {
         // monitor is closed by caller in super class
     }
 
+    public void changeProjectRootFolderName(IProgressMonitor monitor) {
```
donat: Should be private.

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/wizard/project/GradleProjectWizardPage.java
```diff
@@ -99,6 +99,19 @@ private void bindToConfiguration() {
             public void modifyText(ModifyEvent e) {
                 File projectDir = FileUtils.getAbsoluteFile(GradleProjectWizardPage.this.projectDirText.getText()).orNull();
                 getConfiguration().setProjectDir(projectDir);
+                // TODO currently a
```
donat: Implement the validation properly. 
1. Add the validator into the org.eclipse.buildship.core.util.binding.Validators class.
2. Implement an `and()` validator in the same class. 
3. Combine the existing and the new validator in the controller class with the `and()` validator

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/wizard/project/ProjectPreviewWizardPage.java
```diff
@@ -336,8 +342,29 @@ public void onSuccess(Pair<OmniBuildEnvironment, OmniGradleBuildStructure> resul
             // the job has already taken care of logging the success
             this.latch.countDown();
 
+            OmniGradleBuildStructure gradleBuildStructure = result.getSecond();
             updateSummary(result.getFirst());
-            populateTree(result.getSecond());
+            populateTree(gradleBuildStructure);
+
```
donat: Remove extra whitespaces.

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/wizard/project/GradleProjectWizardPage.java
SimonScholz: Talked with @donat that the aim is different here, because I'd just want to show a warning rather than an error, which is currently not supported by a Validator. Eclipse usually works with an org.eclipse.core.runtime.IStatus to distinguish between different status, like errors and warnings. Therefore we could add a derived buildship specific validator, which can also works IStatus objects or find a different approach for the tooling api to distinguish between certain states.
@etiennestuder What do you think about it?

etiennestuder:
If a folder is located right under the root of the workspace, does the name of the project always have to be the name of the folder?

donat:
Unfortunately yes. We can't set the location entry when the project in the workspace. Consequently the project location resolution is based on the name, like ${workspace_loc}/project_name. A complete solution would be not not to rely on the project names.

donat:
A slightly cleaned-up version of this PR is created here: https://github.com/eclipse/buildship/pull/139

donat:
PR #139 Contains all changes from here plus an extra code clean-up. I close this PR  and continue the conversation at the other one.

PR #122, opened 2015-06-16
Bug 469636 - Add Smart Import support

Ian-Stewart-Binks:
This pull request adds [Smart Import](https://wiki.eclipse.org/E4/UI/Smart_Import) functionality for Buildship. The pull request adds the `org.eclipse.buildship.ui.importer` with the necessary files required by Smart Import, adds tests to the `org.eclipse.buildship.ui.test` module, and modifies three existing Buildship files: `ProjectImportJob.java`, `WorkspaceOperations.java` and `DefaultWorkspaceOperations.java`. The latter two files are changed in a way identical to changes made in [this pull request](https://github.com/eclipse/buildship/pull/100/files), with the purpose of exposing the nature adding functionality.

This pull request comes with an issue: due to the Smart Import dependency (`org.eclipse.e4.ui.importer`), this feature requires Eclipse Mars as the target platform, whereas Buildship supports platforms as far back as e36.

Note: Only Gradle projects with the Gradle wrapper are supported.

PR #137, opened 2015-07-24
Bug 471095 - Close project preview before job finishes results in SWTException

Ian-Stewart-Binks:
This pull request brings many issues into the light:

-- Edit: The proposed fix for 1 has been merged in.

1. **The first set of SWTExceptions**
  When a project preview is started, a `ProjectPreviewJob` is triggered with a `DelegatingProgressListener`. The method `DelegatingProgressListener.statusChanged()` is called periodically as the project preview is initialized. When this method is called, the `DelegatingProgressListener` updates the corresponding monitor by calling `monitor.subtask...` When the user cancels the request to view the project preview, and closes the project import wizard, the `ProjectPreviewJob.fetchGradleBuildStructure()` method continues, and the `DelegatingProgressListener` continues to listen for events. The `ProjectPreviewJob.fetchGradleBuildStructure()` continually updates the `DelegatingProgressListener`, which then tries to call `monitor.subtask...`, however, the monitor has already been canceled at this point, at which time an `SWTException` is thrown.
  * The point at which the first `SWTException` is thrown is [here](https://github.com/gradle/tooling-commons/blob/master/toolingmodel/src/main/java/com/gradleware/tooling/toolingmodel/repository/internal/DefaultModelRepository.java#L370), which throws a `UncheckedExecutionException`. I'm assuming that the listeners are notified at some point here.
  * In `DelegatingProgressListener`, the events that the monitor is being updated with oscillate between "Compile script into cache", "Configure project".
  * The solution is to check whether the monitor is canceled in `DelegatingProgressListener`. A better solution would be to remove the listener from the `ProjectPreviewJob`, and let the job die silently. I've looked into unregistering/cancelling listeners/transient request attributes right when the job is canceled, but with no success.
  * Current solution _may_ still be flawed, should monitor be canceled in between when the `DelegatingProgressListener` checks if it's canceled, and when it calls `monitor.subtask...`.

2. **The second set of SWTExceptions**
  When that same job is canceled, the `JobChangeListener` that's listening for `done` in `ProjectPreviewJob` will call `resultHandler.onFailure...` which then subsequently calls `ProjectPreviewWizardPage.clearTree()`. The tree is disposed at this point, and naturally results in an `SWTException`.
  * The solution is to just check if the tree is disposed before trying to access it.
  * This PR does *not* introduce a test case for this solution. I am uncertain as to how to go about testing this.

3. **Induction of SWT tests within Spock tests**
  This PR uses SWTBot tests inside Spock tests, which appears to be unorthodox for Buildship. In the event that this PR is accepted, it would be wise to create an issue to further the conquest of Spock in the testing of Buildship. Spock uses the Sputnik test runner, and SWTBotTests use the SWTBotJunit4ClassRunner. This means that the SWTBot tests can be run in Spock, but we lose the ability to take screenshots. I've looked into how we can add this functionality back in, and it should be doable, but I have so far been unsuccessful. I think it would be a good idea to open another issue on this topic, or at least merge it in with the SWTBot -> Spock issue. I'm very interested in looking into this!
  * Note that this PR essentially ports over the `BaseSWTBotTest` class into the `SWTBotTestHelper` class. Not much of the functionality has changed, however `waitForJobs` has. I have implemented it such that it no longer checks to see if all jobs from the `JobManager` are null, but rather checks to see if the `JobManager` is idle. With the latter approach, the tests would not close properly, because `waitForJobs` would not wait for the long running `ProjectPreviewJob`  (interestingly, the job would still run, but the `JobManager` would return null for `currentJob` inside of _ProjectPreviewWizardPageUiTest_, but not null inside of _ProjectPreviewWizardPage_. It turns out that they're being run on different threads, but `currentJob` only works for the same thread that it's called from.
  * In _SWTBotTestHelper_, access modifiers have been preserved.

4. **A note on the logging that this PR introduces**
  Logging has been added for the sole purpose of being able to properly test. 
  * Not all instances where widgets in the `ProjectPreviewWizardPage` are accessed are tested. You will notice that methods that are called in `onSuccess`, such as `populateTree` have had the same fix as `clearTree` applied to them. This is meant for consistency, and for a potential race condition when a job successfully completes, but the user simoultaneously disposes the wizard before these methods are called. This may be _too_ precautious.

### DIFF
--- a/org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/SWTBotTestHelper.groovy
+++ b/org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/SWTBotTestHelper.groovy
@@ -1,3 +1,14 @@
+/*
+ * Copyright (c) 2015 the original author or authors.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     Ian Stewart-Binks (Red Hat, Inc.) - Bug 471095
+ */
+
 package org.eclipse.buildship.ui
 
 import org.eclipse.swtbot.eclipse.finder.SWTWorkbenchBot
--- a/org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/SWTBotTestHelper.groovy
+++ b/org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/SWTBotTestHelper.groovy
@@ -1,0 +1,94 @@
+package org.eclipse.buildship.ui
+
+import org.eclipse.swtbot.eclipse.finder.SWTWorkbenchBot
+import org.eclipse.swtbot.eclipse.finder.widgets.SWTBotView
+import org.eclipse.swtbot.swt.finder.exceptions.WidgetNotFoundException
+import org.eclipse.swtbot.swt.finder.widgets.SWTBotShell
+import org.eclipse.swtbot.swt.finder.finders.UIThreadRunnable
+import org.eclipse.swtbot.swt.finder.results.VoidResult
+import org.eclipse.swtbot.swt.finder.results.BoolResult
+import org.eclipse.ui.PlatformUI
+import org.eclipse.core.runtime.jobs.Job
+import org.eclipse.swt.widgets.Display
+
+class SWTBotTestHelper {
+    static SWTWorkbenchBot swtBot;
+
+    public static SWTWorkbenchBot getBot() {
+        if (swtBot == null) {
+            return new SWTWorkbenchBot()
+        }
+        return swtBot
+    }
+
+    private static void closeWelcomePageIfAny() throws Exception {
+        try {
+            SWTBotView view = getBot().activeView()
+            if (view.getTitle().equals("Welcome")) {
+                view.close()
+            }
+        } catch (WidgetNotFoundException e) {
+            UiPlugin.logger().error("Failed to initialize SWTBot test.", e)
+        }
+    }
+
+    public static void closeAllShellsExceptTheApplicationShellAndForceShellActivation() {
+        SWTWorkbenchBot bot = getBot()
+
+        // in case a UI test fails some shells might not be closed properly, therefore we close
+        // these here and log it
+        SWTBotShell[] shells = bot.shells()
+        for (SWTBotShell swtBotShell : shells) {
+            if (swtBotShell.isOpen() && !isEclipseApplicationShell(swtBotShell)) {
+                bot.captureScreenshot(swtBotShell.getText() + " NotClosed.jpg")
+                UiPlugin.logger().warn(swtBotShell.getText() + " was not closed properly.")
+                swtBotShell.close()
+            }
+        }
+
+        // http://wiki.eclipse.org/SWTBot/Troubleshooting#No_active_Shell_when_running_SWTBot_tests_in_Xvfb
+        UIThreadRunnable.syncExec(new VoidResult() {
+
+                    @Override
+                    public void run() {
+                        PlatformUI.getWorkbench().getActiveWorkbenchWindow().getShell().forceActive()
+                    }
+                })
+    }
+
+    public static boolean isEclipseApplicationShell(final SWTBotShell swtBotShell) {
+        return UIThreadRunnable.syncExec(new BoolResult() {
+
+                    @Override
+                    public Boolean run() {
+                        return PlatformUI.getWorkbench().getActiveWorkbenchWindow().getShell().equals(swtBotShell.widget)
+                    }
+                })
+    }
+
+    protected static void waitForJobsToFinish() {
+        while (!Job.getJobManager().isIdle()) {
+            delay(500)
+        }
+    }
+
+    private static void delay(long waitTimeMillis) {
+        Display display = Display.getCurrent()
+        if (display != null) {
+            long endTimeMillis = System.currentTimeMillis() + waitTimeMillis
+            while (System.currentTimeMillis() < endTimeMillis) {
+                if (!display.readAndDispatch()) {
+                    display.sleep()
+                }
+            }
+            display.update()
+        } else {
+            try {
+                Thread.sleep(waitTimeMillis)
+            } catch (InterruptedException e) {
+                // ignore
+            }
+        }
+    }
+
+}
--- a/org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/wizard/project/ProjectPreviewWizardPageUiTest.groovy
+++ b/org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/wizard/project/ProjectPreviewWizardPageUiTest.groovy
@@ -1,3 +1,14 @@
+/*
+ * Copyright (c) 2015 the original author or authors.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     Ian Stewart-Binks (Red Hat, Inc.) - Bug 471095
+ */
+
 package org.eclipse.buildship.ui.wizard.project
 
 import org.eclipse.swtbot.eclipse.finder.waits.Conditions
--- a/org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/wizard/project/ProjectPreviewWizardPageUiTest.groovy
+++ b/org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/wizard/project/ProjectPreviewWizardPageUiTest.groovy
@@ -1,0 +1,76 @@
+package org.eclipse.buildship.ui.wizard.project
+
+import org.eclipse.swtbot.eclipse.finder.waits.Conditions
+import org.eclipse.core.runtime.NullProgressMonitor
+import org.eclipse.buildship.core.CorePlugin
+import org.eclipse.buildship.core.Logger
+import org.eclipse.buildship.ui.SWTBotTestHelper
+import org.eclipse.swtbot.swt.finder.widgets.SWTBotShell
+import org.eclipse.swtbot.eclipse.finder.SWTWorkbenchBot
+import org.junit.Rule
+import org.junit.rules.TemporaryFolder
+import spock.lang.Specification
+import org.eclipse.buildship.ui.test.fixtures.TestEnvironment
+import org.eclipse.core.runtime.jobs.Job
+
+
+class ProjectPreviewWizardPageUiTest extends Specification {
+    SWTWorkbenchBot bot = SWTBotTestHelper.getBot()
+
+    @Rule
+    TemporaryFolder tempFolder
+    File location
+
+    def setup() {
+        SWTBotTestHelper.closeAllShellsExceptTheApplicationShellAndForceShellActivation()
+    }
+
+    def setupSpec() {
+        SWTBotTestHelper.closeWelcomePageIfAny()
+    }
+
+    def "Stop preview and close import wizard before import job finishes"() {
+        given:
+        Logger logger = Mock()
+        TestEnvironment.registerService(Logger, logger)
+        location = tempFolder.newFolder("new-folder")
+
+        new File(location.toString() + "/build.gradle").withWriter('utf-8') { writer ->
+            writer.writeLine "Thread.sleep(5000)"
+            writer.writeLine "task takesLongTimeToLoad {"
+            writer.writeLine "    Thread.sleep(5000)"
+            writer.writeLine "    doLast {"
+            writer.writeLine "        Thread.sleep(5000)"
+            writer.writeLine "    }"
+            writer.writeLine "}"
+        }
+
+        CorePlugin.workspaceOperations().createProject('project-name', location, [], [], new NullProgressMonitor())
+
+        when:
+        startImportPreviewAndCancelWizard()
+        SWTBotTestHelper.waitForJobsToFinish()
+
+        then:
+        0 * logger.error(_, _)
+
+        cleanup:
+        TestEnvironment.cleanup()
+    }
+
+    private void startImportPreviewAndCancelWizard() {
+        bot.menu("File").menu("Import...").click()
+        SWTBotShell shell = bot.shell("Import")
+        shell.activate()
+        bot.waitUntil(Conditions.shellIsActive("Import"))
+        bot.tree().expandNode("Gradle").select("Gradle Project")
+        bot.button("Next >").click()
+        bot.button("Next >").click()
+        bot.textWithLabel("Project root directory").setText(location.toString())
+        bot.button("Next >").click()
+        bot.button("Next >").click()
+        bot.toolbarButtonWithTooltip("Cancel Operation").click()
+        bot.button("Cancel").click()
+    }
+
+}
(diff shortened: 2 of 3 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=471095

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/wizard/project/ProjectPreviewWizardPage.java
```diff
@@ -376,14 +391,21 @@ private void populateTree(final OmniGradleBuildStructure buildStructure) {
...
+                              ProjectPreviewWizardPage.this.projectPreviewTree.removeAll();
+
+                              // populate the tree from the build structure
+                              OmniGradleProjectStructure rootProject = buildStructure.getRootProject();
+                              TreeItem rootTreeItem = new TreeItem(ProjectPreviewWizardPage.this.projectPreviewTree, SWT.NONE);
+                              rootTreeItem.setExpanded(true);
+                              rootTreeItem.setText(rootProject.getName());
+                              populateRecursively(rootProject, rootTreeItem);
+                        }
+                    } catch (SWTException e) {
```
fbricon: I don't think you should catch SWTException here. Since you check for the disposed widget before. Anything goes wrong after that, I'd rather get it popped up, so we notice it and fix it. Simply logging the error, we might lose track of potential errors in the future.

btw it doesn't necessarily means `Failed to properly close ProjectPreviewWizardPage` when that happens, even if there's a good chance it does.

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/wizard/project/ProjectPreviewWizardPage.java
```diff
@@ -376,14 +391,21 @@ private void populateTree(final OmniGradleBuildStructure buildStructure) {
 
                 @Override
                 public void run() {
```
fbricon: if (!ProjectPreviewWizardPage.this.projectPreviewTree.isDisposed()) {
  return;
}

--> less lines changed

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/wizard/project/ProjectPreviewWizardPage.java
```diff
@@ -393,7 +415,14 @@ private void clearTree() {
...
-                    ProjectPreviewWizardPage.this.projectPreviewTree.removeAll();
+                    try {
+                        if (!ProjectPreviewWizardPage.this.projectPreviewTree.isDisposed()) {
+                            ProjectPreviewWizardPage.this.projectPreviewTree.removeAll();
+                        }
+                    } catch (SWTException e) {
```
fbricon: don't catch that, see previous comment

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/wizard/project/ProjectPreviewWizardPage.java
```diff
@@ -353,20 +356,32 @@ private void updateSummary(final OmniBuildEnvironment buildEnvironment) {
...
 
-                    // update Java home
-                    String javaHome = buildEnvironment.getJava().getJavaHome().getAbsolutePath();
-                    ProjectPreviewWizardPage.this.javaHomeLabel.setText(javaHome);
+                        if (!ProjectPreviewWizardPage.this.javaHomeLabel.isDisposed()) {
+                            // update Java home
+                            String javaHome = buildEnvironment.getJava().getJavaHome().getAbsolutePath();
+                            ProjectPreviewWizardPage.this.javaHomeLabel.setText(javaHome);
+                        }
+                    } catch (SWTException e) {
```
fbricon: don't catch that, see similar comments

org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/SWTBotTestHelper.groovy
```diff
@@ -0,0 +1,94 @@
...
+        if (swtBot == null) {
+            return new SWTWorkbenchBot()
+        }
+        return swtBot
+    }
+
+    private static void closeWelcomePageIfAny() throws Exception {
+        try {
+            SWTBotView view = getBot().activeView()
+            if (view.getTitle().equals("Welcome")) {
```
fbricon: is it possible the active view is null?

org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/SWTBotTestHelper.groovy
```diff
@@ -0,0 +1,94 @@
...
+import org.eclipse.swtbot.swt.finder.exceptions.WidgetNotFoundException
+import org.eclipse.swtbot.swt.finder.widgets.SWTBotShell
+import org.eclipse.swtbot.swt.finder.finders.UIThreadRunnable
+import org.eclipse.swtbot.swt.finder.results.VoidResult
+import org.eclipse.swtbot.swt.finder.results.BoolResult
+import org.eclipse.ui.PlatformUI
+import org.eclipse.core.runtime.jobs.Job
+import org.eclipse.swt.widgets.Display
+
+class SWTBotTestHelper {
```
fbricon: if this is supposed to be a simgleton, then you need to make the constructor private

org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/SWTBotTestHelper.groovy
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    private static void closeWelcomePageIfAny() throws Exception {
+        try {
+            SWTBotView view = getBot().activeView()
+            if (view.getTitle().equals("Welcome")) {
+                view.close()
+            }
+        } catch (WidgetNotFoundException e) {
+            UiPlugin.logger().error("Failed to initialize SWTBot test.", e)
```
fbricon: error doesn't match  "closeWelcomePageIfAny" action

org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/SWTBotTestHelper.groovy
```diff
@@ -0,0 +1,94 @@
...
+                    @Override
+                    public Boolean run() {
+                        return PlatformUI.getWorkbench().getActiveWorkbenchWindow().getShell().equals(swtBotShell.widget)
+                    }
+                })
+    }
+
+    protected static void waitForJobsToFinish() {
+        while (!Job.getJobManager().isIdle()) {
+            delay(500)
```
fbricon: cut it down to 100ms

org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/SWTBotTestHelper.groovy
```diff
@@ -0,0 +1,94 @@
+package org.eclipse.buildship.ui
```
fbricon: missing copyright header

org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/SWTBotTestHelper.groovy
```diff
@@ -0,0 +1,94 @@
...
+
+    public static void closeAllShellsExceptTheApplicationShellAndForceShellActivation() {
+        SWTWorkbenchBot bot = getBot()
+
+        // in case a UI test fails some shells might not be closed properly, therefore we close
+        // these here and log it
+        SWTBotShell[] shells = bot.shells()
+        for (SWTBotShell swtBotShell : shells) {
+            if (swtBotShell.isOpen() && !isEclipseApplicationShell(swtBotShell)) {
+                bot.captureScreenshot(swtBotShell.getText() + " NotClosed.jpg")
```
fbricon: why take a picture before closing the shell?

org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/SWTBotTestHelper.groovy
```diff
@@ -0,0 +1,94 @@
...
+                UiPlugin.logger().warn(swtBotShell.getText() + " was not closed properly.")
```
fbricon: afaict you haven't tried to close it yet

org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/wizard/project/ProjectPreviewWizardPageUiTest.groovy
```diff
@@ -0,0 +1,76 @@
+package org.eclipse.buildship.ui.wizard.project
```
fbricon: missing copyright header

org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/wizard/project/ProjectPreviewWizardPageUiTest.groovy
```diff
@@ -0,0 +1,76 @@
...
+
+
+class ProjectPreviewWizardPageUiTest extends Specification {
+    SWTWorkbenchBot bot = SWTBotTestHelper.getBot()
+
+    @Rule
+    TemporaryFolder tempFolder
+    File location
+
+    def setup() {
```
fbricon: don't you want to screenshot the thing if the test fail or was it punted to later?

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/wizard/project/ProjectPreviewWizardPage.java
Ian-Stewart-Binks: If we don't catch it, then there is no way to log it, which means that there is no way to test for it. In an earlier conversation between @donat and I, it was suggested that this fix implement the logging.

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/wizard/project/ProjectPreviewWizardPage.java
Ian-Stewart-Binks: But is less lines changed necessarily a good reason to do one approach over another?

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/wizard/project/ProjectPreviewWizardPage.java
Ian-Stewart-Binks: I agree on the second note you said. I will change the message to something more general and informative. As for the first comment, I'll copy my later comment here:
_If we don't catch it, then there is no way to log it, which means that there is no way to test for it. In an earlier conversation between @donat and I, it was suggested that this fix implement the logging._

org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/SWTBotTestHelper.groovy
Ian-Stewart-Binks: As far as I understand, no. If there is no active view, then a `WidgetNotFoundException` is thrown.

org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/SWTBotTestHelper.groovy
Ian-Stewart-Binks: Interesting. I agree, but perhaps there was some earlier intention to log it. @donat, do you have any insight?

org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/SWTBotTestHelper.groovy
Ian-Stewart-Binks: These seem to be magic numbers to me, and I only used 500 because I'm directly copying existing Buildship code. What's the reason behind decreasing it?

org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/SWTBotTestHelper.groovy
Ian-Stewart-Binks: I believe that these shells were not intended to stay open. If one is open, then this must be an error.

org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/SWTBotTestHelper.groovy
Ian-Stewart-Binks: This class is a bit weird. It's not entirely meant to be a singleton. The _BaseSWTBotTest_ class was inherited by the Buildship test cases, which means that both the `BaseSWTBotTest` and its child could share the same `SWTWorkbenchBot`. This approach delegates all bot control to the `SWTBotTestHelper` class, and uses only static methods, much like our implementation in the [Smart Import](https://github.com/eclipse/buildship/pull/122) feature.

This class is no doubt weird, though, and is something that I pointed out in my [request for review](https://github.com/Ian-Stewart-Binks/buildship/pull/7).

org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/SWTBotTestHelper.groovy
Ian-Stewart-Binks: Donat and Etienne technically wrote most of the code. Should I add them to the copyright header as well?

org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/SWTBotTestHelper.groovy
Ian-Stewart-Binks: As I understand it, it's the responsibility of the test to close the shell. If a shell was not closed, then a test is not doing its duty.

org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/SWTBotTestHelper.groovy
Ian-Stewart-Binks: @fbricon told me yesterday that the reason for cutting the time from 500 to 100 is to decrease the overall time that a series of tests that depend on this method would take.

org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/wizard/project/ProjectPreviewWizardPageUiTest.groovy
Ian-Stewart-Binks: Copyright header has been added.

org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/SWTBotTestHelper.groovy
Ian-Stewart-Binks: @fbricon ?

fbricon: Yes, add them.

donat:
Based on your reviews the changes were pulled, altered and merged to master. The full list of changes can be reviewed here: https://github.com/eclipse/buildship/compare/e5b4e78cf364...01349f752c92

Ian-Stewart-Binks:
@donat Thank you.

PR #141, opened 2015-07-28
Bug 472191 - UI tests hang when run from Eclipse

Ian-Stewart-Binks:
SWTBot tests simply need to be run on a non-UI thread, this is not a fault of Buildship. I believe that the best solution to this problem is to simply inform the developer.


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=472191

etiennestuder:
Thanks for your investigations and the findings.

PR #142, opened 2015-07-29
Bug 473348 - Exclusion of transitive dependencies is ignored

SimonScholz:
It seems that the order of the IClasspathEntry objects is important here. The Gradle support from STS also uses this new order. I also reproduced this by adding a project dependency directly in the project properties and changing the order.


eclipsewebmaster:
Issue tracker reference:

SimonScholz:
The following image shows how this can be reproduced

project_dependency-order-counts

PR #143, opened 2015-07-30
Bug 473348 - Exclusion of transitive dependencies is ignored

SimonScholz:
Without the second boolean in the newContainerEntry method this boolean is false by default, which means that not all container dependencies are exported

### DIFF
--- a/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/GradleClasspathContainer.java
+++ b/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/GradleClasspathContainer.java
@@ -7,6 +7,7 @@
  *
  * Contributors:
  *     Etienne Studer & Donát Csikós (Gradle Inc.) - initial API and implementation and initial documentation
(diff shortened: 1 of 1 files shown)


eclipsewebmaster:
Issue tracker reference:

donat:
But if a library is exported under the classpath container, it's still visible from an external project, right?

etiennestuder:
We need to ensure we have inter test for both scenarios. library exported (Gradle 2.4) and library not exported (Gradle 2.5).

Etienne

SimonScholz:
Right. Of course I've tested that beforehand. 
I am now working on a spock test to ensure this.
But manual tests prove this already

etiennestuder:
I know that you certainly tested it. I only meant that we need an automated integration test for those scenarios.

Etienne

SimonScholz:
I am currently working on this, but I still face problems receiving the problem marker. The STS uses some kind of timer for this, which is not really satisfying. Maybe I find a better approach.

etiennestuder:
Are markers the only way to see problems?

PR #145, opened 2015-07-31
Bug 473348 - Added integration tests to check dependency exclusion

SimonScholz:
* Added test to ensure that a project is correctly resolved with exported transitive project dependencies
* Added a test to ensure that the exclusion of a transitive project dependency causes compile errors


eclipsewebmaster:
Issue tracker reference:

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/dependencies/DependencyTests.groovy
```diff
@@ -0,0 +1,253 @@
...
+
+        when:
+        job.schedule()
+        job.join()
+
+        then:
+        CorePlugin.workspaceOperations().findProjectByName('dependent-project').present
+
+        when:
+        def depentProject = CorePlugin.workspaceOperations().findProjectByName('dependent-project')
```
Ian-Stewart-Binks: Possible typo.

PR #147, opened 2015-07-31
Bug 473862 - F5 key shortcut doesn't refresh project folder contents

Ian-Stewart-Binks:
Buildship's refresh functionality current overrides the default Eclipse refresh functionality.

This pull request solves the problem by listening for Eclipse's default refresh command. When Buildship detects that the default refresh is being called, it invokes its own refresh. This means that Buildship's refresh command is not limited to just being bound to the 'F5' key, as it was previously, and it also ensures that any changes made to the default Eclipse refresh are persisted (i.e., this is better than just calling `refreshLocal`, as such an approach only tries to mimic the refresh functionality. If the default refresh changed, then Buildship would break the default Eclipse functionality. Such a change can be found [here](https://github.com/Ian-Stewart-Binks/buildship/tree/Bug-473348).)

This approach also ensures that the refresh can be invoked manually by the user through File > Gradle > Refresh Gradle Project'.

### DIFF
--- a/org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/BaseSWTBotTest.java
+++ b/org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/BaseSWTBotTest.java
@@ -25,6 +25,7 @@ import org.eclipse.swtbot.swt.finder.finders.UIThreadRunnable;
 import org.eclipse.swtbot.swt.finder.junit.SWTBotJunit4ClassRunner;
 import org.eclipse.swtbot.swt.finder.results.BoolResult;
 import org.eclipse.swtbot.swt.finder.results.VoidResult;
+import org.eclipse.swtbot.swt.finder.utils.SWTBotPreferences;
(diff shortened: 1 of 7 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=473862

gamerson:
I can confirm that this pull does restore the original platform refresh behavior.  Thanks.

donat:
I reviewed the PR and merged it to the local buildship/Ian-Stewart-Binks-Bug-473348B branch to check it against our CI environment. Looks good, waiting for @etiennestuder to integrate it to master.

Ian-Stewart-Binks:
@etienne, I've added a new SWTBot test that tests that the default refresh functionality still performs correctly. I can also make UI tests for the overall refresh functionality, if you wish.

etiennestuder:
Thanks for the fix, Ian. Reviewed and integrated.

etiennestuder:
I've asked @donat to integrate the SWTBot today.

PR #149, opened 2015-08-10
Bug 469011 - Show actual values during a Gradle build in the command line

Ian-Stewart-Binks:
Prior to this pull request, when running rasks from the Gradle Tasks view in Eclipse on a project that has not specified values for the Gradle Home and Java Home attributes, their values would not be displayed to the user in the console. This pull request changes the output such that those values are included.

Known issues:
- Should the user choose to specify a version of Gradle, then the Gradle User Home attribute becomes `null`. 
    Example:
    ```
Gradle Distribution: Specific Gradle version 2.2

Gradle User Home: Use default defined by Gradle (null)

Gradle Version: 2.2
```
This value is null in both the `buildEnvironment` and the launch configurations.

- A better solution might embed the Gradle fetching functionality into the launch configurations, instead of mending the launch configurations and the build environment together. Do the launch configurations really have no access to the java home/gradle home variables otherwise?

- The new class, `GradleBuildFetcher`, implicitly assumes the fetching strategy. This may be undesirable, but can certainly be refactored. Nevertheless, something to note.

- The changes also introduce the `Gradle Version` output. I saw that this was output in the `ProjectPreviewWizardPage` and thought it would be helpful here, too, regardless of whether or not the wrapper is being used.


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=469011

etiennestuder:
Thanks, Ian. I've merged it after some tweaking.

PR #150, opened 2015-08-11
update gradle wrapper version to use gradle 2.6

breskeby:
- update wrapper version to be 2.6
- fix EclipseTestExecuter to be 2.6 compliant
- remove fixed configuration of wrapper version (quite inconvenient when updating)


breskeby:
test by running "./gradlew clean check" locally on osx

PR #152, opened 2015-08-18
Bug 475015 - Have preview pages reviewed by native speaker

Ian-Stewart-Binks:
This pull request makes minor modifications to the project import/creation messages to make them sound more natural.

The existing messages already sounded natural, and are very well-written. I was curious about some of the uses of *best of* and *the best of*, but I concluded that their current uses are appropriate. 

A message that stuck out to me the most was *All imported projects that already contain an Eclipse .project file will be left alone, aside from being added the Gradle nature.*. I've changed it, but I'm uncertain whether it's any better.

@donat / @etienne, Are there any messages in particular that you're concerned about? I can ask some of the other people in the Red Hat office what they think.


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=475015

etiennestuder:
Thanks for the review, Ian.

PR #155, opened 2015-08-24
Bug 474472 - Use the common Request<T> interface

SimonScholz:
Change-Id: Ic7f0aebb4ae04e69496d5bcd026ca5ea31086609
Signed-off-by: Simon Scholz <simon.scholz@vogella.com>


SimonScholz:
Short descriptions what the PR does:
=============================
Currently most methods and classes use the com.gradleware.tooling.toolingclient.BuildLaunchRequest interface, which causes that other com.gradleware.tooling.toolingclient.Request<T> instances cannot be used.

This patch replaces the BuildLaunchRequest type with the Request<T> type, because the Request<T> type contains all desired methods already.
So the com.gradleware.tooling.toolingclient.TestLaunchRequest can now be used.

PR #156, opened 2015-08-24
Bug 474472 - Enable test runs

SimonScholz:
Have static factories for different kinds of RunGradleConfigurationDelegateJobs


SimonScholz:
Short descriptions what the PR does:
=============================

* This patch makes it possible to create different kinds of RunGradleConfigurationDelegateJobs by using static factories (Effective Java Book: Item 1)
 * One for default builds, which were available before
 * One for testreruns

* The new createTestReRunDelegateJob method allows to pass TestOperationDescriptors to the job, which will be rerun, when scheduling this job.

etiennestuder:
A similar implementation is now in master.

PR #158, opened 2015-08-26
Execution view rerun failed tests

SimonScholz:
Added Action to rerun failed tests from the execution view


SimonScholz:
* Adds a new "Rerun failed tests" item to the toolbar of the ExecutionView
* The Icon for the Item is the same as the one for rerunning failed tests in the JUnit View from JDT
* The Action for rerunning failed tests is also a org.gradle.tooling.events.ProgressListener, which is attached to the buildrequest in order to store all failed tests, so that they can be easily rerun, if there are test failures
* In case there are no test failures the toolbar item won't be enabled

etiennestuder:
A similar implementation is now on master.

PR #160, opened 2015-08-27
Use Predicates#hasGradleNature() in Classpath refresher and ContextActivatingSelectionListener

SimonScholz:
Resolves the Task: "There is a mismatch between the logic when the "Refresh Project" context entry is made visible in the context menu and when the execution starts. We should unify the logic from the predicate used by ContextActivatingSelectionListener and GradleClasspathContainerRefresher#collectSelectedProjects()."

* It lets the GradleClasspathContainerRefresher also use the Predicates#hasGradleNature() predicate for getting the right projects to be refreshed

### DIFF
--- a/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/util/collections/AdapterFunction.java
+++ b/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/util/collections/AdapterFunction.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright (c) 2015 the original author or authors.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     Simon Scholz (vogella GmbH) - initial API and implementation and initial documentation
+ */
+
+package org.eclipse.buildship.core.util.collections;
+
+import org.eclipse.core.runtime.IAdapterManager;
+
+import com.google.common.base.Function;
+
+/**
+ * {@link Function} which is used to get a certain adapter from an input object.
+ *
+ * @param <T>
+ */
+public final class AdapterFunction<T> implements Function<Object, T> {
+
+    private Class<T> adapter;
+    private IAdapterManager adapterManager;
+
+    public AdapterFunction(Class<T> adapter, IAdapterManager adapterManager) {
+        this.adapter = adapter;
+        this.adapterManager = adapterManager;
+    }
+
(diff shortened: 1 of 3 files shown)


org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/util/collections/AdapterFunction.java
```diff
@@ -0,0 +1,39 @@
...
+import org.eclipse.core.runtime.IAdapterManager;
+
+import com.google.common.base.Function;
+
+/**
+ * {@link Function} which is used to get a certain adapter from an input object.
+ *
+ * @param <T>
+ */
+public class AdapterFunction<T> implements Function<Object, T> {
```
etiennestuder: can be final

org.eclipse.buildship.ui/plugin.xml
```diff
@@ -171,6 +171,22 @@
...
+                     </equals>
+                  </iterate>
+                  <iterate
+                        operator="or">
+                  </iterate>
+               </with>
+            </and>
+         </activeWhen>
       </handler>
     </extension>
```
etiennestuder: do we really need two (nested) <or> here?

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/workspace/GradleClasspathContainerRefresher.java
```diff
@@ -47,22 +46,16 @@ public static void execute(final ExecutionEvent event) {
...
-            }
+            @SuppressWarnings("unchecked")
+            ImmutableList<IProject> selectedProjects = FluentIterable.from(selection.toList())
+                    .transform(new AdapterFunction<IProject>(IProject.class, Platform.getAdapterManager()))
+                    .filter(Predicates.hasGradleNature()).toList();
+
+            return selectedProjects;
         }
-        return result.build();
+        return ImmutableList.of();
```
etiennestuder: can you put this into an else statement?

etiennestuder:
Once my 3 comments have been addressed, I will merge this in.

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/workspace/GradleClasspathContainerRefresher.java
SimonScholz: The two iterate elements came in by accident and where not intended.
But I am curious what's the benefit of having an else statement here?

SimonScholz:
The two iterate elements came in by accident and where not intended.
But I am curious what's the benefit of having an else statement here?

PR #161, opened 2015-09-01
Eclipse proxy integration

Ian-Stewart-Binks:
**Update:** This is slightly out of date - there is support for concurrent jobs now in the `EclipseProxySettingsSupporter` class.

This is an update for the [Proxy story](https://github.com/eclipse/buildship/blob/f57052cb1489196a21698c92efc64b2c3775ee59/docs/stories/Proxy.md). This PR is not meant to be merged into master.

#### How it works:
The `CorePlugin` registers the `IProxyService`. The `ToolingApiJob` and `ToolingApiWorkspaceJob` each call `EclipseProxySettingsSupporter` to complete three tasks:
1. Save the current proxy settings in the `System` properties.
2. Configure the `System` properties based on the Eclipse proxy settings (retrieved via the `IProxyService`).
3. Restore the saved `System` properties upon termination.

The main work that has been done on this story in this pull request is on the _ProxySettingsTest_. I will leave some line notes to clarify some of the decisions that were made. The tests have only successfully run from Eclipse. Please note that quite a few of the files have been taken from the Gradle repository.

#### Some current difficulties:
1. I can no longer set the proxy userId and password. I was playing around with Eclipse, and set up a secure storage password. With this in place, you apparently cannot change the proxy user or password. I got rid of the secure storage password, however I now get the exception: `org.eclipse.equinox.security.storage.StorageException: No secure storage modules found.`. (Before this, it would say that no default value could be found for the user or password, and that some unencrypted value was being stored there). This is my current blocking issue.

    Funnily enough, I recently rebased this branch onto master, and needed to perform some Eclipse tasks to make things compile/work. Prior to the rebase, I had the secure storage password enabled, however I was able to change the proxy user/password freely. As such, the cause of the problem is currently unknown.

    This affects every other test that needs to use the proxies, at this point. This issue may affect my ability to assess some of the other difficulties that I'm having with some of the other tests.

    **Update:** It works now, and I don't think I did anything besides eat lunch. I'll add this as a solution to my what-to-do-when-eclipse-won't-work list.
    **Update 2:** False alarm. It just started to remember the password that I was trying to feed it... Can't change it from `test-password` now...

2. The tests fail from the command line, but not from Eclipse. I have not looked into this yet.
The error output states that there are seven failures, and says that the errors are:
    ```
AvailablePortFinder. initializationError
AvailablePortFinder. initializationError
HttpServer. initializationError
UserAgentMatcher. initializationError
UserAgentMatcher. initializationError
TestProxyServer. initializationError
TestProxyServer. initializationError
```

3. I haven't been able to figure out what to do for the _JVM arguments are automatically set when Eclipse proxy settings are available_ test yet. I had been working on this about a week ago, and I was unable to retrieve the JVM args from within the Gradle JVM (`BuildEnvironment`). Etienne pointed out that I shouldn't make the assumption that just because the `BuildEnvironment` doesn't output the JVM args I'm expecting, the JVM args that I'm expecting aren't being set.

4. For the _Proxies are accessed from Gradle build VM_ test:
    - The test would pass, and it's mirror test _"Proxies are not accessed from Gradle build VM with incorrect proxy settings"_ would also pass. The  _Proxies are accessed from Gradle build VM_ test would cause several 407 errors before being able to get through the proxy server. I cannot reproduce this right now due to difficulty number 1.
    - The test would only issue one request to the proxy server under the current configuration. Under different configurations, it would also request the `jar` and the `sha1`, with approx one HEAD and one GET request to each. The current configuration only causes the `pom` file to be downloaded. I don't know why it doesn't download the `jar` as well, but this could be due to my lack of familiarity with how Maven/pom files work.

5. For the Gradle distribution test, the zip is currently empty. This doesn't affect the outcome of the test, but it does raise an exception `org.gradle.tooling.GradleConnectionException: Could not install Gradle distribution from 'http://not.a.real.domain/gradlew/dist'.`.

#### Things left to do:
1. Ensure that the proxy settings can be stored, accessed, and reset atomically (or some such) to ensure that the original proxy settings are persisted even when jobs are running concurrently. One can imagine race conditions such that the restored System properties are actually the properties set by a previously run `ToolingApiJob`. Under this as well, I'd need to ask for clarification on whether this is something that needs to be worried about.

2. Include HTTPS in the tests.

3. Clean up imports in _ProxySettingsTest_, add JavaDocs, add Documentation about Proxy settings, ensure that all dependencies are necessary and correct, ensure that build.gradle file and MANIFEST.MF are set up correctly in _org.eclipse.buildship.core_.

4. Ensure that tests are correct and cover all cases well. Create test to test for correct setting of JVM arguments. Ensure that tests run correctly from command line.

5. Make sure the checkStyle passes given files taken from Gradle.

6. Ensure that fewer proxy servers/Gradle daemons are created.

#### Note
When trying to run the tests, the Buildship UI plugins need to be switched off in the _Run Configurations..._. Otherwise, the Buildship UI plugin will try to register a listener to the events in _RunGradleConfigurationDelegateJob_, and the tests will not run. This is also true of the _RunGradleConfigurationDelegateJobTest_ class. I intent to add this to the documentation in a separate PR.

If I remember any other difficulties, or encounter any more today, I'll be sure to update this PR description.

### CODE REVIEW

File: org.eclipse.buildship.core.test/META-INF/MANIFEST.MF [hunk 1]
```diff
@@ -10,11 +10,19 @@ Comment: junit has to be a required bundle, otherwise the tests won't launch for
...
 Bundle-ClassPath: .,
  lib/cglib-nodep-3.1.jar,
+ lib/commons-lang-2.6.jar,
+ lib/commons-codec-1.6.jar,
  lib/groovy-all-2.0.5.jar,
  lib/guava-18.0.jar,
  lib/hamcrest-core-1.3.jar,
+ lib/hamcrest-library-1.3.jar,
+ lib/jetty-6.1.25.jar,
+ lib/jetty-util-6.1.25.jar,
```
Ian-Stewart-Binks: I added dependencies here, and to the classpath/build.gradle file based on what dependencies were required by the incoming Gradle files. If you see anything that looks out of place, let me know. It probably is.

org.eclipse.buildship.core.test/build.gradle
```diff
@@ -18,9 +26,7 @@ eclipseTest {
 
 // append the sources of each first-level dependency and its transitive dependencies of
 // the 'bundled' configuration to the 'bundledSource' configuration
-configurations.bundled.resolvedConfiguration.firstLevelModuleDependencies.each { dep ->
-    addSourcesRecursively(dep)
-}
+configurations.bundled.resolvedConfiguration.firstLevelModuleDependencies.each { dep -> addSourcesRecursively(dep) }
```
Ian-Stewart-Binks: I have a tick of pressing `Ctrl+Shift+F`. I'll revert changes like this.

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/proxy/ProxySettingsTest.groovy
```diff
@@ -0,0 +1,348 @@
...
+
+
+        // Gets rid of The HTTP proxy is enabled in the preferences but disabled in the system settings
+        final IProxyData httpProxyData = proxyService.getProxyData(IProxyData.HTTP_PROXY_TYPE)
+
+        httpProxyData.setHost(host)
+        httpProxyData.setPort(port)
+        httpProxyData.setUserid(userId)
+        httpProxyData.setPassword(password)
+        proxyService.setProxyData((IProxyData[]) [
```
Ian-Stewart-Binks: It turns out that `proxyService.setProxyData` automatically sets the System settings given certain `IProxyService` configurations (under some configurations, it just wipes them out). Lines 245-248 and 262-273 are meant to combat this behaviour to ensure that _"Different proxy settings can be used by subsequent builds"_ succeeds.

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/proxy/ProxySettingsTest.groovy
```diff
@@ -0,0 +1,348 @@
...
+import org.eclipse.buildship.core.test.fixtures.TestEnvironment
+import com.gradleware.tooling.toolingclient.BuildLaunchRequest
+import com.gradleware.tooling.toolingclient.ToolingClient
+import org.hamcrest.*
+import org.eclipse.buildship.core.util.gradle.GradleDistributionWrapper
+import com.gradleware.tooling.toolingclient.GradleDistribution
+
+import org.eclipse.buildship.core.test.fixtures.ProjectImportSpecification
+
+class ProxySettingsTest extends ProjectImportSpecification {
```
Ian-Stewart-Binks: > ProjectImportSpecification

Probably unecessary.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/CorePlugin.java
```diff
@@ -187,7 +195,30 @@ private UserNotification createUserNotification() {
         return new ConsoleUserNotification();
     }
 
+    /*
+     * https://resheim.net/2008/10/connecting-through-proxy.html
```
Ian-Stewart-Binks: I'm not certain how to attribute work to other people, especially to what seem to be tutorials. Any suggestions are appreciated.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/CorePlugin.java
```diff
@@ -187,7 +195,30 @@ private UserNotification createUserNotification() {
...
+     */
+    private IProxyService retrieveProxyService() {
+        final String CORE_NET_BUNDLE = "org.eclipse.core.net";
+        Bundle bundle = Platform.getBundle(CORE_NET_BUNDLE);
+        while (bundle.getState() != Bundle.ACTIVE) {
+            try {
+                Thread.sleep(1000);
+            } catch (InterruptedException e) {
+                e.printStackTrace();
```
Ian-Stewart-Binks: Will likely need to change this function to conform with Buildship practices.

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/proxy/ProxySettingsTest.groovy
Ian-Stewart-Binks: http://stackoverflow.com/questions/32318136/iproxyservice-setproxydata-affects-system-properties

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/proxy/ProxySettingsTest.groovy
```diff
@@ -0,0 +1,348 @@
...
+        }
+
+    }
+
+    def newProjectImportJob(File location) {
+        ProjectImportConfiguration configuration = new ProjectImportConfiguration()
+        configuration.gradleDistribution = GradleDistributionWrapper.from(GradleDistribution.forRemoteDistribution("http://not.a.real.domain/gradlew/dist".toURI()))
+
+        // Ian-todo: In PR, emphasize that gradleUserHome has been changed. Perhaps also test to see if gradle distribution exists in workspace.
+        configuration.gradleUserHome = new File(ResourcesPlugin.getWorkspace().root.rawLocation.toString())
```
Ian-Stewart-Binks: Using gradle user home in workspace to contain downloaded dependencies, and to also flush the cache on test rerun.

donat: If your test class extends `ProjectImportSpecification` then you can i) you get the `workspaceLocation` property for free and 2) it is easier to define project layouts and import jobs.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/CorePlugin.java
donat: We shouldn't wait for the a bundle to become active. If a dependency exists from plug-in A to B, then B will be always active before A. If it's not the case, than further investigation is needed.

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/proxy/ProxySettingsTest.groovy
Ian-Stewart-Binks: @donat Thanks for the tip. `ProjectImportSpecification` doesn't allow the `ProjectImportConfiguration.gradleUserHome` attribute to be specified in `newProjectImportJob`, or any method for that matter. I would need to add the `newProjectImportJob` method to `ProjectImportSpecification`, or change the functionality of the `ProjectImportSpecification` class to support this.

@donat What are your thoughts?

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/CorePlugin.java
donat: If you have the link here, it's perfectly fine. Maybe use the one-line comment.

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/proxy/ProxySettingsTest.groovy
donat: The ProjectImportSpecification is not complete in the sense of parametrization. You can define a new `newProjectPreviewJob()` method there. I'd suggest the following signature: `newProjectPreviewJob(File, ProjectImportConfiguration)`.

Ian-Stewart-Binks: Good suggestion. Will do.

org.eclipse.buildship.core.test/META-INF/MANIFEST.MF
donat: What is needed from hamcrest-library which is not available in hamcrest-core?

Ian-Stewart-Binks: When taking it away, `Matchers` in the copied file `UserAgentMatcher` cannot be resolved.

Ian-Stewart-Binks: I've just noticed that there are some dependencies in the _lib_ folder that are not necessary. I'll take care of those.

PR #162, opened 2015-09-01
escape slashes in project name

idlsoft:
Slashes can appear in gradle project names, for example when using includeFlat('libs/lib1').
They need to be escaped when importting into eclipse


donat:
Thank you for the pull request! Would you mind updating the commit such that it has a signed-off-by entry at the end of the commit message? It is a requirement for all Eclipse projects: https://wiki.eclipse.org/Development_Resources/Contributing_via_Git#via_GitHub

idlsoft:
I'll try to do that for the next PR, in the meantime, this being a one line change. perhaps it can be applied manually.
Thank you.

donat:
PR Integrated manually.
Thanks again!

idlsoft:
Thank you.

etiennestuder:
I had to revert this commit since it is incomplete. We use project names in several places of the code base and all those usages must be converted in the same way, otherwise cross-references do not work. I think we will deal with this in tooling-commons, i.e. make it transparent to Buildship.

PR #163, opened 2015-09-03
Bug 471943 - Make Buildship work behind the firewall

Ian-Stewart-Binks:
This pull request allows Buildship to propagate proxy settings defined in Eclipse to the System settings. The original story can be found [here](https://github.com/eclipse/buildship/blob/f57052cb1489196a21698c92efc64b2c3775ee59/docs/stories/Proxy.md).

This pull request also tackles the issue of concurrency for different `ToolingApiJob`/`ToolingApiWorkspaceJob`s that are trying to store default system properties.

Tests have been excluded from this pull request such that it can be merged in before the Eclipse RC build. I will continue working on the tests. I have not performed smoke testing on the current code.

### DIFF
--- a/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/proxy/EclipseProxySettingsSupporter.java
+++ b/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/proxy/EclipseProxySettingsSupporter.java
@@ -1,0 +1,94 @@
+/*
+ * Copyright (c) 2015 the original author or authors.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     Ian Stewart-Binks (Red Hat, Inc.) - Bug 471943 - Make Buildship work behind the firewall
+ */
+
+package org.eclipse.buildship.core.proxy;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+import com.google.common.base.Optional;
+import org.eclipse.core.net.proxy.IProxyData;
+import org.eclipse.buildship.core.CorePlugin;
+
+/**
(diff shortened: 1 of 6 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=471943

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/proxy/EclipseProxySettingsSupporter.java
```diff
@@ -0,0 +1,94 @@
...
+public class EclipseProxySettingsSupporter {
+
+    private static String savedHTTPProxyHost, savedHTTPProxyPort, savedHTTPProxyUser, savedHTTPProxyPassword;
+    private static final Lock lock = new ReentrantLock();
+
+    /**
+     * Configures the System properties proxy settings based on the Eclipse proxy settings.
+     */
+    public static void configureEclipseProxySettings() {
+        synchronized(EclipseProxySettingsSupporter.class) {
```
Ian-Stewart-Binks: I've attempted to make this block atomic. Suppose a thread, thread A, attempts to call `configureEclipseProxySettings`, and a thread, thread B, tries to do the same. Let thread A execute `storeSystemProxySettings` first, and acquire the lock in that method. Now run thread B. Thread B fails and does not get the lock. Thread B then continues on and executes `configureHTTPProxySettings`. Thread B updates the system settings with the Eclipse proxy settings. Thread A runs, and tries to store the default system settings, however Thread B just overwrote them.

Ian-Stewart-Binks:
This PR only adds in HTTP support. I've added in HTTPS support before as well, but since we decided not to test HTTPS for now, I took it out. I can add it back in easily, if need be.

etiennestuder:
@Ian-Stewart-Binks I'm closing this PR since Donat (based on this PR from you) found a very simple solution, as explained in our channel. Thanks for haven laid the foundation for this fix.

Ian-Stewart-Binks:
Understood. Not a problem. I still have the WIP tests for this fix, Donat mentioned that they would be useful for the future.

PR #164, opened 2015-09-07
Bug 475637 - Project refresh add new and removes old projects

donat:
A few comments:
- RefreshGradleProjectJob extends WorkspaceJob to properly update the workspace creating/deleting files and folders.
- RefreshGradleProjectJob contains some duplication on how to add/remove natures, project configurations and how to import projects. I considered unifying these snippets out of this PR's scope.
- The Gradle classpath container works properly only if the project is set up correctly. Therefore I modified the GradleClasspathContainerInitializer to load the dependencies only if the Gradle nature is present on the project. If it's not the case, an empty list of dependencies is assigned to the project.
- Adding/removing modules in a Gradle project and calling refresh drives JDT into an inconsistent state and make Eclipse throw a strange exception: some preferences can't be read for the project. As a workaround, a clean build is called on all workspace project which are excluded from a Gradle build.


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=475637

PR #165, opened 2015-09-08
Update Buildship developer document

Ian-Stewart-Binks:
This PR updates the buildship developer document with information on
how to run isolated tests without encountering an exception caused by
the UI plugin.

The UI plugin will add a listener onto tests such as `RunGradleConfigurationDelegateJobTest`, but since the UI plugin isn't invoked, an exception is thrown due to the workbench not being created yet (and the workbench never gets created).

Perhaps it would be good to start a 'known issues' section for running the tests.


etiennestuder:
Thanks, Ian.

PR #170, opened 2015-09-14
Traverse from root OperationItem with TreeTraverser

SimonScholz:
Uses the Guava TreeTraverser to traverse through the structure, which allows to remove the usage of a ISetChangeListener on the knownElements of the TreeViewer's ContentProvider, which is called quite often, which should be avoided.

I also tested a solution, where the failed Tests are stored separately, so that traversing would not be necessary any more, but this didn't had a real impact on the performance. So I'd rather stick to the Guava TreeTraverser.


etiennestuder:
A really nice solution. Thanks, Simon.

PR #171, opened 2015-09-16
Bug 477365 - Show multiple exceptions on one error dialog

SimonScholz:
The ExceptionDialog is now able to handle multiple exceptions and Throwables can also be added at runtime an opened ExceptionDialog. Therefore the ExceptionDialog has a new addException method, which adds Throwables to the observable list, which is bound to an errorViewer. This errorViewer is shown on the dialog in case there are more than one Throwable, which should be shown.

### DIFF
--- a/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/notification/UserNotification.java
+++ b/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/notification/UserNotification.java
@@ -19,12 +19,18 @@ public interface UserNotification {
     /**
      * Notifies the user about the occurrence of an error.
      *
-     * @param headline the headline of the error
-     * @param message the concise error description
-     * @param details the detailed error description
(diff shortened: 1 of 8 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=477365

SimonScholz:
exceptiondialoganimated

Just for the showcase I added the NPE and RuntimeException by clicking on the copy button.
So once new errors are added to the ExceptionDialog by using its addThrowable method, we get this errorViewer with a list of Throwables, which can be selected and seen in the details. Multiselection is also possible, so that the user may copy all stackstraces that are selected in the errorViewer.
Of course it is now also possible to directly instantiate the  ExceptionDialog with multiple Throwables.

etiennestuder:
Looking good. I will give feedback soon.

SimonScholz:
Added several SWTBot tests for the dialog

donat:
[Merged](https://github.com/eclipse/buildship/commit/953da91a5502023a81161a8066ef3cb27c87efce)

PR #172, opened 2015-09-21
Consider classpath entries for projects with .project file

SimonScholz:
Added features in this PR:
* It respects existing .classpath files, so that previous contents remains if it is appropriate
* It also creates a .classpath file, in case a .project file already exists during the project import


oehme:
This has been superseded by the `Clean Import` feature that will be part of 1.0.9

PR #173, opened 2015-09-22
Bug 478054 - Eclipse does not prompt for workspace any more after installing BuildShip

SimonScholz:
This patch only starts the UI Plugin in case it is not started yet and therefore does not cause certain osgi errors any more.


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=478054

PR #174, opened 2015-09-22
Bug 478054 - Eclipse does not prompt for workspace any more after installing Buildship

SimonScholz:
Added the Bundle.START_TRANSIENT flag as suggested by Thomas Watson. See https://bugs.eclipse.org/bugs/show_bug.cgi?id=478054


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=478054

PR #175, opened 2015-09-23
Bug 474473 - Allow to reruns tests from the Java editor

SimonScholz:
This PR adds the possibility to run tests with Gradle from the "Run as" menu in the editor or the package explorer.

It currently just prints what is going to be run.
The actual implementation of the test run needs to be done later, as well as finding already existent ILaunchConfiguration[] objects.
See the TODOs in the org.eclipse.buildship.ui.launch.TestLaunchShortcut class

### DIFF
--- a/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/RunGradleJvmTestLaunchRequestJob.java
+++ b/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/RunGradleJvmTestLaunchRequestJob.java
@@ -1,0 +1,103 @@
+/*
+ * Copyright (c) 2015 the original author or authors.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     Simon Scholz (vogella GmbH) - initial API and implementation and initial documentation
+ */
+
+package org.eclipse.buildship.core.launch;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.OutputStreamWriter;
+import java.text.DateFormat;
+import java.util.Date;
+
+import org.eclipse.buildship.core.CorePlugin;
+import org.eclipse.buildship.core.console.ProcessDescription;
+import org.eclipse.buildship.core.i18n.CoreMessages;
+
+import com.google.common.base.Joiner;
+import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableList;
+import com.gradleware.tooling.toolingclient.Request;
+import com.gradleware.tooling.toolingclient.TestConfig;
+
+/**
+ * Executes tests through Gradle based on a given list of {@code IJavaElement}
+ * instances and a given set of {@code GradleRunConfigurationAttributes}.
+ */
+public final class RunGradleJvmTestLaunchRequestJob extends BaseLaunchRequestJob {
+
+    private final GradleRunConfigurationAttributes configurationAttributes;
+    private ImmutableList<String> testClasses;
+
+    public RunGradleJvmTestLaunchRequestJob(Iterable<String> testClasses,
+            GradleRunConfigurationAttributes configurationAttributes) {
+        super("Launching Gradle Tests", false);
+        this.testClasses = ImmutableList.copyOf(testClasses);
+        this.configurationAttributes = Preconditions.checkNotNull(configurationAttributes);
+    }
+
+    @Override
+    protected String getJobTaskName() {
+        return "Launch Gradle Tests";
+    }
+
+    @Override
+    protected GradleRunConfigurationAttributes getConfigurationAttributes() {
+        return this.configurationAttributes;
+    }
+
+    @Override
+    protected ProcessDescription createProcessDescription() {
+        String processName = createProcessName(this.configurationAttributes.getWorkingDir());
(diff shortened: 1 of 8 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=474473

org.eclipse.buildship.ui/plugin.xml
```diff
@@ -263,6 +263,50 @@
...
+                      <adapt
+                            type="org.eclipse.core.resources.IResource">
+                         <test
+                               forcePluginActivation="true"
+                               property="org.eclipse.core.resources.projectNature"
+                               value="org.eclipse.buildship.core.gradleprojectnature">
+                         </test>
+                      </adapt>
+                      <adapt type="org.eclipse.jdt.core.IJavaElement">
+		                <test property="org.eclipse.jdt.core.isInJavaProject"/>
```
donat: As discussed today, we should not restrict the users what classes are they trying to launch as tests. Just let the users launch any classes/methods they can find.

donat: Please check if it works with Groovy tests.

org.eclipse.buildship.ui/plugin.xml
```diff
@@ -263,6 +263,50 @@
 
     <!-- properties view for tasks and projects -->
```
donat: The comment belongs to the subsequent <extension> node

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/launch/TestLaunchShortcut.java
```diff
@@ -0,0 +1,160 @@
...
+            ITextSelection textSelection = (ITextSelection) selection;
+
+            IJavaElement elementAtOffset = SelectionConverter.getElementAtOffset(element, textSelection);
+            if (!(elementAtOffset instanceof IMethod)) {
+                return Optional.absent();
+            }
+
+            IMethod method = (IMethod) elementAtOffset;
+
+            ISourceRange nameRange = method.getNameRange();
```
donat: Please add comment about what it does (this this used to check if the user just selected the method name and not something in the method body)
Please add a reference where you found the same approach (just a simple comment, no javadoc)

org.eclipse.buildship.ui/plugin.xml
```diff
@@ -263,6 +263,50 @@
...
+             </contextLabel>
+             <enablement>
+                <with
+                      variable="selection">
+                   <iterate
+                         ifEmpty="false"
+                         operator="and">
+                      <adapt
+                            type="org.eclipse.core.resources.IResource">
+                         <test
```
donat: Can we define an enablement as follows
- project with Gradle nature
- selection instanceof ITextSelection

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/RunGradleJvmTestLaunchRequestJob.java
```diff
@@ -0,0 +1,220 @@
...
+import com.gradleware.tooling.toolingclient.Request;
+import com.gradleware.tooling.toolingclient.TestConfig;
+
+/**
+ * Executes tests through Gradle based on a given list of {@code IJavaElement}
+ * instances and a given set of {@code GradleRunConfigurationAttributes}.
+ */
+public final class RunGradleJvmTestLaunchRequestJob extends BaseLaunchRequestJob {
+
+    private final ImmutableList<IJavaElement> javaElements;
```
donat: The Job class should not know about JDT (just like the `RunGradleTestLaunchRequestJob` doesn't know about `OperationItem`s). 
The conversion should be part of the launch shortcut implementation (or in a helper class possibly).

SimonScholz:
@donat All of your comments should be resolved now

In case the groovy tooling for Eclipse is installed everything works, because the specific Groovy types are derived from IJavaElements. I also added a test whether "spock.lang.Specification" is on the classpath. (see UI's plugin.xml line 314) You might want to add additional items for the org.eclipse.jdt.core.hasTypeOnClasspath tests.


Please also check the createProcessName and writeExtraConfigInfo methods in both new BaseLaunchRequestJob implementations. I am not sure what content should be in the output there.

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/launch/TestLaunchShortcut.java
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    protected void performLaunch(List<? extends IJavaElement> elementsToLaunch, String mode) {
+        // get the right workingDirExpression for the
+        // GradleRunConfigurationAttributes
+        IJavaElement javaElement = elementsToLaunch.get(0);
+        IProject project = javaElement.getResource().getProject();
+        String workingDirExpression = project.getLocation().toOSString();
+
+        GradleRunConfigurationAttributes configurationAttributes = GradleRunConfigurationAttributes.with(
```
SimonScholz: @donat Are the GradleRunConfigurationAttributes configured right? I am not sure, how to easily  get the GradleDistribution or other Gradle specific information from an IProject

donat: CorePlugin.projectConfigurationManager().readProjectConfiguration(IProject).getRequestAttributes()

PR #177, opened 2015-09-25
Bug 478425 - The Gradle menu should also be available on project members

SimonScholz:
This PR makes it possible to have the Gradle Menu on any Project member and also improves the code with a new ProjectFunction class, which has a convenience method for getting an IProject from an object.


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=478425

donat:
The PR doesn't work correctly: if the ui plugin is not activated then the menu items don't show up.

PR #179, opened 2015-10-14
Bug 479243 - Runtasks command not enabled in several Eclipse perspectives

SimonScholz:
This PR turn the NodeSelection Object into an IStructuredSelection, which is propagated to the Workbench by introducing a NodeSelectionProviderWrapper, which is applied by the TaskView like this:

```java
 this.selectionProvider = new NodeSelectionProviderWrapper(this, getTreeViewer());
 getSite().setSelectionProvider(this.selectionProvider);
```

This PR not only provides a different approach to handle the enablement of Handlers, but also makes the ISelection, which is propagated to the workbench more detailed and let's NodeSelections behave in a natural Eclipse manner.

### DIFF
--- a/org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/util/nodeselection/NodeSelection.java
+++ b/org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/util/nodeselection/NodeSelection.java
@@ -7,9 +7,13 @@
  *
  * Contributors:
  *     Etienne Studer & Donát Csikós (Gradle Inc.) - initial API and implementation and initial documentation
+ *     Simon Scholz <simon.scholz@vogella.com> - Bug 479243
  */
 
 package org.eclipse.buildship.ui.util.nodeselection;
+
+import java.util.Iterator;
+import java.util.List;
 
 import com.google.common.base.Function;
 import com.google.common.base.Predicate;
@@ -17,15 +21,14 @@ import com.google.common.base.Predicates;
 import com.google.common.collect.FluentIterable;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
+
 import org.eclipse.jface.viewers.ISelection;
 import org.eclipse.jface.viewers.IStructuredSelection;
-
-import java.util.List;
 
 /**
  * Provides information about a given set of selected nodes.
  */
-public final class NodeSelection {
+public final class NodeSelection implements IStructuredSelection {
 
     private static final NodeSelection EMPTY = new NodeSelection(ImmutableList.of());
 
@@ -35,13 +38,9 @@ public final class NodeSelection {
         this.nodes = ImmutableList.copyOf(nodes);
     }
 
-    /**
-     * Returns whether the selection is empty.
-     *
-     * @return {@code true} if the selection is empty, {@code false} otherwise
-     */
+    @Override
     public boolean isEmpty() {
-        return this.nodes.isEmpty();
+        return getNodes().isEmpty();
     }
 
     /**
@@ -50,7 +49,12 @@ public final class NodeSelection {
      * @return {@code true} if a single node is selected, {@code false} otherwise
      */
     public boolean isSingleSelection() {
-        return this.nodes.size() == 1;
+        return size() == 1;
+    }
+
+    @Override
+    public Object getFirstElement() {
+        return isEmpty() ? null : this.nodes.get(0);
     }
 
     /**
@@ -59,6 +63,8 @@ public final class NodeSelection {
      * @return the first node
      * @throws java.lang.IllegalStateException thrown if the selection is empty
      */
+    // TODO remove this method and use getFirstElement(), which is defined by
+    // IStructuredSelection
     public Object getFirstNode() {
         if (isEmpty()) {
             throw new IllegalStateException("Selection is empty.");
@@ -74,6 +80,8 @@ public final class NodeSelection {
      * @return the first node
      * @throws java.lang.IllegalStateException thrown if the selection is empty
      */
+    // TODO rename this method to getFirstElement(Class<T> expectedType) to have
+    // consistent names
     public <T> T getFirstNode(Class<T> expectedType) {
         if (isEmpty()) {
             throw new IllegalStateException("Selection is empty.");
@@ -87,6 +95,8 @@ public final class NodeSelection {
      *
      * @return the list of all nodes
      */
+    // TODO rename this method to "toList" to have
+    // consistent names
     public ImmutableList<?> getNodes() {
         return this.nodes;
     }
@@ -98,6 +108,8 @@ public final class NodeSelection {
      * @return the list of all nodes
      * @throws ClassCastException thrown if a node is not of the expected type
      */
+    // TODO rename this method to "toList" to have
+    // consistent names
     public <T> ImmutableList<T> getNodes(final Class<T> expectedType) {
         return FluentIterable.from(this.nodes).transform(new Function<Object, T>() {
 
@@ -106,6 +118,26 @@ public final class NodeSelection {
                 return expectedType.cast(input);
             }
         }).toList();
+    }
+
+    @Override
+    public Iterator<?> iterator() {
+        return this.nodes.iterator();
+    }
+
+    @Override
+    public int size() {
+        return this.nodes.size();
+    }
+
+    @Override
+    public Object[] toArray() {
+        return this.nodes.toArray();
+    }
+
+    @Override
+    public List<?> toList() {
+        return this.nodes;
     }
 
     /**
--- a/org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/util/nodeselection/SelectionHistoryManager.java
+++ b/org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/util/nodeselection/SelectionHistoryManager.java
@@ -7,22 +7,33 @@
  *
  * Contributors:
  *     Etienne Studer & Donát Csikós (Gradle Inc.) - initial API and implementation and initial documentation
+ *     Simon Scholz <simon.scholz@vogella.com> - Bug 479243
  */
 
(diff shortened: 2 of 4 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=479243

SimonScholz:
I updated the PR and now the SelectionHistoryManager acts as ISelectionProvider and the proper NodeSelection is available at once.

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/util/nodeselection/NodeSelection.java
```diff
@@ -135,7 +136,7 @@ public boolean allMatch(Predicate<Object> predicate) {
      * @param newSelection the node selection to merge with
      * @return a new instance containing the merge result
      */
-    public NodeSelection mergeWith(NodeSelection newSelection) {
+    public NodeSelection mergeWith(IStructuredSelection newSelection) {
```
donat: Do we have to prepare to merge with selections other than NodeSelection? I think the argument should be as it was before.

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/util/nodeselection/NodeSelectionProvider.java
```diff
@@ -23,4 +24,12 @@
      */
     NodeSelection getSelection();
 
+    /**
+     * Set the current selection.
+     *
+     * @param selection
+     *            the {@link NodeSelection} to be set
+     */
+    void setSelection(NodeSelection selection);
```
donat: This method is not used anywhere.

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/util/nodeselection/SelectionHistoryManager.java
```diff
@@ -7,22 +7,33 @@
...
+ * Stores the sequence in which the currently selected nodes were selected in
+ * the {@link TreeViewer} and acts as {@link ISelectionProvider}, which can be
+ * added to an {@link org.eclipse.ui.IViewSite}.
+ *
+ * @see org.eclipse.buildship.ui.view.task.TaskView
  */
-public final class SelectionHistoryManager {
+public final class SelectionHistoryManager implements ISelectionProvider {
+
+    private List<ISelectionChangedListener> selectionChangedListeners = new CopyOnWriteArrayList<ISelectionChangedListener>();
```
donat: Use Lists.newCopyOnWriteArrayList(), just like at MultiPageView#336

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/util/nodeselection/SelectionHistoryManager.java
```diff
@@ -40,32 +51,61 @@ private void init() {
...
+
+    protected void handleSelection(ISelection selection) {
         NodeSelection nodeSelection = NodeSelection.from(selection);
         this.selectionHistory = this.selectionHistory.mergeWith(nodeSelection);
+        fireSelectionChanged();
     }
 
-    public void dispose() {
-        this.treeViewer.removeSelectionChangedListener(this.listener);
+    protected void fireSelectionChanged() {
```
donat: Can you please elaborate on why we need a concurrent list and a SafeRunner in this case?

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/util/nodeselection/NodeSelection.java
```diff
@@ -74,7 +64,7 @@ public Object getFirstNode() {
      * @return the first node
      * @throws java.lang.IllegalStateException thrown if the selection is empty
      */
-    public <T> T getFirstNode(Class<T> expectedType) {
+    public <T> T getFirstElement(Class<T> expectedType) {
```
donat: Instead of renaming please just leave a todo there, it makes the PR unnecessarily big. This applies everywhere in this PR.

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/view/task/TaskView.java
```diff
@@ -33,13 +39,7 @@
...
 import org.eclipse.ui.part.ViewPart;
 
-import org.eclipse.buildship.core.CorePlugin;
-import org.eclipse.buildship.core.configuration.ProjectConfiguration;
-import org.eclipse.buildship.ui.external.viewer.FilteredTree;
-import org.eclipse.buildship.ui.external.viewer.PatternFilter;
-import org.eclipse.buildship.ui.util.nodeselection.NodeSelection;
-import org.eclipse.buildship.ui.util.nodeselection.NodeSelectionProvider;
-import org.eclipse.buildship.ui.util.nodeselection.SelectionHistoryManager;
+import com.gradleware.tooling.toolingmodel.repository.FetchStrategy;
```
donat: Please use the the import order from the Buildship Oomph model. In case you don't use it, take the import order definition from [here](https://dl.dropboxusercontent.com/u/1159431/buildship.importorder)

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/view/task/TaskView.java
```diff
@@ -201,11 +204,15 @@ public NodeSelection getSelection() {
...
+    }
+
+    @Override
     public void dispose() {
         this.state.dispose();
         this.selectionHistoryManager.dispose();
         this.uiContributionManager.dispose();
         super.dispose();
     }
-
```
donat: Re-add whitespace.

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/util/nodeselection/SelectionHistoryManager.java
SimonScholz: Ok in this case you're right, but when working with listeners this is the reliable way to ensure that everything will work as expected. Therefore we kind of always use this approach.

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/util/nodeselection/NodeSelection.java
```diff
@@ -179,6 +211,10 @@ public static NodeSelection empty() {
      * @return the new instance
      */
     public static NodeSelection from(IStructuredSelection selection) {
+        if (selection instanceof NodeSelection) {
```
donat: If the argument `selection` is a `NodeSelection`, then the method brakes it's contract and doesn't create a new instance. I'd remove the if statement.

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/view/task/SelectionDependentHandler.java
```diff
@@ -28,29 +32,30 @@
...
-        } else {
-            enabled = false;
+        if (evaluationContext instanceof IEvaluationContext) {
+            IEvaluationContext evalContext = (IEvaluationContext) evaluationContext;
+            // getting the current ISelection from the IEvaluationContext
+            // See https://wiki.eclipse.org/Command_Core_Expressions
+            Object variable = evalContext.getVariable(ISources.ACTIVE_CURRENT_SELECTION_NAME);
+            if (variable instanceof ISelection) {
+                setBaseEnabled(isEnabledFor(NodeSelection.from((IStructuredSelection) variable)));
+            }
```
donat: Enablement not explicitly set in all cases.
Replace with
```
boolean enabled = false;
if (evaluationContext instanceof IEvaluationContext) {
    IEvaluationContext evalContext = (IEvaluationContext) evaluationContext;
    // getting the current ISelection from the IEvaluationContext
    // See https://wiki.eclipse.org/Command_Core_Expressions
    Object variable = evalContext.getVariable(ISources.ACTIVE_CURRENT_SELECTION_NAME);
    if (variable instanceof ISelection) {
        enabled = isEnabledFor(NodeSelection.from((IStructuredSelection) variable)));
    }
}
setBaseEnabled(enabled)
```

donat: Use `if (variable instanceof NodeSelection)` statement and remove the `NodeSelection.from()` invocation.

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/view/task/SelectionDependentHandler.java
```diff
@@ -28,29 +32,30 @@
...
+        if (taskView != null) {
+            NodeSelection selectionHistory = taskView.getSelection();
+            if (!isEnabledFor(selectionHistory)) {
+                throw new IllegalStateException(
+                        String.format("Cannot execute command '%s' for current selection.", getCommandName(event)));
+            }
+            return selectionHistory;
+        } else {
+            ISelection currentSelection = HandlerUtil.getCurrentSelection(event);
+            return NodeSelection.from(currentSelection);
```
donat: The whole method should contain only the two lines above. The `getTaskView()`, `getTaskView(ExecutionEvent)` and the `getCommandName(ExecutionEvent)` is no longer used and can be deleted.

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/util/nodeselection/SelectionHistoryManager.java
donat: Please add a comment that we run potentially untrusted code here (maybe from external plug-ins), hence we protect the code execution by using SafeRunner

PR #181, opened 2015-10-20
Add initial WebTools story for web app dir usage

Ian-Stewart-Binks:
Initial story for WTP integration. The purpose of this story is to make use of the web application directory, which is supposed to be exposed by a new model in Gradle's Tooling Api. Not meant for merge, meant for discussion.

I've written a story for the web application model in the Tooling Api (minor adjustments still need to be made before a PR), however I am now waiting to sign Gradle's CLA.

### DIFF
--- a/docs/stories/WebTools.md
+++ b/docs/stories/WebTools.md
@@ -1,0 +1,53 @@
+# Web Tools
+
+## Configure web application directory in WTP configurations for web application projects
+
+### Motivation
+The `ide-integration.md` document in Gradle core specifies a story ["Expose web application directory"](...). The goal is to define an initial end-to-end framework for Buidship/WTP integration, and to
+communicate the location of the web application directory to WTP.
+
+### Estimate
+
+- ? day(s)
+
+### Implementation
+- Create branch in buildship repository `wtp-integration`.
+- Define `org.eclipse.buildship.wtp` plugin in `wtp-integration`.
+- Define `org.eclipse.buildship.wtp.test` plugin in `wtp-integration`.
+- Wait for corresponding story in TAPI to be completed.
+- Upgrade Tooling Commons (if necessary).
+- Define configuration class in `org.eclipse.buildship.wtp`, either configurator class, or a similar class to `DefaultWorkspaceGradleOperations`.
+    - Retrieve Web application model from TAPI
+    - If a project is a web application project (model will be `null` if project doesn't apply `war` plugin)
+        - Add Java and Dynamic Web Facets to project, which will generate component file
+        - Configure web app dir location in component file
+    - Otherwise, do nothing
+- Invoke WTP configuration from Buildship core
+
+### Test cases
+- Configuration only invoked on web application projects
+- Java facet is added to project
+- Dynamic Web Facet is added to project
+- If Java Facet already on project but has incorrect version, version is updated to correct version based on source Java version
+- If Dynamic Web Facet already on project but has incorrect version, version is updated to correct version
+- Default Java Facet version is x.
+- Default Dynamic Web Facet version is x.
+- If Java Facet exists on project with correct version, Facet is not affected.
+- If Dynamic Web Facet exists on project with correct version, Facet is not affected.
+- Custom web application directory location is communicated successfully to component file, file exists.
+- Default web application directory location is communicated successfully to component file, file exists.
+- Custom web application directory location is communicated successfuly to component file, file does not exist.
+- Default web application directory location is communicated successfully to component file, file does not exist.
+
+### Open Questions
+
+- Buildship is now a default plugin of Eclipse. WTP contains a lot of dependencies, which should not be dragged into the Java distribution.
+
+- What should the default facet versions be? The Dynamic Web Facet version does have constraints based on which source Java version is used by the project.
+
+- The Java Facet relies on information about the source/target Java versions. Should this story be blocked by the [Java source level story](https://github.com/eclipse/buildship/blob/master/docs/stories/ToolingAPI.md#set-source-level-for-java-projects)?
+
+- What changes need to be made to the tooling-commons?
+
+- How should buildship core call the WTP configuration plugin?
+
--- a/docs/stories/WebTools.md
+++ b/docs/stories/WebTools.md
@@ -11,11 +11,11 @@ communicate the location of the web application directory to WTP.
 - ? day(s)
 
 ### Implementation
-- Create branch in buildship repository `wtp-integration`.
-- Define `org.eclipse.buildship.wtp` plugin in `wtp-integration`.
-- Define `org.eclipse.buildship.wtp.test` plugin in `wtp-integration`.
-- Wait for corresponding story in TAPI to be completed.
-- Upgrade Tooling Commons (if necessary).
+- Create branch in buildship repository `wtp-integration`
+- Define `org.eclipse.buildship.wtp` plugin in `wtp-integration`
+- Define `org.eclipse.buildship.wtp.test` plugin in `wtp-integration`
+- Wait for corresponding story in TAPI to be completed
+- Upgrade Tooling Commons (if necessary)
 - Define configuration class in `org.eclipse.buildship.wtp`, either configurator class, or a similar class to `DefaultWorkspaceGradleOperations`.
     - Retrieve Web application model from TAPI
     - If a project is a web application project (model will be `null` if project doesn't apply `war` plugin)
@@ -30,14 +30,15 @@ communicate the location of the web application directory to WTP.
 - Dynamic Web Facet is added to project
 - If Java Facet already on project but has incorrect version, version is updated to correct version based on source Java version
 - If Dynamic Web Facet already on project but has incorrect version, version is updated to correct version
-- Default Java Facet version is x.
-- Default Dynamic Web Facet version is x.
+- Default Dynamic Web Facet version is 2.5.
 - If Java Facet exists on project with correct version, Facet is not affected.
 - If Dynamic Web Facet exists on project with correct version, Facet is not affected.
+- Dynamic Web Facet version inferred from _web.xml_ file.
 - Custom web application directory location is communicated successfully to component file, file exists.
 - Default web application directory location is communicated successfully to component file, file exists.
-- Custom web application directory location is communicated successfuly to component file, file does not exist.
-- Default web application directory location is communicated successfully to component file, file does not exist.
+- Custom web application directory location is communicated successfuly to component file, file does not exist
+- Default web application directory location is communicated successfully to component file, file does not exist
+- Existing `wb-resource` entries in component file not overwritten. 
 
 ### Open Questions
 
@@ -51,3 +52,7 @@ communicate the location of the web application directory to WTP.
 
 - How should buildship core call the WTP configuration plugin?
 
+### Notes
+
+The default Dynamic Web Facet version is 2.5, as it allows for a missing web.xml file, and doens't have a strong constraint on the Java version (
+at least Java 5).
--- a/docs/stories/WebTools.md
+++ b/docs/stories/WebTools.md
@@ -1,58 +1,161 @@
 # Web Tools
 
-## Configure web application directory in WTP configurations for web application projects
+# Provide Web Tools Integration for Buildship
+
+## Higher-level goals
+- Configure web application directory for deployment
+
+    Purpose: The web application directory contains all main web application resources, excluding compiled Java classes and non-compiled resources necessary for those Java classes to run. This folder contains resources such as web pages, images, JavaScript and CSS assets. Upon deployment, this folder is considered the default root. For example, if the ‘myapp’ project is deployed as myapp.war, then the src/main/webapp folder would be deployed at the root of the archive.
+
+- Add Java and Dynamic Web Facet to project
+
+    Purpose:
+
+    _Java Facet_: The Java facet allows the user to more easily set a project’s Java version, and also acts as a version constraint for other facets. For instance, the Dynamic Web facet with version 3.0 requires a Java facet with a version of at least 1.7. The Java Facet is necessary, as the Dynamic Web Facet depends on the existence of the Java facet.
+    
+    _Dynamic Web Facet_: The Dynamic Web Facet is necessary, as it is required by server adapters. Without the Dynamic Web facet, it wouldn’t be possible to deploy a project. The Dynamic Web Facet adds a /WebContent web application directory by default. All the deployment configuration is configured under .settings/org.eclipse.wst.common.component. The Web Facet enables different Web validators on the project. Additionally the Javascript Facet is automatically added along the Web Facet, which in turn enables javascript validators.
+
+- Add Utility Facet to deployable project dependencies
+
+    Purpose: The Utility facet indicates to a server adapter that the project can be deployed as a jar. Project dependencies that are listed in a web application that do not have the utility facet cannot be deployed.
+
+- Customize the Gradle container so that each classpath entry deploys or not, as defined by the gradle
+
+    Purpose: Certain dependencies shouldn’t be deployed, such as test dependencies and provided dependencies. The classpath entries, and classpath container entries, need to be marked as deployable or non-deployable. During deployment, if the dependency is an archive, then it is copied into the runtime-path of the deployment area. If it’s a project, an archive will be assembled, depending on the .component configuration of that project, then copied over the runtime path.
+
+## Story - Define internal project configuration API
 
 ### Motivation
-The `ide-integration.md` document in Gradle core specifies a story ["Expose web application directory"](...). The goal is to define an initial end-to-end framework for Buidship/WTP integration, and to
-communicate the location of the web application directory to WTP.
 
-### Estimate
+Ad hoc functionality will be defined by plugins separate from Buildship core, but will need to be invoked by Buildship core.
 
-- ? day(s)
+### The (internal) API
+
+The `IProjectConfigurator` will be the interface that is implemented by all configurators that add an extension to the extension point.
+- canConfigure determines whether the configurator should be applied
+- configure invokes the configurators configuration routine
+- getWeight returns this configurators weight. Configurators may have prerequisite configurators, and as such Buildship needs to know the order to apply these configurators. (lower weight = greater priority)
+
+```
+interface IProjectConfigurator {
+    boolean canConfigure(ProjectConfigurationRequest);
+    IStatus configure(ProjectConfigurationRequest);
+    int getWeight();
+}
+```
+
+The `ProjectConfigurationRequest` class contains all information necessary for project configuration. The purpose of having this class is to make the api more extensible.
+
+```
+class ProjectConfigurationRequest {
+    IProject getWorkspaceProject();
+    OmniEclipseProject getProject();
+    WorkspaceOperations getWorkspaceOperations();
+}
+```
 
 ### Implementation
-- Create branch in buildship repository `wtp-integration`
-- Define `org.eclipse.buildship.wtp` plugin in `wtp-integration`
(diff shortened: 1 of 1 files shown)


docs/stories/WebTools.md
```diff
@@ -0,0 +1,53 @@
...
+- Define `org.eclipse.buildship.wtp.test` plugin in `wtp-integration`.
+- Wait for corresponding story in TAPI to be completed.
+- Upgrade Tooling Commons (if necessary).
+- Define configuration class in `org.eclipse.buildship.wtp`, either configurator class, or a similar class to `DefaultWorkspaceGradleOperations`.
+    - Retrieve Web application model from TAPI
+    - If a project is a web application project (model will be `null` if project doesn't apply `war` plugin)
+        - Add Java and Dynamic Web Facets to project, which will generate component file
+        - Configure web app dir location in component file
+    - Otherwise, do nothing
+- Invoke WTP configuration from Buildship core
```
fbricon: Invoke WTP configuration from Buildship *WTP*

core mustn't have any dependency to WTP

docs/stories/WebTools.md
```diff
@@ -0,0 +1,53 @@
...
+
+### Test cases
+- Configuration only invoked on web application projects
+- Java facet is added to project
+- Dynamic Web Facet is added to project
+- If Java Facet already on project but has incorrect version, version is updated to correct version based on source Java version
+- If Dynamic Web Facet already on project but has incorrect version, version is updated to correct version
+- Default Java Facet version is x.
```
fbricon: x should be inferred from current Java settings applied to the project

docs/stories/WebTools.md
```diff
@@ -0,0 +1,53 @@
...
+- Default Dynamic Web Facet version is x.
```
fbricon: I'd say Facet 2.5, as it allows for missing web.xml, doesn't have a strong constraint on the Java version.

docs/stories/WebTools.md
```diff
@@ -0,0 +1,53 @@
...
+        - Configure web app dir location in component file
+    - Otherwise, do nothing
+- Invoke WTP configuration from Buildship core
+
+### Test cases
+- Configuration only invoked on web application projects
+- Java facet is added to project
+- Dynamic Web Facet is added to project
+- If Java Facet already on project but has incorrect version, version is updated to correct version based on source Java version
+- If Dynamic Web Facet already on project but has incorrect version, version is updated to correct version
```
fbricon: Web Facet version inferred from web.xml.
If inferred version > installed version, then update version, else do nothing (WTP doesn't support downgrading versions)

docs/stories/WebTools.md
```diff
@@ -0,0 +1,53 @@
...
+- Default Java Facet version is x.
+- Default Dynamic Web Facet version is x.
+- If Java Facet exists on project with correct version, Facet is not affected.
+- If Dynamic Web Facet exists on project with correct version, Facet is not affected.
+- Custom web application directory location is communicated successfully to component file, file exists.
+- Default web application directory location is communicated successfully to component file, file exists.
+- Custom web application directory location is communicated successfuly to component file, file does not exist.
+- Default web application directory location is communicated successfully to component file, file does not exist.
+
```
fbricon: If existing default webapp dir (in .component) is different from the webapp dir from build.gradle, it should be updated

docs/stories/WebTools.md
```diff
@@ -0,0 +1,53 @@
...
+- Buildship is now a default plugin of Eclipse. WTP contains a lot of dependencies, which should not be dragged into the Java distribution.
+
+- What should the default facet versions be? The Dynamic Web Facet version does have constraints based on which source Java version is used by the project.
+
+- The Java Facet relies on information about the source/target Java versions. Should this story be blocked by the [Java source level story](https://github.com/eclipse/buildship/blob/master/docs/stories/ToolingAPI.md#set-source-level-for-java-projects)?
+
+- What changes need to be made to the tooling-commons?
+
+- How should buildship core call the WTP configuration plugin?
```
fbricon: I vote for extension point / configurator interface, marked as *internal* (i.e. NOT API), contributed/implemented by Buildship WTP.

docs/stories/WebTools.md
Ian-Stewart-Binks: Ah, yes. No default.

docs/stories/WebTools.md
Ian-Stewart-Binks: Ah, so that other attributes of an entry are persisted? What are other attributes that can be found in the same entry as the web app dir?

docs/stories/WebTools.md
Ian-Stewart-Binks: > Web Facet version inferred from web.xml.

Where abouts from the web.xml is the facet version inferred? From the version attribute?
e.g.: `<web-app xmlns="http://java.sun.com/xml/ns/javaee" version="2.5">`

Is this something that would be good to have defined in the web application model provided by the TAPI?

fbricon: It's implemented in the https://github.com/Ian-Stewart-Binks/buildship-wtp/blob/master/org.eclipse.buildship.wtp.core/src/main/java/org/eclipse/buildship/wtp/core/configurator/WebApplicationConfigurator.java#L188

TAPI *could* do the same in theory. I'll let @donat decide whether this is relevant to add it there or not.

docs/stories/WebTools.md
fbricon: main webappdir can be identified by tag="defaultRootSource" in .component. Other external resources could be deployed as well. The problem is you don't know if those other entries were adde manually or were added by previous buildship configuration refreshes. For instance m2e-wtp used to delete everything and re-add all published directories there. But it turns out some people want to always add extra stuff in their deployment assembly page, and keep that over mutiple config refreshes.

docs/stories/WebTools.md
Ian-Stewart-Binks: > It's implemented in the https://github.com/Ian-Stewart-Binks/buildship-wtp/blob/master/org.eclipse.buildship.wtp.core/src/main/java/org/eclipse/buildship/wtp/core/configurator/WebApplicationConfigurator.java#L188

Yes, just trying to get what was actually going on rather than the implementation. That bit was more copy + paste from m2e.wtp when I asked the question last time, and I want to make sure I know what I'm doing for this rather than how to do it.

docs/stories/WebTools.md
Ian-Stewart-Binks: > But it turns out some people want to always add extra stuff in their deployment assembly page, and keep that over mutiple config refreshes.

Is this something that we should also try to do? In that case, it would be beneficial to have a test case to make sure that a previous .component file isn't overwritten. However, I'm not certain that the web application directory entry is the best way to test this, given that the only extra attribute that you've specified is the `tag` attribute.

For instance, if we want to test that the previous version of the webappdir entry is persisted, and say Gradle specifies `src/main/ian` as the webAppDir, then we would be testing for 
```
<wb-resource deploy-path="/" source-path="/src/main/webapp" tag="defaultRootSource"/>
```
to be changed to
```
<wb-resource deploy-path="/" source-path="/src/main/ian" tag="defaultRootSource"/>
```
but this doesn't test whether the .component file is updated.

Please correct me if I am wrong.

This does bring up another test though, the default webAppDir name shouldn't be propagated to the .component file if the .component file specifies a web app dir. However, the Gradle model is the truth of the build, so should we propagate the default value then?

Is there any case where the web application directory is not specified in the component file? Would this be an error, or would we add the default webAppDir in?

fbricon: you do want to see /src/main/webapp changed into /src/main/ian if the build.gradle file changed its webappdirname.  build.gradle is the truth. tag=defaultRootSource should _always_ match gradle truth. Even if no entries existed before in .component.
But if other entries exist in that file, I think they should remain there.

Ian-Stewart-Binks: My questions may have been a bit convoluted.

> you do want to see /src/main/webapp changed into /src/main/ian if the build.gradle file changed its webappdirname. build.gradle is the truth. tag=defaultRootSource should always match gradle truth. Even if no entries existed before in .component.
But if other entries exist in that file, I think they should remain there.

Okay, I'll create a test accordingly. My example was an example of a different test, but was meant to show that such a test cannot test whether the WTP plugin persists entries in the component file.

If the build.gradle file does not define a webAppDir location, but the component file of an imported project does, is the web app dir entry in the component file overwritten with the default webAppDir location taken from the Gradle model?

What should happen if a project is imported that doesn't have a web app dir defined in the component file?

fbricon: TAPI model will *always* return a webAppDir location, the default value being src/main/webapp. Since it'll always have a value, it'll always overwrite the defaultRootSource of .component, even more so if an existing .component as no source defined

Ian-Stewart-Binks: Yes, okay. Thanks. Just wanted to make sure.

docs/stories/WebTools.md
Ian-Stewart-Binks: Right, but I meant this to be something along the lines of an internal configurator API. Something that buildship core does such that the WTP configuration is invoked in some way, unless this is being left for a later story. Once someone has the WTP plugin downloaded, core needs some way to invoke the WTP configuration is some way, unless we register some sort of listener that listens for Buildship core to be applied, which is something I don't think we intend to do.

docs/stories/WebTools.md
Ian-Stewart-Binks: Just reading this over from yesterday, have a quick question:
> But if other entries exist in that file, I think they should remain there.

What other entries do we want to persist?

fbricon: so the war plugin docs shows 

war {
    from 'src/rootContent' // adds a file-set to the root of the archive
    webInf { from 'src/additionalWebInf' } // adds a file-set to the WEB-INF dir.
    classpath fileTree('additionalLibs') // adds a file-set to the WEB-INF/lib dir.
    classpath configurations.moreLibs // adds a configuration to the WEB-INF/lib dir.
    webXml = file('src/someWeb.xml') // copies a file to WEB-INF/web.xml
}

so content from `from` and `webinf` should be added as entries in the component file. unfortunately, we have no way, between 2 project refreshes, to know if the entries in .component were added manually via the Deployment Assembly page in eclipse (which you shouldn't delete), or via a previous `from` / `webinf` configuration (which you should delete). So in doubt, leave everything there.

Ian-Stewart-Binks: Ah, okay. Thanks.

I'm not certain that the details of the `from` and `webinf` are necessarily related to the webAppDir story, but I can create a testcase that tests whether previous arbitrary `wb-resource` entries persist such that we don't implement this in a way that we would later need to refactor. Sound good?

fbricon: Ok

On Wed, Oct 21, 2015 at 11:18 AM, Ian Stewart-Binks <
notifications@github.com> wrote:

> In docs/stories/WebTools.md
> <https://github.com/eclipse/buildship/pull/181#discussion_r42637532>:
>
> > +
> > +### Test cases
> > +- Configuration only invoked on web application projects
> > +- Java facet is added to project
> > +- Dynamic Web Facet is added to project
> > +- If Java Facet already on project but has incorrect version, version is updated to correct version based on source Java version
> > +- If Dynamic Web Facet already on project but has incorrect version, version is updated to correct version
> > +- Default Java Facet version is x.
> > +- Default Dynamic Web Facet version is x.
> > +- If Java Facet exists on project with correct version, Facet is not affected.
> > +- If Dynamic Web Facet exists on project with correct version, Facet is not affected.
> > +- Custom web application directory location is communicated successfully to component file, file exists.
> > +- Default web application directory location is communicated successfully to component file, file exists.
> > +- Custom web application directory location is communicated successfuly to component file, file does not exist.
> > +- Default web application directory location is communicated successfully to component file, file does not exist.
> > +
>
> Ah, okay. Thanks.
>
> I'm not certain that the details of the from and webinf are necessarily
> related to the webAppDir story, but I can create a testcase that tests
> whether previous arbitrary wb-resource entries persist such that we don't
> implement this in a way that we would later need to refactor. Sound good?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/eclipse/buildship/pull/181/files#r42637532>.
>



-- 
"Have you tried turning it off and on again" - The IT Crowd
And if that fails, then http://goo.gl/tnBgH5

Ian-Stewart-Binks: > so content from from and webinf should be added as entries in the component file. unfortunately, we have no way, between 2 project refreshes, to know if the entries in .component were added manually via the Deployment Assembly page in eclipse (which you shouldn't delete), or via a previous from / webinf configuration (which you should delete). So in doubt, leave everything there.

Two more questions:
1. Could this be solved by adding some sort of 'from gradle model' attribute to the `wb-resource` entry?
2. Upon multiple imports, wouldn't the same `wb-resource`s be continuously added?

fbricon: 1) AFAIK, the component model doesn't support adding attributes to wb-resource, I remember I wanted to do something similar. But you can double-check.

There's an alternative though, but I find it pretty ugly. You could add some other metadata to the .component, like a list of gradle-controlled paths: <property name="gradle.webresources" value="path1;path2;path3"/>

2) wb-resources would only be added if they're missing from the file

Ian-Stewart-Binks: > 1) AFAIK, the component model doesn't support adding attributes to wb-resource, I remember I wanted to do something similar. But you can double-check.

Do you mean that the component file would break if we added attributes to the `wb-resource` entries, or just that WTP offers no way to do this?

> There's an alternative though, but I find it pretty ugly. You could add some other metadata to the .component, like a list of gradle-controlled paths:

How do you mean 'gradle-controlled paths'?

> 2) wb-resources would only be added if they're missing from the file
Yes, what was I thinking.

docs/stories/WebTools.md
```diff
@@ -0,0 +1,58 @@
+# Web Tools
+
+## Configure web application directory in WTP configurations for web application projects
+
+### Motivation
+The `ide-integration.md` document in Gradle core specifies a story ["Expose web application directory"](...). The goal is to define an initial end-to-end framework for Buidship/WTP integration, and to
+communicate the location of the web application directory to WTP.
+
+### Estimate
```
donat: There's no need for an estimate in this case, just remove this section.

docs/stories/WebTools.md
donat: I don't know the details, but my gut feeling is that every necessary information should come from the Tooling API model.

docs/stories/WebTools.md
donat: I also vote for this. With the extension point we can use Buildship with and without the wtp feature, and loose coupling/lazy plugin loading comes for free.

donat:
One general comment about the this document: I recommend to split the story into several smaller ones. 
There should be a big feature title: "Provide Web Tools Integration for Buildship", explaining the intent of the work and how the smaller stories contribute into it. Then the individual stories come. Something like this:
- Story - Expose wtp model from the Tooling API
- Story - Define project configuration API in Buildship project synchronization
- Story - Define wtp plugins and features
- Story - Implement configurator for wtp projects
There are plenty of examples in the https://github.com/gradle/gradle/tree/master/design-docs folder
One good example would be this one: https://github.com/gradle/gradle/blob/master/design-docs/testing-toolkit.md

docs/stories/WebTools.md
Ian-Stewart-Binks: @donat sounds good. I'll write all documentation with this in mind.

Ian-Stewart-Binks:
> One general comment about the this document: I recommend to split the story into several smaller ones. 
There should be a big feature title: "Provide Web Tools Integration for Buildship", explaining the intent of the work and how the smaller stories contribute into it. Then the individual stories come. Something like this:

Story - Expose wtp model from the Tooling API
Story - Define project configuration API in Buildship project synchronization
Story - Define wtp plugins and features
Story - Implement configurator for wtp projects There are plenty of examples in the https://github.com/gradle/gradle/tree/master/design-docs folder One good example would be this one: https://github.com/gradle/gradle/blob/master/design-docs/testing-toolkit.md

I absolutely agree. I'll start doing this now.

docs/stories/WebTools.md
Ian-Stewart-Binks: @fbricon 
> How do you mean 'gradle-controlled paths'?
You mean just a list of arbitrary entries that contain the same path as the `wb-resource` entries?

> 1) AFAIK, the component model doesn't support adding attributes to wb-resource, I remember I wanted to do something similar. But you can double-check.

Do you mean that the component file would break if we added attributes to the wb-resource entries, or just that WTP offers no way to do this?

Ian-Stewart-Binks:
@donat I've made changes based on your input. Can you take a look and let me know what you think? Thank you.

docs/stories/WebTools.md
fbricon: gradle-controlled paths = list of paths inferred from build.gradle, that were added as wb-resource entries during the previous configuration round

there's no API to fine tune wb-resource entries and add extra metadata. Writing to .component from outside the API is a no-go, would lead to synchronization issues. There's also a very good chance every change made to custom wb-resource attributes would be deleted when the user interacts with the Deployment Assembly page.

Ian-Stewart-Binks: > gradle-controlled paths = list of paths inferred from build.gradle, that were added as wb-resource entries during the previous configuration round

Yes, but how do you suggest we would store this list?

> there's no API to fine tune wb-resource entries and add extra metadata. Writing to .component from outside the API is a no-go, would lead to synchronization issues. There's also a very good chance every change made to custom wb-resource attributes would be deleted when the user interacts with the Deployment Assembly page.

Ah, yes. That's not so good.

fbricon: <property name="gradle.derived.paths" value="/src/main/webapp,/src/rootContent/"/>

Ian-Stewart-Binks: Alright, cool. That's what I suspected. Thanks.

If it's ugly, but it's the only way to implement the functionality we want, then I don't see any reason not to do it. The question is whether or not we want that functionality.

fbricon: I'm fine if you implement it. Results will certainly be more correct than
if you don't. And I might steal your code to do something similar to
m2e-wtp when I find the time.

On Thu, Oct 22, 2015 at 3:52 PM, Ian Stewart-Binks <notifications@github.com
> wrote:

> In docs/stories/WebTools.md
> <https://github.com/eclipse/buildship/pull/181#discussion_r42797215>:
>
> > +
> > +### Test cases
> > +- Configuration only invoked on web application projects
> > +- Java facet is added to project
> > +- Dynamic Web Facet is added to project
> > +- If Java Facet already on project but has incorrect version, version is updated to correct version based on source Java version
> > +- If Dynamic Web Facet already on project but has incorrect version, version is updated to correct version
> > +- Default Java Facet version is x.
> > +- Default Dynamic Web Facet version is x.
> > +- If Java Facet exists on project with correct version, Facet is not affected.
> > +- If Dynamic Web Facet exists on project with correct version, Facet is not affected.
> > +- Custom web application directory location is communicated successfully to component file, file exists.
> > +- Default web application directory location is communicated successfully to component file, file exists.
> > +- Custom web application directory location is communicated successfuly to component file, file does not exist.
> > +- Default web application directory location is communicated successfully to component file, file does not exist.
> > +
>
> Alright, cool. That's what I suspected. Thanks.
>
> If it's ugly, but it's the only way to implement the functionality we
> want, then I don't see any reason not to do it. The question is whether or
> not we want that functionality.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/eclipse/buildship/pull/181/files#r42797215>.
>



-- 
"Have you tried turning it off and on again" - The IT Crowd
And if that fails, then http://goo.gl/tnBgH5

PR #183, opened 2016-01-06
set local m2 repository to temp directory

breskeby:
This avoids polluting the local m2 folder with large junks of temporary data.

Especially used to avoid in the ci invironment to keep ~/.m2 folder clean.
incremental builds are not affected at all as the stuff that is (accidently) pushed to ~/m2 is neither input nor output of any gradle task. pushing files there is just a known issue of the underlaying ant.deploy task used by the installTargetPlatform gradle task


donat:
The CLA is in fact signed. Cheery-picked locally, added signed-off-by footer and pushed to master.

PR #185, opened 2016-01-22
Use eclipse preference color and font for gradle consoles

Mumfrey:
Currently [GradleConsole](https://github.com/eclipse/buildship/blob/master/org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/console/GradleConsole.java) does not use the eclipse preference colour or preference font for consoles. However, because it **does** use the preference *foreground* colour this leads to invisible (or difficult to read) text if a "reversed" colour scheme is used. This PR alters the behaviour of `GradleConsole` to apply the font and background colour selections defined in the eclipse preferences.

Whilst this is admittedly a minor change, the effects are quite debilitating since it requires highlighting the text in the console window in order to read it. Buildship is fantastic but this minor tweak will make a huge difference to those of us foolish enough to use dark console colour schemes.

##### Console output with default settings

Gradle console output with default settings

##### Console colour preferences changed to a reversed colour scheme

Preferences dialogue showing reversed colours

##### produces the following output:

Gradle console output with reversed colour scheme

##### After applying the changes in this PR: colour and font selection is applied

Gradle console output after fix


oehme:
Thank you for the fix

PR #187, opened 2016-03-23
Update manifest versions to allow projects to be open in Eclipse SDK

gamerson:
We often have buildship bundle projects open in Eclipse and also we have downstream bundles that depend on specific versions of buildship bundles (v1.0.11), and since all of the buildship bundle manfiests are still set at 1.0.0.qualifier, Eclipse marks my bundle manifests with an error.  

Would the buildship team be able to keep the manifest versions up-to-date?

### DIFF
--- a/org.eclipse.buildship.branding/META-INF/MANIFEST.MF
+++ b/org.eclipse.buildship.branding/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: Buildship, Eclipse Plug-ins for Gradle
 Bundle-SymbolicName: org.eclipse.buildship.branding
-Bundle-Version: 1.0.0.qualifier
+Bundle-Version: 1.0.11.qualifier
 Bundle-Vendor: Eclipse Buildship
 Bundle-RequiredExecutionEnvironment: JavaSE-1.6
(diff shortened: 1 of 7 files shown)


donat:
Your change request make sense to me, I'll merge it in a second.
In fact we want to integrate the automatic version increment into our release infrastructure. This task should include updating the versions in the bundle manifest as well.

donat:
@gamerson I updated our build scripts such that the bundle version is also updated when the `incrementVersion` task is executed. This task is called automatically by our CI server when a new release comes out.

https://github.com/eclipse/buildship/blob/master/build.gradle#L368-L373

gamerson:
Very nice, thanks for the update.

On Fri, Mar 25, 2016 at 9:20 AM, Donát Csikós <notifications@github.com>
wrote:

> @gamerson <https://github.com/gamerson> I updated our build scripts such
> that the bundle version is also updated when the incrementVersion task is
> executed. This task is called automatically by our CI server when a new
> release comes out.
>
> https://github.com/eclipse/buildship/blob/master/build.gradle#L368-L373
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/eclipse/buildship/pull/187#issuecomment-201305278>
>



-- 
Greg Amerson
Liferay Developer Tools
Liferay, Inc. www.liferay.com

PR #188, opened 2016-04-28
Import the stsmigration plugin during oomph setup

Bananeweizen:
Since that plugin is referenced in the feature, it is mandatory and should be
in the workspace to avoid feature related warnings and build errors.

I have an Eclipse CLA.

Change-Id: Ie280721b66d8b7289dc0cb13b65a511355a4c523
Signed-off-by: Michael Keppler <Michael.Keppler@gmx.de>


donat:
This makes perfect sense, I'll merge it in a second. 

PR #189, opened 2016-05-06
As a user, I'd like to press F5 for refresh project

jtydhr88:


donat:
Some could argue that this feature is annoying, as the F5 key is usually associated to the resource refresh command. Fortunately you can configure this locally. In Eclipse go to Preferences > Keys, search for  'Refresh Gradle Project',  specify F5 in the key binding textbox and you are done.

lukeu:
As a (involuntary) m2e user, I've got used to using Alt+F5 for this. So I've thought that might be a nice shortcut to set up by default for people migrating over to Gradle. The only problem is that keybindings don't be able to be set such that it can be used for both Maven and Gradle projects, which is a shame since projects would generally be one or the other, not both.

I don't suppose there's any scope for adding some other 'category' of keybinding such that the 2 shortcuts using the same keypress could co-exist?

PR #190, opened 2016-05-13
Fix plugin version attributes in feature.xml

gamerson:


gamerson:
Added: https://bugs.eclipse.org/bugs/show_bug.cgi?id=493688

PR #192, opened 2016-06-02
Allow Java Platform to be first in classpath

jsarman:
Change order of classpath building to allow java platform to load before external dependencies.


donat:
First of all thank you for contribution! The change looks simple. But before we merge it would you mind giving us some context why is this change needed, perhaps with a sample project exhibiting the problem?
Also, please note that the ip-validation has failed for your project, probably due to a missing [signed-off-by entry in the commit message](https://wiki.eclipse.org/Development_Resources/Contributing_via_Git#via_GitHub).

donat:
Ah, I just found [Bug 495328](https://bugs.eclipse.org/bugs/show_bug.cgi?id=495328). I suppose that is related to this pull request.

jsarman:
Yeah I worked with Aaron to verify and create  Bug 495328.  I was able to setup buildship and debug it to the simple code order change.  I made the change with the github built in editor and I guess by doing it that way  github does not sign the code.  I can go through all the steps to ensure its signed off if needed.

oehme:
I think sign-off needs to be the last statement in a commit message.

Also, please include an integration test that demonstrates this change works as expected.

oehme:
Thank you for the fix! I will backport it to 1.0.16 too.

PR #194, opened 2016-11-18
Simplify project synchronization

oehme:
This change ensures that the model in the workspace actually matches the Gradle model. Any manual changes to the .project/.classpath file are overwritten if the Gradle model contains information. For things where the Gradle model does not contain information, the user-defined values are kept.

This fixes a long-standing issue that Buildship couldn't handle two build commands with the same id, but different arguments.

It also removes the need for writing anything to the buildship.core preference file. Instead, we use persistent properties for the few things that we still need to remember. In the long run, we might just want to write these things in a custom file inside the workspace metadata.

I'm not too happy with how the code for the ClasspathContainerUpdater turned out. It's very efficient, but maybe having a separate loop for the JRE replacement would make it more readable. 

With all this in place, we can remove the confusing "keep/overwrite" dialog next.

I also removed the huge Java model transaction around the whole sync operation. It was creating lots of memory pressure on larger projects and is no longer necessary, as our views now behave properly even when faced with hundreds of "project created events".


org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/LinkedResourcesUpdater.java
```diff
@@ -39,30 +47,30 @@
  * Note that currently, we only include linked resources that are folders.
  */
 final class LinkedResourcesUpdater {
+    private static final QualifiedName LINKED_RESOURCES_KEY = new QualifiedName(CorePlugin.PLUGIN_ID, "linkedResources");
```
donat: Persistent properties have a size limitation which means that if a user defines lots of linked resources, Buildship will fail with `org.eclipse.core.internal.resources.ResourceException: Could not set property: org.eclipse.buildship.core linkedResources. Value is too long` exception.

I think we should save the existing state into a properties file at the same place where we store the classpath container cache.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/LinkedResourcesUpdater.java
```diff
@@ -73,6 +81,14 @@ private void removeOldLinkedResources(StringSetProjectProperty knownLinkedResour
         }
     }
 
+    private Collection<String> getKnownLinkedResources(IProject project) throws CoreException {
+        String serializedForm = project.getPersistentProperty(LINKED_RESOURCES_KEY);
+        if (serializedForm == null) {
+            return Collections.emptyList();
+        }
+        return Splitter.on(File.pathSeparator).omitEmptyStrings().splitToList(serializedForm);
```
donat: Perhaps we could push this to a superclass and reuse it in `DeriverResourcesUpdater` .

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/view/task/TaskView.java
```diff
@@ -193,8 +193,8 @@ public void setContent(TaskViewContent content) {
      * Reloads the task model in the background and updates this view once the reload is complete.
      * Can be safely called outside the UI thread.
```
donat: Minor: add `@param`

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/LinkedResourcesUpdater.java
oehme: Yes good idea, though probably some utility we delegate to instead of a superclass.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/LinkedResourcesUpdater.java
oehme: Yeah I think that's a good idea and we might actually store more things there, like which folders are subprojects and which folder is the build folder. Then it doesn't matter anymore whether the build folder is actually there at project synchronization time.

oehme:
Superseded by #196

PR #195, opened 2016-11-23
Define invocationcustomizer extension point

donat:


org.eclipse.buildship.core/META-INF/MANIFEST.MF
```diff
@@ -25,6 +25,7 @@ Export-Package: org.eclipse.buildship.core;x-friends:="org.eclipse.buildship.ui"
  org.eclipse.buildship.core.configuration;x-friends:="org.eclipse.buildship.ui",
  org.eclipse.buildship.core.console;x-friends:="org.eclipse.buildship.ui",
  org.eclipse.buildship.core.event;x-friends:="org.eclipse.buildship.ui",
+ org.eclipse.buildship.core.extensions,
```
oehme: Let's find a better name for this. We don't want to throw all extensions into one package. The package name should reflect the usage. Something like `invocation` or `synchronization`.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/extensions/result/ContributionManager.java
```diff
@@ -0,0 +1,27 @@
...
+package org.eclipse.buildship.core.extensions.result;
+
+import java.util.List;
+
+/**
+ * Provides access to external plugin contributions.
+ *
+ * @author Donat Csikos
+ */
+public interface ContributionManager {
```
oehme: We don't need a new interface here. We can reuse the existing InvocationCustomizer interface and have an implementation that collects all of them from the extension point.

org.eclipse.buildship.core/META-INF/MANIFEST.MF
donat: Let's go with `org.eclipse.buildship.core.invocation`.

PR #196, opened 2016-12-04
Simplify project synchronization

donat:


org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/preferences/PersistentModel.java
```diff
@@ -0,0 +1,41 @@
...
+     *
+     * @param key the key
+     * @param value the value
+     */
+    void setValue(String key, String value);
+
+    /**
+     * Persists all changes on the disk.
+     */
+    void flush();
```
oehme: This should be on ModelPersistence

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/PersistentUpdater.java
```diff
@@ -0,0 +1,53 @@
...
+
+import org.eclipse.buildship.core.CorePlugin;
+import org.eclipse.buildship.core.preferences.PersistentModel;
+
+/**
+ * Base updater class to load and store updated item names.
+ *
+ * @author Donat Csikos
+ */
+public abstract class PersistentUpdater {
```
oehme: I think this should be a helper instead of a base class.

donat: Frankly, I like the superclass better because it states that from all the existing updaters only the subclasses persist their state.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/preferences/PersistentModel.java
donat: I chose this because this way we don't have to use `instanceof` when saving the state in `DefaultModelPersistence`.

PR #197, opened 2016-12-14
Improvements on persistent properties and on project synchronization

donat:
This pull request contains a series of changes that fixes a few loose ends in the project synchronisation mechanism. 

 The subproject and build folder filters were still using persistent properties, which in some cases [were out-of-sync  with the model](https://bugs.eclipse.org/bugs/show_bug.cgi?id=500380). Now, accessing these attributes are unified via persistent properties and the UI always show an up-to-date state. For that the `PersistentModel` interface was extended to expose structured  data (`getBuildFolder()` and `getSubProjects()`).

Frequently accessing the model persistence interface - especially with the improvements above - can be IO-heavy for large projects. To solve that, `DefaultModelPersistence` now caches values in the memory and only writes the property files when the workbench is about to shut down.
 
The `ModelPersistence` interface were adjusted to provide load, save and delete operations. The delete operation is invoked when the workspace project is uncoupled from the build.


org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/WorkbenchShutdownEvent.java
```diff
@@ -0,0 +1,22 @@
...
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+
+package org.eclipse.buildship.core.workspace;
+
+import org.eclipse.buildship.core.event.Event;
+
+/**
+ * Project shutdown event.
```
oehme: Let's not repeat the obvious in comments, as that's already captured by the class name. Also note the copy-paste error :)

Instead the javadoc should say when this event occurs (i.e. probably before the workbench is shutdown, the listener can react synchronously I guess?)

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/navigator/BuildFolderViewerFilter.java
```diff
@@ -31,7 +36,25 @@ public boolean select(Viewer viewer, Object parentElement, Object element) {
     }
 
     private boolean isBuildFolder(IResource resource) {
-        return resource instanceof IFolder && CorePlugin.workspaceOperations().isBuildFolder((IFolder) resource);
+        if (resource instanceof IFolder) {
```
oehme: I personally find the && version easier to read, but no big deal.

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/navigator/BuildFolderViewerFilter.java
```diff
@@ -31,7 +36,25 @@ public boolean select(Viewer viewer, Object parentElement, Object element) {
...
     }
 
+    public static boolean isBuildFolderInPerstentModel(IFolder folder) {
+        try {
+            IProject project = folder.getProject();
+            Optional<IPath> modelBuildDir = CorePlugin.modelPersistence().loadModel(project).getBuildDir();
+            if (!modelBuildDir.isPresent()) {
+                return false;
+            }
+            IPath relativePath = RelativePathUtils.getRelativePath(project.getFullPath(), folder.getFullPath());
```
oehme: Why don't we store the build folder as a full path, so we don't have to relativize again and again? We need to keep in mind that this is called for every folder that is ever added/removed/changed.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/preferences/internal/DefaultModelPersistence.java
```diff
@@ -20,120 +21,149 @@
...
 /**
  * Default implementation for {@link PersistentModel}.
  *
  * @author Donat Csikos
  */
-public final class DefaultModelPersistence implements ModelPersistence, EventListener  {
+public final class DefaultModelPersistence implements ModelPersistence, EventListener {
+
+    // Cache holding (project name - persistent model property) pairs.
+    private final LoadingCache<String, Map<String, String>> modelCache;
```
oehme: I think we could improve perceived performance slightly by launching a background job at startup which loads all preferences into memory.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/preferences/PersistentModel.java
```diff
@@ -8,34 +8,77 @@
...
  *
  * @author Donat Csikos
  */
 public interface PersistentModel {
 
     /**
+     * Returns the build directory.
+     * @return the build directory
+     */
+    Optional<IPath> getBuildDir();
```
donat: I've managed to change the interface such that all methods are now strongly typed.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/preferences/internal/DefaultModelPersistence.java
```diff
@@ -20,120 +21,149 @@
...
+
+        File preferencesFile = preferencesFile(event.getPreviousName());
         if (preferencesFile.exists()) {
-            try (Reader reader = new InputStreamReader(new FileInputStream(preferencesFile(project)), Charsets.UTF_8)) {
+            Files.move(preferencesFile, preferencesFile(event.getProject().getName()));
+        }
+    }
+
+    private void deleteProjectPreferences(ProjectDeletedEvent event) {
+        this.modelCache.invalidate(event.getProject().getName());
```
donat: I removed the duplication.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/preferences/internal/DefaultModelPersistence.java
```diff
@@ -20,120 +21,149 @@
...
         } else {
-            return new DefaultPersistentModel(this, project, Collections.<String, String>emptyMap());
+            return Collections.emptyMap();
         }
     }
 
-    void persistPrefs(DefaultPersistentModel preferences) {
+    private void persistAllProjectPrefs() {
+        Map<String, Map<String, String>> modelCacheMap = this.modelCache.asMap();
+        for (String projectName : modelCacheMap.keySet()) {
```
donat: Fixed.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/WorkbenchShutdownEvent.java
donat: The Javadoc is more clear now.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/preferences/internal/DefaultModelPersistence.java
donat: I've added a prefetch logic there.

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/navigator/BuildFolderViewerFilter.java
donat: Turns out it's not necessary to use `RelativePathUtils` as there's `IFolder.getProjectRelativePath()`.

donat:
@oehme Review items addressed

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/preferences/internal/DefaultPersistentModel.java
```diff
@@ -36,6 +41,9 @@
 
     private static final String PROPERTY_SUBPROJECTS = "subprojectPaths";
     private static final String PROPERTY_BUILD_DIR = "buildDir";
+    private static final String PROPERTY_CLASSPATH = "classpath";
+    private static final String PROPERTY_DERIVED_RESOURCES = "derivedResources";
+    private static final String PROPERTY_LINKED_RESOURCES = "linkedResources";
 
     private final IProject project;
     private final Map<String, String> entries;
```
oehme: I think we should only serialize/deserialize on startup/teardown and keep the results in memory otherwise. I.e. don't have a single Map field, but actually have fields for classpath, derived resources etc.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/preferences/PersistentModel.java
```diff
@@ -9,76 +9,41 @@
...
  *
  * @author Donat Csikos
  */
 public interface PersistentModel {
 
-    /**
-     * Returns the build directory.
-     * @return the build directory
-     */
     Optional<IPath> getBuildDir();
```
oehme: Personally I don't find Optional very useful as long as we're not on Java 8 and heavily using functional idioms. Also, if we ever open this up to others (which there are good use cases for), we'd have to remove the Guava types from the signatures anyway so we don't leak the dependency. Let's stick with `null` and reconsider this once our minimum version is Java 8.

oehme: Why is this an IPath instead of an IResource? Then we could just do a simple equals where it is queried.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/preferences/PersistentModel.java
```diff
@@ -8,34 +8,42 @@
...
+
+    void setSubprojectPaths(Collection<IPath> subprojectPaths);
+
+    Optional<List<IClasspathEntry>> getClasspath();
+
+    void setClasspath(List<IClasspathEntry> classpath);
+
+    Collection<IResource> getDerivedResources();
+
+    void setDerivedResources(Collection<IResource> derivedResources);
```
oehme: I think these should be IPath instances like buildDir and Subprojects. Their presence can change over time. E.g. the `.gradle` folder might not yet be there when we synchronize the project.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/DerivedResourcesUpdater.java
```diff
@@ -50,35 +70,78 @@ private void update(IProgressMonitor monitor) {
...
-        Collection<String> previouslyKnownDerivedResources = getKnownItems();
+    private void markBuildFolder(Optional<IPath> buildDirectoryPath) throws CoreException {
+        PersistentModel preferences = CorePlugin.modelPersistence().loadModel(this.project);
+        preferences.setBuildDir(buildDirectoryPath);
+        CorePlugin.modelPersistence().saveModel(preferences);
+    }
+
+    private List<IResource> getDerivedResources(Optional<IPath> possibleBuildDirectoryPath, SubMonitor progress) {
+        List<IResource> derivedResources = Lists.newArrayList();
+        IResource dotGradle = this.project.findMember(".gradle");
```
oehme: We shouldn't work with `IResource` here, but with `IPath`. The `.gradle` folder should be in the list of derived resources no matter if it currently exists or not. 

Then we can add a resource change listener which listens for new resources and marks them as derived if they are in the derived list.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/preferences/internal/DefaultPersistentModel.java
donat: Done.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/preferences/PersistentModel.java
donat: Agreed. The interface uses nulls instead of Optionals.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/preferences/PersistentModel.java
donat: Done

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/DerivedResourcesUpdater.java
donat: Fixed

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/preferences/PersistentModel.java
```diff
@@ -11,37 +11,35 @@
...
 
     void setSubprojectPaths(Collection<IPath> subprojectPaths);
 
-    Optional<List<IClasspathEntry>> getClasspath();
+    List<IClasspathEntry> getClasspath();
 
     void setClasspath(List<IClasspathEntry> classpath);
 
-    Collection<IResource> getDerivedResources();
+    Collection<IPath> getDerivedResources();
```
oehme: Does it make sense for these lists to be nullable? I don't see a semantical difference between an empty or absent list of derived resources. That would get rid of the null checks further down.

If we only saved the model at the end of synchronization, then we could guarantee that:

- either the whole model is there with valid information
- or the whole model is absent

Then we don't need individual null checks and don't need to allow null as a value.

donat: I did this to have a consistent semantics across all getters/setters defined on this interface. It makes sense for the `getClasspath()` method because if the value is not present, then we query the `EclipseProject` model.

oehme: That's why I proposed to make the whole model nullable instead of the individual entries in it. The model is either there or not there. If it is there, the entries are valid and can't be null.

donat: Ah ok, it's all clear now. Seems like a better approach :)

donat: What about `PersistentModel.getBuildDir()`. For that should it be the only method to return `null`, or should we which to `Optional`?

oehme: I don't understand the question I'm afraid. Can the build dir of a valid model ever be null?

donat: That's just me getting confused about it. No, I was wrong, the build folder is never set to null. If it's not available in the Gradle model, we just set it to `build`.

donat: What you've said in the initial comment of this thread is not entirely true. Imagine the first sync operation where the first updater receives an empty/null model. Then it sets a partial data on it. The proceeding updaters will no longer see an empty/null model, just a partially filled one. That's why - currently - we check for the nullity of an individual property.

donat: Maybe a better solution would be to extract a common `Updater` interface and pass the `PersistentModel` to the updaters individually. Anyhow, I would extract that work into a separate story.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/preferences/PersistentModel.java
```diff
@@ -11,37 +11,35 @@
...
 
-    void setDerivedResources(Collection<IResource> derivedResources);
+    void setDerivedResources(Collection<IPath> derivedResources);
 
     Collection<IFolder> getLinkedResources();
```
donat: Fixed

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/preferences/internal/DefaultModelPersistence.java
```diff
@@ -120,50 +122,44 @@ private void deleteProjectPreferences(ProjectDeletedEvent event) {
         deleteModel(event.getProject());
     }
 
-    private static Map<String, String> loadPrefs(String projectName) throws IOException, FileNotFoundException {
+    private static DefaultPersistentModel loadModel2(IProject project) throws IOException, FileNotFoundException {
```
donat: Fixed

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/preferences/internal/DefaultModelPersistence.java
```diff
@@ -120,50 +122,44 @@ private void deleteProjectPreferences(ProjectDeletedEvent event) {
...
-                HashMap<String, String> preferences = Maps.newHashMap();
-                for (Object propKey : props.keySet()) {
-                    preferences.put(propKey.toString(), props.get(propKey).toString());
-                }
-                return preferences;
+                return DefaultPersistentModel.fromProperties(project, props);
             }
         } else {
-            return Collections.emptyMap();
+            return DefaultPersistentModel.fromEmpty(project);
```
donat: Fixed

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/preferences/PersistentModel.java
oehme: Sure, feel free to do this in another story. And yes, the persistent model should be passed along the different updaters and only saved at the end of the sync operation. I'm not sure we need an Updater interface for that.

PR #202, opened 2017-01-27
Add prototype for Gradle Kotlin script support

donat:


build.gradle
```diff
@@ -240,13 +240,14 @@ subprojects {
             targetCompatibility = 1.8
         }
 
-        tasks.withType(AbstractCompile).all {
+        tasks.matching { it instanceof JavaCompile || it instanceof GroovyCompile }.all {
```
oehme: Why no stick with AbstractCompile?

build.gradle
```diff
@@ -240,13 +240,14 @@ subprojects {
...
             // enable all warnings except for different sourceCompatibility and targetCompatibility value
             options.compilerArgs << '-Xlint:all'
             options.compilerArgs << '-Xlint:-options'
-            if (targetPlatformVersion >= '46') {
-                options.compilerArgs << '-Werror'                
-            }
+// TODO (donat) kotlin compilation currently fails with bad path error
```
oehme: Kotlin uses the -Werror flag too?

org.eclipse.buildship.kotlin/build.gradle
```diff
@@ -0,0 +1,21 @@
...
+    ext.kotlin_version = '1.0.4'
+    repositories.jcenter()
+    dependencies.classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
+}
+
+apply plugin: eclipsebuild.BundlePlugin
+apply plugin: 'kotlin'
+
+dependencies {
+    compile 'org.gradle:gradle-core:3.3'
```
oehme: Should be parametrized, some as the TAPI version we use.

org.eclipse.buildship.kotlin/src/main/java/org/eclipse/buildship/kotlin/KotlinPlugin.java
```diff
@@ -0,0 +1,117 @@
...
+        appendPluginResourceIfExists(classpath, "/build/kotlin-classes/main");
+        return classpath;
+    }
+
+    private static List<String> templateClasspathFor(File projectDir) {
+        ProjectConnection connection = null;
+        try {
+            connection = GradleConnector.newConnector()
+                    .forProjectDirectory(projectDir)
+                    .useDistribution(new URI("https://repo.gradle.org/gradle/dist-snapshots/gradle-script-kotlin-3.3-20161205200654+0000-all.zip"))
```
oehme: Do we still need a special distribution to use Kotlin? I thought this worked now with the standard Gradle distros? We should use whatever is specified in the project settings.

org.eclipse.buildship.kotlin/src/main/java/org/eclipse/buildship/kotlin/KotlinPlugin.java
```diff
@@ -0,0 +1,117 @@
...
+                    .connect();
+            KotlinBuildScriptModel model = connection.getModel(KotlinBuildScriptModel.class);
```
oehme: Should respect offline mode, console listener etc. like our other model operations.

org.eclipse.buildship.kotlin/src/main/kotlin/org/eclipse/buildship/kotlin/GradleKotlinScriptDependenciesResolver.kt
```diff
@@ -0,0 +1,79 @@
...
+    private fun makeDependencies(classPath: Iterable<File>, sources: Iterable<File> = emptyList()): KotlinScriptExternalDependencies =
+        object : KotlinScriptExternalDependencies {
+            override val classpath = classPath
+            override val imports = implicitImports
+            override val sources = sources
+        }
+
+    companion object {
+        val implicitImports = listOf(
+            "org.gradle.api.plugins.*",
```
oehme: Why is this duplicated between GSK and here?

org.eclipse.buildship.kotlin/src/main/kotlin/org/eclipse/buildship/kotlin/KotlinBuildScript.kt
```diff
@@ -0,0 +1,30 @@
...
+
+import org.gradle.api.Project
+import org.gradle.api.plugins.ObjectConfigurationAction
+import org.gradle.script.lang.kotlin.KotlinScriptHandler
+import org.jetbrains.kotlin.script.ScriptTemplateDefinition
+
+@ScriptTemplateDefinition(
+    resolver = GradleKotlinScriptDependenciesResolver::class,
+    scriptFilePattern = ".*\\.gradle\\.kts")
+abstract class KotlinBuildScript(project: Project) : Project by project {
```
oehme: Why is this duplicated? This should already be in GSK

org.eclipse.buildship.kotlin/src/main/kotlin/org/eclipse/buildship/kotlin/KotlinModelQuery.kt
```diff
@@ -0,0 +1,43 @@
...
+		return withConnectionFrom(connectorFor(projectDir), gradleUserHome) {
+			model(KotlinBuildScriptModel::class.java)
+				.setJvmArguments("-D${KotlinScriptPluginFactory.modeSystemPropertyName}=${KotlinScriptPluginFactory.classPathMode}")
+                .withArguments(if(isOffline) listOf("--offline") else listOf<String>())
+				.get()
+		}
+	}
+
+	fun connectorFor(projectDir: File): GradleConnector =
+			GradleConnector.newConnector().forProjectDirectory(projectDir).useDistribution(URI("https://repo.gradle.org/gradle/dist-snapshots/gradle-script-kotlin-3.3-20161205200654+0000-all.zip"))
```
oehme: Hardcoded gradle-script-kotlin distro

org.eclipse.buildship.kotlin/src/main/java/org/eclipse/buildship/kotlin/KotlinPlugin.java
```diff
@@ -0,0 +1,117 @@
...
+        // java classes from this plugin for development time
+        appendPluginResourceIfExists(classpath, "/bin");
+        // Kotlin classes can't be loaded from the IDE as they're compiled into memory
+        // therefore we can only load them from the Gradle build folder
+        appendPluginResourceIfExists(classpath, "/build/kotlin-classes/main");
+        return classpath;
+    }
+
+    private static List<String> templateClasspathFor(File projectDir) {
+        ProjectConnection connection = null;
```
oehme: Duplicated code between here and KotlinModelQuery.

org.eclipse.buildship.kotlin/src/main/java/org/eclipse/buildship/kotlin/KotlinPlugin.java
```diff
@@ -0,0 +1,117 @@
...
+        appendPluginResourceIfExists(classpath, "/");
+        // java classes from this plugin for development time
+        appendPluginResourceIfExists(classpath, "/bin");
+        // Kotlin classes can't be loaded from the IDE as they're compiled into memory
+        // therefore we can only load them from the Gradle build folder
+        appendPluginResourceIfExists(classpath, "/build/kotlin-classes/main");
+        return classpath;
+    }
+
+    private static List<String> templateClasspathFor(File projectDir) {
```
oehme: This should not even be necessary, we should talk to the JetBrains team about this. There absolutely no need to load the template class at this point. It introduces a synchronous call which may result in synchronously downloading Gradle etc. Our ScriptTemplateProvider can already tell the Kotlin plugin if it applies and what dependency resolver to use. So there is no need to load the template in order to look for the same information on some annotation.

org.eclipse.buildship.kotlin/build.gradle
```diff
@@ -0,0 +1,21 @@
...
+    compile 'org.gradle:gradle-process-services:3.3'
```
oehme: Why do we need core and process-services?

build.gradle
donat: The `kotlin` plugin contributes `KotlinCompile` which extends `AbstractCompile` but doesn't have attributes like `options`.

build.gradle
donat: I got the following Java compilation error which I wanted to deal with later:
`warning: [path] bad path element "/Users/donat/.gradle/caches/modules-2/files-2.1/org.jetbrains.kotlin/kotlin-reflect/1.1-M04/50e442efdd3eab1dcf2ddbf4b3ec5dabecf1b2b2/kotlin-runtime.jar": no such file or directory`

org.eclipse.buildship.kotlin/build.gradle
donat: Fixed (will push the changes later).

org.eclipse.buildship.kotlin/build.gradle
donat: Some random class was needed for the compilation.

org.eclipse.buildship.kotlin/src/main/java/org/eclipse/buildship/kotlin/KotlinPlugin.java
donat: I created https://github.com/gradle/buildship/issues/153 for that.

org.eclipse.buildship.kotlin/src/main/java/org/eclipse/buildship/kotlin/KotlinPlugin.java
donat: Fixed (will push the changes later).

build.gradle
oehme: But below you commented out options.compilerArgs, which seems to contradict this.

org.eclipse.buildship.kotlin/build.gradle
oehme: Some random class? Can you be more specific? As far as I can see this should only be using the TAPI, not Gradle core.

org.eclipse.buildship.kotlin/src/main/java/org/eclipse/buildship/kotlin/KotlinPlugin.java
donat: Fixed (will push the changes later).

org.eclipse.buildship.kotlin/src/main/java/org/eclipse/buildship/kotlin/KotlinPlugin.java
donat: When I first created the prototype, this is how it was done in `gradle-script-kotlin`. It's no longer necessary, I've changed that.  (will push the changes later)

org.eclipse.buildship.kotlin/build.gradle
donat: Sorry for the sketchy answer. We need them to compile `KotlinBuildScript.kt` and `GradleKotlinScriptDependenciesResolver.kt`. These classes were copied from `gradle-script-kotlin` which depends on those libraries. Without the dependencies we get the following compile errors:
```
:org.eclipse.buildship.kotlin:compileKotlin
e: /Development/git/buildship/org.eclipse.buildship.kotlin/src/main/kotlin/org/eclipse/buildship/kotlin/KotlinBuildScript.kt: (11, 23): Unresolved reference: Project
e: /Development/git/buildship/org.eclipse.buildship.kotlin/src/main/kotlin/org/eclipse/buildship/kotlin/KotlinBuildScript.kt: (12, 23): Unresolved reference: plugins
e: /Development/git/buildship/org.eclipse.buildship.kotlin/src/main/kotlin/org/eclipse/buildship/kotlin/KotlinBuildScript.kt: (19, 43): Unresolved reference: Project
e: /Development/git/buildship/org.eclipse.buildship.kotlin/src/main/kotlin/org/eclipse/buildship/kotlin/KotlinBuildScript.kt: (19, 54): Only interfaces can be delegated to
e: /Development/git/buildship/org.eclipse.buildship.kotlin/src/main/kotlin/org/eclipse/buildship/kotlin/KotlinBuildScript.kt: (19, 54): Unresolved reference: Project
e: /Development/git/buildship/org.eclipse.buildship.kotlin/src/main/kotlin/org/eclipse/buildship/kotlin/KotlinBuildScript.kt: (28, 49): Unresolved reference: ObjectConfigurationAction
e: /Development/git/buildship/org.eclipse.buildship.kotlin/src/main/kotlin/org/eclipse/buildship/kotlin/KotlinBuildScript.kt: (29, 9): Unresolved reference: project
e: /Development/git/buildship/org.eclipse.buildship.kotlin/src/main/kotlin/org/eclipse/buildship/kotlin/KotlinBuildScript.kt: (29, 17): Public-API inline function cannot access non-public-API 'internal open fun <ERROR FUNCTION>(): [ERROR : <ERROR FUNCTION RETURN TYPE>] defined in root package'
e: /Development/git/buildship/org.eclipse.buildship.kotlin/src/main/kotlin/org/eclipse/buildship/kotlin/KotlinBuildScript.kt: (29, 25): Unresolved reference: it
e: /Development/git/buildship/org.eclipse.buildship.kotlin/src/main/kotlin/org/eclipse/buildship/kotlin/KotlinBuildScript.kt: (29, 28): Public-API inline function cannot access non-public-API 'internal open fun <ERROR FUNCTION>(): [ERROR : <ERROR FUNCTION RETURN TYPE>] defined in root package'
```
I had to copy and modify those classes to make the Kotlin support working under Eclipse. We should discuss with Rodrigo how to share the implementation between Buildship and gradle-script-kotlin.

org.eclipse.buildship.kotlin/src/main/kotlin/org/eclipse/buildship/kotlin/GradleKotlinScriptDependenciesResolver.kt
donat: Fixed (will push the changes later).
Again, we should see how could we reuse the implementation from `gradle-script-kotlin`.

org.eclipse.buildship.kotlin/src/main/kotlin/org/eclipse/buildship/kotlin/KotlinBuildScript.kt
donat: Here the `resolver = GradleKotlinScriptDependenciesResolver::class,` part is the interesting one: it references the class from the `org.eclipse.buildship.kotlin` plugin instead of from GSK.

org.eclipse.buildship.kotlin/src/main/kotlin/org/eclipse/buildship/kotlin/KotlinModelQuery.kt
donat: Fixed (will push the changes later).

org.eclipse.buildship.kotlin/build.gradle
oehme: This just shows that defining this stuff in the annotation doesn't make sense. It should be provided by the ScriptTemplateProvider. Same as with the isApplicable check. We should not need to load the template at all at this point.

donat:
I've created new issues for all your concerns @oehme. Once I manage to get rid of the `ip-validation` failure I'll merge the branch to master.

org.eclipse.buildship.kotlin/build.gradle
donat: Yes, I agree with you.

org.eclipse.buildship.kotlin/src/main/kotlin/org/eclipse/buildship/kotlin/KotlinBuildScript.kt
oehme: Okay, same comment as above: This shows that loading the template to configure the IDE is wrong.

PR #323, opened 2017-02-06
In classpath container declare project dependencies before external dependencies

### LINKED ISSUE #244: Project dependencies should be at the beginning of the classpath container
<a href="https://github.com/donat"></img></a> **Issue by [donat](https://github.com/donat)**
_Wednesday Feb 01, 2017 at 12:53 GMT_

----

## Motivation
https://discuss.gradle.org/t/eclipse-classpath-ordering-project-dependency-always-at-the-bottom-of-dependencies

## Code review

- [ ] Does the change work?
- [ ] Is the change sufficiently tested?
- [ ] Is the public API modification handled appropriately?
- [ ] Is it sufficiently documented?
- [ ] Is it implemented well?

See also the [full checklist](https://github.com/gradle/process/wiki/Process#implementation-checklist).

donat:
This PR resolves #244 

Please ignore the merge issues. I've branched from the dc-linked-resources in order to avoid duplicate work in `GradleClasspathContainerUpdater`.

### DIFF
--- a/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/GradleClasspathContainerUpdater.java
+++ b/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/GradleClasspathContainerUpdater.java
@@ -107,28 +107,34 @@ final class GradleClasspathContainerUpdater {
         Builder<IClasspathEntry> result = ImmutableList.builder();
         for (OmniExternalDependency dependency : this.gradleProject.getExternalDependencies()) {
             File dependencyFile = dependency.getFile();
-            String dependencyName = dependencyFile.getName();
-            if (!dependencyFile.exists()) {
-                IPath path = new Path("/" + dependencyFile.getPath());
-                IResource member = this.eclipseProject.getProject().findMember(path);
-                if (member != null) {
-                    IClasspathEntry entry = JavaCore.newLibraryEntry(member.getFullPath(), null, null);
+            boolean linkedResourceCreated = tryCreatingLinkedResource(dependencyFile, result);
+            if (!linkedResourceCreated) {
+                String dependencyName = dependencyFile.getName();
+                // Eclipse only accepts folders and archives as external dependencies (but not, for example, a DLL)
+                if (dependencyFile.isDirectory() || dependencyName.endsWith(".jar") || dependencyName.endsWith(".zip")) {
+                    IPath path = org.eclipse.core.runtime.Path.fromOSString(dependencyFile.getAbsolutePath());
+                    File dependencySource = dependency.getSource();
+                    IPath sourcePath = dependencySource != null ? org.eclipse.core.runtime.Path.fromOSString(dependencySource.getAbsolutePath()) : null;
+                    IClasspathEntry entry = JavaCore.newLibraryEntry(path, sourcePath, null, ClasspathUtils.createAccessRules(dependency), ClasspathUtils
+                            .createClasspathAttributes(dependency), dependency.isExported());
                     result.add(entry);
-                    continue;
                 }
-            }
-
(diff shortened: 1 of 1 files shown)


org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/GradleClasspathContainerUpdater.java
```diff
@@ -80,39 +81,64 @@ private void updateClasspathContainer(IProgressMonitor monitor) throws JavaModel
...
+        for (OmniExternalDependency dependency : this.gradleProject.getExternalDependencies()) {
+            File dependencyFile = dependency.getFile();
+            String dependencyName = dependencyFile.getName();
+            if (!dependencyFile.exists()) {
+                IPath path = new Path("/" + dependencyFile.getPath());
+                IResource member = this.eclipseProject.getProject().findMember(path);
+                if (member != null) {
+                    IClasspathEntry entry = JavaCore.newLibraryEntry(member.getFullPath(), null, null);
+                    result.add(entry);
+                    continue;
```
oehme: should use else below instead of continue here

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/GradleClasspathContainerUpdater.java
```diff
@@ -80,39 +81,64 @@ private void updateClasspathContainer(IProgressMonitor monitor) throws JavaModel
...
-                    @Override
-                    public IClasspathEntry apply(OmniEclipseProjectDependency dependency) {
-                        IPath path = new Path("/" + dependency.getPath());
-                        return JavaCore.newProjectEntry(path, ClasspathUtils.createAccessRules(dependency), true, ClasspathUtils.createClasspathAttributes(dependency), dependency.isExported());
-                    }
-                }).toList();
-
-        // external dependencies
-        List<IClasspathEntry> externalDependencies = FluentIterable.from(this.gradleProject.getExternalDependencies()).filter(new Predicate<OmniExternalDependency>() {
+        boolean hasExportedEntry = FluentIterable.from(externalDependencies).anyMatch(new Predicate<IClasspathEntry>() {
```
oehme: We should check for the Gradle version directly if possible.

donat: That's unfortunately not possible atm, not without loading an additional tooling model. So far we relied on the workaround that certain features are not available in the `EclipseProject` model if loaded from an older Gradle version. I'll create a new issue for that.

donat: https://github.com/eclipse/buildship/issues/370

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/GradleClasspathContainerUpdater.java
donat: You mean `if (!dependencyFile.exists()) { ... } else { /* add classic deps */ }`?
That way we wouldn't declare those dependencies that are deleted from the Gradle cache. I'm referring your comment here: https://github.com/gradle/tooling-commons/commit/5eddc913cf92392d8bb77d14878f2ca2a7f861f9#commitcomment-20713851

donat: Anyway I refactored the code to make it more readable.

PR #342, opened 2017-02-08
PersistentModel never returns null and is immutable

donat:


org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/preferences/internal/DefaultModelPersistenceTest.groovy
```diff
@@ -6,163 +6,121 @@ import org.eclipse.core.runtime.Path
...
-        then:
-        model.buildDir == buildDir
-        model.subprojectPaths == subProjectPaths
-        model.classpath == classpath
-        model.derivedResources == derivedResources
-        model.linkedResources == linkedResources
     }
 
-    def "Can set null values"() {
+    def "By default an empty model is returned"() {
```
oehme: By default, null should be returned.

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/preferences/PersistentModelTest.groovy
```diff
@@ -0,0 +1,136 @@
...
+
+class PersistentModelTest extends WorkspaceSpecification {
+
+    IProject project
+
+    void setup() {
+        project = newProject('sample-project')
+    }
+
+    def "Empty model has sensible defaults"() {
```
oehme: There should be no "empty model", but null, because we could no longer make a difference between "we don't know" and "empty" otherwise.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/preferences/internal/DefaultPersistentModelBuilder.java
```diff
@@ -0,0 +1,97 @@
...
+
+    private static final IPath DEFAULT_BUILD_DIR = new Path("build");
+    private static final Collection<IPath> DEFAULT_SUBPROJECTS = ImmutableList.<IPath>of();
+    private static final List<IClasspathEntry> DEFAULT_CLASSPATH = ImmutableList.<IClasspathEntry>of();
+    private static final Collection<IPath> DEFAULT_DERIVED_RESOURCES = ImmutableList.<IPath>of();
+    private static final Collection<IPath> DEFAULT_LINKED_RESOURCES = ImmutableList.<IPath>of();
+
+    private final IProject project;
+
+    private IPath buildDir = DEFAULT_BUILD_DIR;
```
oehme: There should be no defaults. The builder should fail hard if something isn't specified.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/GradleClasspathContainerUpdater.java
```diff
@@ -118,42 +117,27 @@ public IClasspathEntry apply(OmniExternalDependency dependency) {
...
-     * @throws JavaModelException if the classpath cannot be assigned
      */
     public static boolean updateFromStorage(IJavaProject eclipseProject, IProgressMonitor monitor) throws JavaModelException {
-        List<IClasspathEntry> storedClasspath = CorePlugin.modelPersistence().loadModel(eclipseProject.getProject()).getClasspath();
-        if (storedClasspath != null) {
-            setClasspathContainer(eclipseProject, storedClasspath, monitor);
-            return true;
-        } else {
+        PersistentModel model = CorePlugin.modelPersistence().loadModel(eclipseProject.getProject());
+        if (model.isEmptyModel()) {
```
oehme: We should just use null here. Alternatively you can let the empty model throw exceptions on all getters. 

But we cannot let the empty model just silently return defaults, as that would hide programming errors.

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/preferences/internal/DefaultModelPersistenceTest.groovy
donat: Fixed.

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/preferences/PersistentModelTest.groovy
donat: Fixed

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/GradleClasspathContainerUpdater.java
donat: Fixed.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/preferences/internal/DefaultModelPersistence.java
```diff
@@ -54,32 +52,33 @@
 public final class DefaultModelPersistence implements ModelPersistence, EventListener {
 
     // Cache holding (project name - persistent model property) pairs.
```
donat: Commenting here, since I can't do that on your change request. 

There's one thing which worries me a bit.  Originally I wanted to avoid the null valules in the `PersistentModel` and in the `PersistentModelBuilder`. With your proposal, there'll be a bunch of null checks in the updaters, which seem a bit error-prone to me.

oehme: You could also have 

```
PersistentModelBuilder {
  PersistentModel getPrevious()
}
```

Then you could check `builder.previous.present?` instead of doing null checks.

donat: `builder.prevous` could return the default values if there's no previous state.

oehme: I guess that would work too. If we are inside our synchronization code, we will definitely update the state anyway. The difference between present and default values is only relevant to user code.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/preferences/internal/DefaultModelPersistence.java
```diff
@@ -54,32 +52,33 @@
...
     }
 
     @Override
     public PersistentModel loadModel(IProject project) {
-        return this.modelCache.getUnchecked(project);
+        PersistentModel model = this.cache.get(project);
+        if (model == null) {
+            model = readModel(project);
+            if (model != null) {
+                this.cache.put(project, model);
```
donat: I reverted to the `CacheLoader` implementation and used a new `AbsentPersistentModel` class to distinguish nonexisting models.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/preferences/internal/DefaultModelPersistence.java
donat: Implemented.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/preferences/PersistentModel.java
```diff
@@ -15,32 +15,21 @@
...
  * @author Donat Csikos
  */
-public abstract class PersistentModel {
-
-    public abstract IProject getProject();
-
-    public abstract IPath getBuildDir();
-
-    public abstract Collection<IPath> getSubprojectPaths();
+public interface PersistentModel {
```
oehme: Missing `isPresent` method

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/preferences/internal/DefaultModelPersistence.java
```diff
@@ -51,34 +53,37 @@
...
     public PersistentModel loadModel(IProject project) {
-        PersistentModel model = this.cache.get(project);
-        if (model == null) {
-            model = readModel(project);
-            if (model != null) {
-                this.cache.put(project, model);
-            }
+        PersistentModel model = this.modelCache.getUnchecked(project);
+        if (model instanceof AbsentPersistentModel) {
+            return null;
```
donat: I'm totally lost now. Two iterations ago you this comment: 

> The current solution would silently hide programming errors (forgetting to check isEmptyModel).

and

> By default, null should be returned.

Frankly, I think both of them are equally right, and it's only a syntactic sugar whether to check for `null` or for `isPresent` models. So, `isPresent()` is your preference?

oehme: I provided two alternatives: "An empty model that throws exceptions" or "null". Then we saw that we need a null object for caching anyway, so the "empty, throwing model" is the clear winner.

donat: Fine with me. Finally, we are so close! :)

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/preferences/PersistentModelFactory.java
```diff
@@ -0,0 +1,34 @@
...
+import org.eclipse.jdt.core.IClasspathEntry;
+
+import org.eclipse.buildship.core.preferences.internal.DefaultPersistentModel;
+
+/**
+ * Factory to create {@link PersistentModel} instances.
+ *
+ * @author Donat Csikos
+ */
+public final class PersistentModelFactory {
```
donat: Deleted

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/preferences/internal/DefaultModelPersistence.java
donat: Done

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/preferences/PersistentModel.java
donat: Added

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/preferences/ModelPersistence.java
```diff
@@ -22,7 +22,9 @@
 
     /**
      * Reads the project model and returns a snapshot of the current state. If no model has been
-     * saved for the project then the method returns {@code null}.
+     * saved for the project then an empty model is returned. For an empty model the
+     * {@link PersistentModel#isPresent()} method returns {@code true} and the other getter methods
```
oehme: returns false

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/SynchronizeGradleBuildOperation.java
```diff
@@ -219,8 +219,8 @@ private void synchronizeOpenWorkspaceProject(OmniEclipseProject project, IProjec
 
     private PersistentModel loadCurrentModel(IProject workspaceProject) {
         PersistentModel model = CorePlugin.modelPersistence().loadModel(workspaceProject);
-        if (model == null) {
-            model = PersistentModelFactory.from(workspaceProject, new Path("build"), ImmutableList.<IPath>of(), ImmutableList.<IClasspathEntry>of(), ImmutableList.<IPath>of(), ImmutableList.<IPath>of());
+        if (!model.isPresent()) {
```
oehme: This should just `return CorePlugin.modelPersistence().loadModel(workspaceProject);`. We should not duplicate the defaults here. The defaults are coded in the individual updaters.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/preferences/internal/AbstractPersistentModel.java
```diff
@@ -0,0 +1,24 @@
...
+package org.eclipse.buildship.core.preferences.internal;
+
+import org.eclipse.buildship.core.preferences.PersistentModel;
+
+/**
+ * Common superclass for {@link PersistentModel} implementations.
+ *
+ * @author Donat Csikos
+ */
+public abstract class AbstractPersistentModel implements PersistentModel {
```
oehme: This superclass seems unnecessary.

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/workspace/internal/GradleClasspathContainerUpdaterTest.groovy
```diff
@@ -59,4 +65,22 @@ class GradleClasspathContainerUpdaterTest extends WorkspaceSpecification {
     IClasspathEntry[] getResolvedClasspath() {
         project.getResolvedClasspath(false)
     }
+
+    private PersistentModel emptyPersistentModel(IProject project) {
```
oehme: Duplicate method

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/preferences/ModelPersistence.java
donat: Fixed.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/SynchronizeGradleBuildOperation.java
donat: Fixed

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/workspace/internal/GradleClasspathContainerUpdaterTest.groovy
donat: Removed.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/SynchronizeGradleBuildOperation.java
donat: As CI shows, it's not fixed. :)

donat: The problem is that the PersistenceModelBuilder is initialized from the PersistenceModel passed as an argument. My gut feeling is that it's better to have a centralized default model passed to the updates.

oehme: The builder needs to check whether the model is present, just like every other client. This default value code should not be here.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/PersistentModelBuilder.java
```diff
@@ -0,0 +1,78 @@
...
+
+    private IPath buildDir;
+    private Collection<IPath> subprojectPaths;
+    private List<IClasspathEntry> classpath;
+    private Collection<IPath> derivedResources;
+    private Collection<IPath> linkedResources;
+
+    public PersistentModelBuilder(PersistentModel previous) {
+        this.previous = Preconditions.checkNotNull(previous);
+        this.buildDir = previous.getBuildDir();
```
oehme: These assignments need to be guarded by `if(previous.present)`

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/SynchronizeGradleBuildOperation.java
donat: Now it's fixed.

PR #369, opened 2017-02-10
Replace GradleClasspathContainer on project only if content changed

pwoelfle:
This pull request fixes the problem that on each synchronization the GradleClasspathContainer is replaced on the project, although there are no changes. This issue caused on big projects unnecessary incremental builds, because the project description is marked as changed after the classpath container is replaced. See also Gradle discussion [1].

Eclipse internal checks if the old classpath container differs from the new classpath container [2]. Therefore the fix only implements the equals() and hashcode() of the GradleClasspathContainer to fix the problem.

[1] https://discuss.gradle.org/t/classpath-container-update-on-synchronization/21499
[2] org.eclipse.jdt.internal.core.SetContainerOperation.executeOperation()


pwoelfle:
I just signed the ECA. How to re-run the ip-validation?

oehme:
Thank you for the fix and for thinking of a test case as well! This looks good to me.

You need to sign off on your commit (commit -s) for the IP validation to accept it.

pwoelfle:
@oehme IP validation is now successful.

oehme:
Thank you Pascal!

pwoelfle:
You're welcome Stefan

PR #396, opened 2017-02-22
Add 'enable build scans' option to run configurations

donat:
The changes contain is built on top of the `dc-bs-workspace` branch. 

The implementation makes the `GradleRunConfigurationAttributes` class a bit convoluted which makes me wonder whether we should do some refactoring there.


org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/GradleRunConfigurationAttributes.java
```diff
@@ -61,23 +65,19 @@
     private final ImmutableList<String> argumentExpressions;
     private final boolean showExecutionView;
     private final boolean showConsoleView;
+    private final boolean overrideWorkspaceSettings;
+    private final boolean overriddenOfflineMode;
+    private final boolean overriddenBuildScansEnabled;
+
+    private GradleRunConfigurationAttributes(List<String> tasks, String workingDirExpression, String gradleDistribution, String javaHomeExpression,
```
oehme: Should probably change the existing constructor, not create a new one.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/GradleRunConfigurationAttributes.java
```diff
@@ -185,6 +188,18 @@ public boolean isShowConsoleView() {
         return this.showConsoleView;
     }
 
+    public boolean isOverrideWorkspaceSettings() {
+        return this.overrideWorkspaceSettings;
+    }
+
+    public boolean isOverriddenOffline() {
```
oehme: This should just be `isOffline`

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/GradleRunConfigurationAttributes.java
```diff
@@ -185,6 +188,18 @@ public boolean isShowConsoleView() {
...
+        return this.overriddenOfflineMode;
+    }
+
+    public boolean isOverriddenBuildScansEnabled() {
```
oehme: This should just be `isBuildScansEnabled`.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/GradleRunConfigurationAttributes.java
```diff
@@ -258,88 +290,83 @@ public static void applyShowConsoleView(boolean showConsoleView, ILaunchConfigur
         launchConfiguration.setAttribute(SHOW_CONSOLE_VIEW, showConsoleView);
     }
 
+    public static void applyWorkspaceOverride(boolean overrideWorkspaceSettings, boolean offlineMode, boolean buildScansEnabled, ILaunchConfigurationWorkingCopy launchConfiguration) {
```
oehme: Why are all these methods public? Shouldn't callers just use the single `apply` method?

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/GradleRunConfigurationAttributes.java
donat: Fixed

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/GradleRunConfigurationAttributes.java
donat: Fixed

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/GradleRunConfigurationAttributes.java
donat: Fixed

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/GradleRunConfigurationAttributes.java
```diff
@@ -197,13 +212,29 @@ public boolean hasSameUniqueAttributes(ILaunchConfiguration launchConfiguration)
...
 
     public FixedRequestAttributes toFixedRequestAttributes() {
-        return FixedRequestAttributesBuilder.fromWorkspaceSettings(getWorkingDir())
-            .gradleUserHome(getGradleUserHome())
-            .gradleDistribution(getGradleDistribution())
-            .javaHome(getJavaHome())
-            .jvmArguments(getJvmArguments())
-            .arguments(getArguments())
-            .build();
+        if (isOverrideWorkspaceSettings()) {
```
donat: It was a good idea. I changed the implementation and the code looks much cleaner now.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/GradleRunConfigurationAttributes.java
donat: Because how the API on the Eclipse run configuration UI is structured. Each page can only update part of the complete run config. So for instance `GradleDistributionTab` calls `applyGradleDistribution`, `ArgumentsTab` calls `applyArgumentExpressions` and so forth.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/GradleRunConfigurationAttributes.java
```diff
@@ -258,88 +290,83 @@ public static void applyShowConsoleView(boolean showConsoleView, ILaunchConfigur
...
+        }
+    }
+
     public static GradleRunConfigurationAttributes with(List<String> tasks, String workingDirExpression, String serializedGradleDistribution,
             String javaHomeExpression, List<String> jvmArgumentExpressions, List<String> argumentExpressions, boolean showExecutionView, boolean showConsoleView, boolean useGradleDistributionFromImport) {
         return new GradleRunConfigurationAttributes(tasks, workingDirExpression, serializedGradleDistribution, javaHomeExpression, jvmArgumentExpressions,
                 argumentExpressions, showExecutionView, showConsoleView, useGradleDistributionFromImport);
     }
 
+    public static GradleRunConfigurationAttributes with(List<String> tasks, String workingDirExpression, String serializedGradleDistribution,
```
donat: Fixed

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/GradleRunConfigurationAttributes.java
oehme: Hmmm can I click `apply` on individual pages? If not then the model doesn't match the reality of the UI. Nothing we need to fix right now, but something worth putting in the backlog for cleanup.

oehme:
Looks good, nice and clean :)

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/GradleRunConfigurationAttributes.java
donat: Sounds good. I checked the launch config UI's api and looks like we can do the simplification.  Here's the card for that: https://github.com/eclipse/buildship/issues/417

PR #402, opened 2017-02-24
Task execution uses Tooling API directly

donat:


org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/BaseLaunchRequestJob.java
```diff
@@ -64,34 +77,52 @@ protected final void runToolingApiJob(IProgressMonitor monitor) {
...
-        Event event = new DefaultExecuteLaunchRequestEvent(processDescription, request);
-        CorePlugin.listenerRegistry().dispatch(event);
-
-        // launch the build
-        request.executeAndWait();
+        final FixedRequestAttributes attributes = getConfigurationAttributes().toFixedRequestAttributes();
+        ProjectConnection connection = null;
+        try {
+            // apply FixedRequestAttributes on connector and establish TAPI connection
+            GradleConnector connector = GradleConnector.newConnector().forProjectDirectory(attributes.getProjectDir());
```
oehme: Ideally, Fixed/Transient attributes should have methods to apply themselves to `GradleConnector`/`ProjectConnection`/`ConfigurableLauncher`, so we don't need to repeat ourselves. We can do that in another PR though.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/BaseLaunchRequestJob.java
```diff
@@ -64,34 +77,52 @@ protected final void runToolingApiJob(IProgressMonitor monitor) {
...
+
+            // print the applied run configuration settings at the beginning of the console output
+            OutputStreamWriter writer = new OutputStreamWriter(processStreams.getConfiguration());
+            writeFixedRequestAttributes(attributes, writer, monitor);
+
+            // execute the build
+            launcher.run();
+        } finally {
+            if (connection != null) {
+                connection.close();
```
oehme: Our `GradleBuild` interface should provide single-use launchers like this:

```
build = CorePlugin.workspaceManager.gradleBuild(FixedAttributes).newBuild(TransientAttributes)
// do more configuration on the build launcher
build.run()
// connection is closed transparently, calling `run` again throws an IllegalStateException
```

Then we can remove most of the boilerplate above, including the try-catch.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/BaseLaunchRequestJob.java
```diff
@@ -41,21 +43,32 @@
...
+ * Base class to execute Gradle builds in a job.
  */
 public abstract class BaseLaunchRequestJob extends ToolingApiJob {
 
+    /**
+     * Common interface to retrieve and execute build and test launch requests. This interface is
+     * needed since there's no common {@code run()} method defined on the
+     * {@link LongRunningOperation} subclasses.
+     */
+    public static interface Launcher {
```
oehme: I would just go with an abstract `launch` method on `BaseLaunchRequestJob` and add a generic type parameter for the kind of thing it launches. For instance

```
TestLaunchRequestJob extends BaseLaunchRequestJob<TestLauncher> {
  @Override
  TestLauncher createLauncher(GradleBuild, TransientRequestAttributes) {
    // see also my comments below
    gradleBuild.createTestLauncher(transientRequestAttributes)
  }

  @Override
  void launch(TestLauncher) {
    testLauncher.run()
  }
}
```

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/BaseLaunchRequestJob.java
donat: I like the idea, although there's one issue I'd like to discuss. The launcher is created from `ProjectConnection` so for instance the `RunGradleJvmTestLaunchRequestJob` would have this:
```
    @Override
    protected GradleLauncher createLauncher(GradleBuild gradleBuild, TransientRequestAttributes transientAttributes) {
        return gradleBuild.newTestLaunch(transientAttributes, new LaunchConfig<TestLauncher>() {

            @Override
            public void apply(TestLauncher launch) {
                launch.withTests(RunGradleTestLaunchRequestJob.this.testDescriptors);
            }
        });
    }
```

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/BaseLaunchRequestJob.java
donat: I've adjusted the implementation but I think there's still room for improvement.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/BaseLaunchRequestJob.java
donat: Here's the issue for that: https://github.com/eclipse/buildship/issues/410

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/BaseLaunchRequestJob.java
donat: Fixed.

oehme: Using a callback is a bit limiting. I see you had to copy-paste some code as a result instead of having it in the base class like before. It also forced you to create a new class (`GradleInvocation`), which is against our goal for removing indirection layers over the TAPI types.

Instead what I was thinking was returning a `BuildLauncher` or `TestLauncher` which automatically closes the underlying `ProjectConnection` after it has first been used. This can for instance be done with a listener that listens for build completion.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/GradleBuild.java
```diff
@@ -84,4 +88,36 @@
...
     FixedRequestAttributes getRequestAttributes();
+
+    /**
+     * Creates a new Gradle build invocation.
+     *
+     * @param transientAttributes the transient attributes for the invocation
+     * @param config the configuration for the build launcher
+     * @return the object that can run the Gradle build
+     */
+    GradleInvocation newBuildInvocation(TransientRequestAttributes transientAttributes, BuildLauncherConfig config);
```
oehme: We should not introduce new types here. Instead we should just return a `TestLauncher` (the interface from the TAPI).

The underlying project connection can be closed transparently with a progress listener that the `DefaultGradleBuild` registers. When the "build finished" event is received, we can call `ProjectConnection.close()`.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/BaseLaunchRequestJob.java
donat: Unfortunately, there's no way to add event listeners to the launchers to detect when the build is finished. We could solve this by creating delegate classes for the launchers.

oehme: Yes, a decorator pattern would work well there.

donat: Fixed, although I'm not happy with the boilerplate code. Maybe we should just switch to Kotlin ;)

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/ConnectionAwareBuildLauncherDelegate.java
```diff
@@ -0,0 +1,158 @@
...
+
+    ConnectionAwareBuildLauncherDelegate(ProjectConnection connection, BuildLauncher delegate) {
+        super(delegate);
+        this.connection = connection;
+    }
+
+    @Override
+    public void run() {
+        this.delegate.run();
+        this.connection.close();
```
oehme: needs to be wrapped in try-finally

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/ConnectionAwareBuildLauncherDelegate.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    @Override
+    public void run(final ResultHandler<? super Void> resultHandler) {
+        this.delegate.run(new ResultHandler<Void>() {
+
+            @Override
+            public void onComplete(Void result) {
+                resultHandler.onComplete(result);
```
oehme: try-finally

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/ConnectionAwareBuildLauncherDelegate.java
```diff
@@ -0,0 +1,158 @@
...
+                ConnectionAwareBuildLauncherDelegate.this.connection.close();
+            }
+
+            @Override
+            public void onFailure(GradleConnectionException exception) {
+                resultHandler.onFailure(exception);
+            ConnectionAwareBuildLauncherDelegate.this.connection.close();
```
oehme: try-finally

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/ConnectionAwareLongRunningOperationDelegate.java
```diff
@@ -0,0 +1,124 @@
...
+
+abstract class ConnectionAwareLongRunningOperationDelegate<T extends LongRunningOperation> implements LongRunningOperation {
+
+    protected final T delegate;
+
+    ConnectionAwareLongRunningOperationDelegate(T delegate) {
+        this.delegate = delegate;
+    }
+
+    protected static ProjectConnection openConnection(FixedRequestAttributes fixedAttributes) {
```
oehme: The connection opening should be done by the `GradleBuild` instance.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/ConnectionAwareBuildLauncherDelegate.java
```diff
@@ -0,0 +1,158 @@
...
+import org.gradle.tooling.ProjectConnection;
+import org.gradle.tooling.ResultHandler;
+import org.gradle.tooling.events.OperationType;
+import org.gradle.tooling.model.Launchable;
+import org.gradle.tooling.model.Task;
+
+import com.gradleware.tooling.toolingmodel.repository.FixedRequestAttributes;
+import com.gradleware.tooling.toolingmodel.repository.TransientRequestAttributes;
+
+final class ConnectionAwareBuildLauncherDelegate extends ConnectionAwareLongRunningOperationDelegate<BuildLauncher> implements BuildLauncher {
```
oehme: I think you could save a lot of code here by using `java.lang.Proxy` instead of implementing the delegation pattern by hand. The reflection overhead would be negligible, since we don't call these methods in a hot loop.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/ConnectionAwareLongRunningOperationDelegate.java
```diff
@@ -0,0 +1,124 @@
...
+        GradleConnector connector = GradleConnector.newConnector().forProjectDirectory(fixedAttributes.getProjectDir());
+        GradleDistributionWrapper.from(fixedAttributes.getGradleDistribution()).apply(connector);
+        connector.useGradleUserHomeDir(fixedAttributes.getGradleUserHome());
+        return connector.connect();
+    }
+
+    protected void applyRequestAttributes(FixedRequestAttributes fixedAttributes, TransientRequestAttributes transientAttributes) {
```
oehme: This should be done in `GradleBuild`. These decorators should only handling the automatic closing, nothing else.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/ConnectionAwareBuildLauncherDelegate.java
donat: Implemented with proxy.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/ConnectionAwareBuildLauncherDelegate.java
donat: Fixed

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/ConnectionAwareBuildLauncherDelegate.java
donat: Fixed

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/ConnectionAwareBuildLauncherDelegate.java
donat: Fixed

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/ConnectionAwareLongRunningOperationDelegate.java
donat: Fixed

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/ConnectionAwareLongRunningOperationDelegate.java
donat: Fixed

oehme:
Excellent cleanup :)

donat:
We are getting there!

PR #413, opened 2017-03-07
Build scans and offline can be enabled from project preferences

donat:
The PR is created against the `dc-bs-project-prefs` branch to be able to continue before the previous reviews are finished.


org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/configuration/internal/DefaultProjectConfigurationManager.java
```diff
@@ -50,26 +45,14 @@ public DefaultProjectConfigurationManager(WorkspaceOperations workspaceOperation
...
-        // configuration (different java home, etc.)
-        // if such an inconsistent state is detected, it means that the Gradle configurations were
-        // changed/corrupted manually
-        Map<String, ProjectConfiguration> rootProjectDirs = Maps.newHashMap();
-        for (ProjectConfiguration rootProjectConfiguration : rootConfigurations.build()) {
-            String rootProjectDirPath = rootProjectConfiguration.toRequestAttributes(ConversionStrategy.IGNORE_WORKSPACE_SETTINGS).getProjectDir().getAbsolutePath();
-            if (!rootProjectDirs.containsKey(rootProjectDirPath)) {
-                rootProjectDirs.put(rootProjectDirPath, rootProjectConfiguration);
-            } else {
-                throw new GradlePluginsRuntimeException(String.format("Inconsistent Gradle project configuration for project at %s.", rootProjectDirPath));
```
oehme: Why did we remove this check?

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/configuration/internal/DefaultProjectConfigurationManager.java
```diff
@@ -112,7 +95,7 @@ private ProjectConfiguration tryReadProjectConfiguration(IProject workspaceProje
             return this.projectConfigurationPersistence.readProjectConfiguration(workspaceProject);
         } catch (RuntimeException e) {
             if (suppressErrors) {
-                CorePlugin.logger().warn(String.format("Cannot load project configuration for project %s.", workspaceProject.getName()), e);
+                CorePlugin.logger().debug(String.format("Cannot load project configuration for project %s.", workspaceProject.getName()), e);
```
oehme: Why did we reduce the log level?

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/configuration/internal/DefaultProjectConfigurationPersistence.java
```diff
@@ -53,9 +61,28 @@ private static ProjectConfiguration loadFromPreferencesApi(IProject project) {
...
+        String projectDir = preferences.readString(PREF_KEY_CONNECTION_PROJECT_DIR);
+        String distribution = preferences.readString(PREF_KEY_CONNECTION_GRADLE_DISTRIBUTION);
+        boolean overrideWorkspaceSettings = preferences.readBoolean(PREF_KEY_OVERRIDE_WORKSPACE_SETTINGS, false);
+        boolean buildScansEnabled = preferences.readBoolean(PREF_KEY_BUILD_SCANS_ENABLED, false);
+        boolean offlineMode = preferences.readBoolean(PREF_KEY_OFFLINE_MODE, false);
+
+        File rootDir = rootProjectFile(project, projectDir);
+        GradleDistribution gradleDistribution = GradleDistributionSerializer.INSTANCE.deserializeFromString(distribution);
+        if (overrideWorkspaceSettings) {
+            return ProjectConfiguration.from(project, rootDir, gradleDistribution, overrideWorkspaceSettings, buildScansEnabled, offlineMode);
```
oehme: I don't think we need two factory methods here, we just always pass the values and correctly respect "overrideWorkspaceSettings" when saving again.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/util/configuration/FixedRequestAttributesBuilder.java
```diff
@@ -100,4 +102,15 @@ public static FixedRequestAttributesBuilder fromWorkspaceSettings(File projectDi
                 .buildScansEnabled(configuration.isBuildScansEnabled())
                 .arguments(CorePlugin.invocationCustomizer().getExtraArguments());
      }
+
+    public static FixedRequestAttributesBuilder fromProjectSettings(IProject project) {
+        WorkspaceConfiguration workspaceConfig = CorePlugin.workspaceConfigurationManager().loadWorkspaceConfiguration();
```
oehme: We should delegate to `fromWorkspaceSettings` here and then just call the methods that are relevant for project settings.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/SynchronizeGradleBuildOperation.java
```diff
@@ -195,8 +194,11 @@ private void synchronizeOpenWorkspaceProject(OmniEclipseProject project, IProjec
 
         CorePlugin.workspaceOperations().addNature(workspaceProject, GradleProjectNature.ID, progress.newChild(1));
 
-        ProjectConfiguration configuration = ProjectConfiguration.from(this.build, project);
-        CorePlugin.projectConfigurationManager().saveProjectConfiguration(configuration, workspaceProject);
+        Optional<ProjectConfiguration> projectConfig = CorePlugin.projectConfigurationManager().tryReadProjectConfiguration(workspaceProject);
```
oehme: Why did we change this logic?

donat:
Here's an overview how I imagine the new config API:

We have the following configurations:
```
interface WorkspaceConfig {
    File getGradleUserHome();
    public boolean isOffline();
    public boolean isBuildScansEnabled();
}

interface ProjectConfig {
    String getPathToRoot();
    BuildConfig getBuildConfig();
}

interface BuildConfig {
    File getRootProjectDirectory();
    GradleDistribution getGradleDistribution();
    File getGradleUserHome();
    boolean isOverrideWorkspaceSettings();
    boolean isBuildScansEnabled();
    boolean isOfflineMode();

    // attributes persisted only in the run configurations
    File getGradleJavaHome();
    List<String> getArguments();
    List<String> getJvmArguments();

    FixedRequestAttributes toRequestAttributes();
}
```

And then, we have on central `ConfigManager` interface available on `CorePlugin`:
``` interface ConfigManager {
    WorkspaceConfig createWorkspaceConfig(File gradleUserHome, boolean gradleIsOffline, boolean buildScansEnabled);
    WorkspaceConfig loadWorkspaceConfig();
    void saveWorkspaceConfig(WorkspaceConfig config);

    BuildConfig createBuildConfig(File rootProjectDirectory, GradleDistribution gradleDistribution, boolean overrideWorkspaceSettings, boolean buildScansEnabled, boolean offlineMode);
    void saveBuildConfig(BuildConfig updatedConfig);

    ProjectConfig createProjectConfig(BuildConfig buildConfig, File projectDirectory);
    ProjectConfig loadProjectConfig(IProject project);
    void saveProjectConfig(ProjectConfig config);
    void deleteProjectConfig(IProject project);
}
```

I added the object creation to the interface in order to hide the customization currently done in `FixedRequestAttributesBuilder`.

oehme:
I like the plan very much overall, just a few minor things:

> // attributes persisted only in the run configurations
    File getGradleJavaHome();
    List<String> getArguments();
    List<String> getJvmArguments();

That shouldn't be on the BuildConfig then, but on the RunConfig. I guess this is what you referred to on our call.

> WorkspaceConfig createWorkspaceConfig

There should be no creation method for that, since there is only one workspace config, completely managed by us.

> String getPathToRoot();

Should be getProjectDir() instead. The "path to root" is an internal detail of our persistence engine.

donat:
The latest changes are not intended to be final by any means. But given the size of the change, I wanted to have an earlier feedback if I'm on the right track. @oehme I propose we focus first on the structural changes  and then the finer-grained details (Javadoc, coverage, style, etc).

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/configuration/internal/BuildConfigurationPersistenceTest.groovy
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    def "read validates input"() {
+        when:
+        persistence.readBuildConfiguratonProperties(null)
+
+        then:
+        thrown NullPointerException
+
+        // TODO (donat) should we do the validation in the persistence or in the Manager?
```
oehme: At the level that knows whether something is wrong. So here I'd say in the persistence.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/configuration/ConfigurationManager.java
```diff
@@ -0,0 +1,49 @@
...
+
+    ProjectConfiguration createProjectConfiguration(BuildConfiguration configuration, File projectDir);
+
+    ProjectConfiguration loadProjectConfiguration(IProject project);
+
+    void saveProjectConfiguration(ProjectConfiguration configuration);
+
+    void deleteProjectConfiguration(IProject project);
+
+    RunConfiguration loadRunConfiguration(GradleRunConfigurationAttributes attributes);
```
oehme: We discussed on our call that this should take a launch configuration instead. Is that still the plan?

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/BaseLaunchRequestJob.java
```diff
@@ -70,12 +72,23 @@ protected final void runToolingApiJob(IProgressMonitor monitor) {
...
+        RunConfiguration runConfig = getRunConfig();
+        BuildConfiguration buildConfig = runConfig.getBuildConfiguration();
+        // TODO (donat) this won't be necessary once the project config can override Gradle user home
+        WorkspaceConfiguration workspaceConfig = CorePlugin.configurationManager().loadWorkspaceConfiguration();
+
+        request.gradleDistribution(buildConfig.getGradleDistribution());
+
+        // apply the fixed attributes
+        request.projectDir(buildConfig.getRootProjectDirectory());
+        request.gradleUserHomeDir(workspaceConfig.getGradleUserHome());
```
oehme: The build config should have a reference to the workspace config, so we don't need to load it separately here.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/BaseLaunchRequestJob.java
```diff
@@ -70,12 +72,23 @@ protected final void runToolingApiJob(IProgressMonitor monitor) {
...
+        request.gradleDistribution(buildConfig.getGradleDistribution());
+        request.javaHomeDir(runConfig.getJavaHome());
+        request.jvmArguments(runConfig.getJvmArguments().toArray(new String[0]));
+        request.arguments(runConfig.getArguments().toArray(new String[0]));
```
oehme: The run config should have an `applyTo(Request)`  method which handles setting all the properties on the request. It should delegate to a similar apply method on its build config, which in turn delegates first to the workspace config.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/SynchronizeGradleBuildsJob.java
```diff
@@ -102,7 +105,9 @@ private void synchronizeBuild(GradleBuild build, SubMonitor progress) throws Cor
     @Override
     public boolean shouldSchedule() {
         for (Job job : Job.getJobManager().find(CorePlugin.GRADLE_JOB_FAMILY)) {
+            System.err.println("job in gradle job family=" + job.getClass().getName());
```
oehme: :)

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/SynchronizeProjectConfigOperation.java
```diff
@@ -0,0 +1,35 @@
...
+
+import org.eclipse.buildship.core.CorePlugin;
+import org.eclipse.buildship.core.configuration.BuildConfiguration;
+
+/**
+ * Synchronizes the project configuration.
+ *
+ * @author Donat Csikos
+ */
+public final class SynchronizeProjectConfigOperation {
```
oehme: Build configuration, not project configuration

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/launch/TestLaunchShortcut.java
```diff
@@ -57,32 +64,30 @@ private void launch(JavaElementResolver resolver) {
...
-        // have execution view and console view enabled by default
-        boolean showExecutionView = true;
-        boolean showConsoleView = true;
-
-        return GradleRunConfigurationAttributes.with(ImmutableList.<String>of(), projectDir, gradleDistribution,
-                javaHome, jvmArguments, arguments, showExecutionView, showConsoleView, true);
+    private RunConfiguration collectRunConfigurationAttributes(IProject project) {
+        ProjectConfiguration projectConfig = CorePlugin.configurationManager().loadProjectConfiguration(project);
+        BuildConfiguration buildConfig = projectConfig.getBuildConfiguration();
+        // TODO donat maybe this belongs to the ConfigManager interface
```
oehme: Yep.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/configuration/ConfigurationManager.java
```diff
@@ -0,0 +1,49 @@
...
+    void saveWorkspaceConfiguration(WorkspaceConfiguration configuration);
+
+    BuildConfiguration createBuildConfiguration(File rootProjectDirectory, GradleDistribution gradleDistribution, boolean overrideWorkspaceSettings, boolean buildScansEnabled,
+            boolean offlineMode);
+
+    Set<BuildConfiguration> loadAllBuildConfigurations();
+
+    BuildConfiguration loadBuildConfiguration(File rootProject);
+
+    void saveBuildConfiguration(BuildConfiguration configuration);
```
oehme: When do we delete the build configuration?

donat: Right now we don't do that because we add/remove projects only upon synchronization. We could register an event listener for project deletion events and delete `BuildConfiguration` if no projects exist in the workspace for that Gradle build.

oehme: Yeah, I was thinking something along those lines. It's not a big issue, but it would be nice if we cleaned up after ourselves. Can you open a follow-up story for that?

donat: Sure, I'll do.

donat: https://github.com/eclipse/buildship/issues/458

donat:
@oehme I addressed you last set of review items. The base branch has diverged from master so I moved the changes to https://github.com/eclipse/buildship/pull/459
I'll continue with the test coverage there.

PR #424, opened 2017-03-10
Make Gradle classpath containers of different projects unequal

oehme:
JDT de-duplicates classpath containers when computing the runtime
classpath for a project. By default all containers with the same path
are considered equal. This breaks when a user manually adds a Gradle
project to the classpath of another Gradle project (which is still a
valid use case while composite builds and WTP don't work together, as
well as for quick experimentation). See
https://github.com/gradle/gradle/issues/1516#issuecomment-284649135

The fix is simple: Our container should be considered different for each
project.

Signed-off-by: Stefan Oehme <st.oehme@gmail.com>


oehme:
@donat I'm afraid I won't have time to create a proper test case, but I wanted to share the solution I found, based on this [example](https://github.com/gradle/gradle/issues/1516#issuecomment-284649135). Could you take this over from here?

donat:
Sure, I'll do that!

PR #434, opened 2017-03-27
Derived folders of subprojects are derived in the parent too

donat:


oehme:
The subprojects should *not* be marked as derived, because they are not. It would e.g. lead EGit to freak out and adding them to .gitignore. Only the derived folders of the subprojects should be marked as derived in the parent as well. For instance:

Structure on the file system:

root
  sub
    build

Structure in Eclipse:

root
  sub
    build <- derived

sub
  build <- derived

donat:
@oehme I've adjusted the implementation to mark the subfolders as derived.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/DerivedResourcesUpdater.java
```diff
@@ -139,4 +157,37 @@ static void update(IProject workspaceProject, OmniEclipseProject project, Persis
...
+        private final Collection<IPath> nestedProjectBuildDirs;
+
+        public DerivedResourcesInfo(IPath projectBuildDir, Collection<IPath> nestedProjectPaths, Collection<IPath> nestedProjectBuildDirs) {
+            this.projectBuildDir = projectBuildDir;
+            this.nestedProjectPaths = nestedProjectPaths;
+            this.nestedProjectBuildDirs = nestedProjectBuildDirs;
+        }
+
+        public IPath getProjectBuildDir() {
+            return this.projectBuildDir == null ? new Path("build") : this.projectBuildDir;
```
oehme: I don't think this can happen anymore, since this is already handled on line 137

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/DerivedResourcesUpdater.java
```diff
@@ -139,4 +157,37 @@ static void update(IProject workspaceProject, OmniEclipseProject project, Persis
         new DerivedResourcesUpdater(workspaceProject, project).update(persistentModel, monitor);
     }
 
+    /**
+     * Helper class to collect derived resources.
+     */
+    private static final class DerivedResourcesInfo {
```
oehme: This is not just about derived resources, but "special folders", probably should rename this and the containing class.

donat: Fixed.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/DerivedResourcesUpdater.java
donat: If the build dir points outside to the Eclipse workspace and no linked resources references that location then `normalizeBuildDirectoryPath()` will return `null`.  Returning the `build` folder as default is not a proper solution, though. How about we make `PersistentModel.getBuildDir()` return `Optional<IPath>`?

oehme: We should just return "null" in that case.

donat: Just to confirm, should we allow nulls for the build dir in the persistent model?

oehme: What I wanted to get at was removing duplication, not changing the behavior on an unknown build dir (that's another discussion). Currently the default (`new Path("build")`) is duplicated here and on line 137. It should only be in one of the two places.

donat: 👍

PR #459, opened 2017-04-25
Support for hierarchical preferences

donat:


donat:
@oehme I've added the fixes you suggested in #413 and added test coverage. Let's continue this monster of a review here :)

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/GradleLaunchConfigurationManager.java
```diff
@@ -33,14 +40,25 @@
...
     /**
+     * Returns either a new Gradle {@link ILaunchConfiguration} instance or an existing one,
+     * depending on whether there is already a Gradle run configuration for the given set of
+     * attributes or not.
+     *
+     * @param configurationAttributes the run configuration attributes, must not be not null
+     * @param saveStrategy determines whether the new run configuration should be saved to the disk.
+     * @return the new or existing Gradle run configuration
+     */
+    ILaunchConfiguration getOrCreateRunConfiguration(GradleRunConfigurationAttributes configurationAttributes, SaveStrategy saveStrategy);
```
oehme: What is this used for?

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/SynchronizeGradleBuildOperation.java
```diff
@@ -154,16 +154,22 @@ public File apply(OmniEclipseProject gradleProject) {
...
                         && (project.getLocation() == null || !gradleProjectDirectories.contains(project.getLocation().toFile()));
             }
         }).toList();
     }
 
     private void synchronizeGradleProjectWithWorkspaceProject(OmniEclipseProject project, SubMonitor progress) throws CoreException {
         progress.setWorkRemaining(1);
+
+        // save the project config before creating/opening workspace project to avoid any premature updates
+        // This ensures that events like `ProjectCreatedEvents` are not fired before the configuration is available
```
oehme: They aren't fired anyway early, because we run in a transaction. Also, at this point we already have an open workspace project, so this comment is misleading.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/SynchronizeGradleBuildsJob.java
```diff
@@ -103,6 +106,7 @@ private void synchronizeBuild(GradleBuild build, SubMonitor progress) throws Cor
     public boolean shouldSchedule() {
         for (Job job : Job.getJobManager().find(CorePlugin.GRADLE_JOB_FAMILY)) {
             if (job instanceof SynchronizeGradleBuildsJob && isCoveredBy((SynchronizeGradleBuildsJob) job)) {
+                System.err.println("false");
```
oehme: ;)

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/launch/TestLaunchShortcut.java
```diff
@@ -57,32 +66,30 @@ private void launch(JavaElementResolver resolver) {
...
+                GradleDistributionSerializer.INSTANCE.serializeToString(buildConfig.getGradleDistribution()),
+                null,
+                Collections.<String>emptyList(),
+                Collections.<String>emptyList(),
+                true,
+                true,
+                buildConfig.isOverrideWorkspaceSettings(),
+                buildConfig.isOfflineMode(),
+                buildConfig.isBuildScansEnabled());
+        ILaunchConfiguration launchConfig = CorePlugin.gradleLaunchConfigurationManager().getOrCreateRunConfiguration(attributes, SaveStrategy.DONT_PERSIST);
```
oehme: This doesn't make sense, we shouldn't create an `ILaunch` if we don't need it. We should just have a `configurationManager().createRunconfiguration()` mehod and the launch configuration manager should work with that, not the other way around.

donat:
> I can't find the support for "overwrite project settings" anymore. Where did that go?

I removed that option because it created a big confusion. The problem is that we reused existing run configurations if the project name and matches which didn't fit well with the new preference hierarchy. 

So instead, I made the run configurations to be a snapshot of the project configuration. The configurations are now fully compared so if you 1) execute a task from the task view 2) change any settings in the run config dialog and 3) re-execute the same task from the task view, you'll end up in 2 run configuration. 

This ensures that for each execution initiated from the task view always have has up-to-date settings and manually created run configurations hold the configuration independently.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/GradleLaunchConfigurationManager.java
donat: I had the idea that we should use `ILaunchConfiguration` for all task executions. This was not the case previously when a test was executed via a build operation. But in that case, we shouldn't save the run configuration as the build operation is not persistable. So I added the method above to create a run config without actually saving it.

oehme:
So every time I change workspace or project settings, I get duplicate run configs? That doesn't sound good. 

What exactly is the problem with the reuse of run configs?

donat:
The issue is that when you load the run configuration how do you determine that it was set by the user or should come from the workspace/project configuration. One confusing example is when we had to introduce the "From import" build configuration. With my solution, I could simply delete that option. Also, if we'll add more options to the workspace/project config, we'd need more 'override' checkboxes which to me feels off.

Buildship already generates a large amount of configuration, that's why I felt it wouldn't be a big issue. But we can easily implement small cleanup algorithm where we'd delete all run config if they were generated by Buildship and never modified manually.

oehme:
For things that can come from the project level we should have a "use project settings" option, just like we have a "use workspace settings" on the project level. That's a well established pattern in Eclipse. I don't see any confusion there. Let's stick with that instead of introducing a different approach.

donat:
Before we proceed with the merge, I still have two concerns I would like to discuss.

- In the `GradleBuild` we handle the project synchronization and the task execution in the same way. For that, I had to change `GradleWorkspaceManager.getGradleBuild()` to accept `RunConfiguration` instances and defined a `BuildConfiguration.toDefaultRunConfiguration()` to use when the project sync is used. 

- I think we should enable one 'Override Build configuration' per run configuration. We'd need to do it in two places: when overriding the build scans/offline mode settings and for the Gradle distribution. I propose we merge those into a single 'Build Settings' launch configuration tab where the user can use or override the existing project configuration properties.

oehme:
The `getGradleBuild()` method should not know anything about run configurations. If we need run-configuration specific information, it should be passed into the methods that create task/test launchers, not into the method that creates a build.

The second point sounds good and we should probably do the same for the project preference page.

donat:
I understand your point, but then why use GradleBuild in the first place? I mean, if we defined `GradleBuild.newBuildLauncher(RunConfiguration)`, then the client could use any `BuildConfiguration` to get the reference to a `GradleBuild` and then call `newBuildLauncher` with an independent `RunConfiguration`.

Maybe we should move the `getBuildLauncher` and `getTestLauncher` methods to `GradleWorkspaceManager`.

oehme:
We can validate that the run configuration has the same build configuration as the build we are currently referring to. I think having `sync/run/getModel` on the `GradleBuild` interface is the right place.

PR #461, opened 2017-05-04
Make executions view more performant

donat:
This commit replaces the databinding-based executions view updates with
content and label providers. The updates are performed with a periodically
scheduled job. As a result, the executions view can display much more
build operations without fooding the UI thread.

This commit also removes the dependencies to the databinding plugins.


org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/view/execution/UpdateExecutionPageJob.java
```diff
@@ -11,47 +11,58 @@
...
+import org.eclipse.swt.widgets.Display;
+import org.eclipse.ui.PlatformUI;
 
 /**
  * Updates the duration of the registered {@link OperationItem} instances in the {@link ExecutionsView} in regular intervals.
  */
-public final class UpdateDurationJob extends Job {
+public final class UpdateExecutionPageJob extends Job {
+
+    private static final int REPEAT_DELAY = 500;
```
oehme: I'd probably make this 100ms, since 500ms delay is quite noticable to a human.

PR #469, opened 2017-05-12
Add 'Open Build Scan' to the tasks and executions view

donat:


donat:
The PR is not 100% complete, there are two things missing:
- Integration tests with different build scan plugin versions
- The icon for the toolbar
I'm thinking about creating a UI test where a project with Gradle version X and build scans version Y is imported and a task is executed. Is it OK to create a bunch of build scans just for the sake of testing?
/cc @etiennestuder

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/scan/internal/DefaultBuildScanRegistry.java
```diff
@@ -0,0 +1,55 @@
...
+import org.eclipse.buildship.core.scan.BuildScanRegistry;
+
+/**
+ * Default implementation for {@link BuildScanRegistry}.
+ *
+ * @author Donat Csikos
+ */
+public final class DefaultBuildScanRegistry implements BuildScanRegistry {
+
+    private final Map<ProcessDescription, String> buildScans = Collections.synchronizedMap(new HashMap<ProcessDescription, String>());
```
oehme: We can't have a service that grows indefinitely. The information about the last build scan should instead be stored somewhere in the execution page.

donat: Fixed.

donat:
I've added the icon for the action. Personally, I like the artwork, but please don't hesitate to raise any concerns in that regard.

org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/view/execution/OpenBuildScanActionTest.groovy
```diff
@@ -0,0 +1,196 @@
...
+
+        then:
+        view.pages.size() == 2
+        view.pages[0].openBuildScanAction.enabled
+        view.pages[0].openBuildScanAction.buildScanUrl == 'https://scans.gradle.com/s/A'
+        view.pages[1].openBuildScanAction.enabled
+        view.pages[1].openBuildScanAction.buildScanUrl == 'https://scans.gradle.com/s/B'
+    }
+
+    @Ignore
```
oehme: Needs to be un-ignored, can use our dogfooding instance.

org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/view/execution/OpenBuildScanActionTest.groovy
```diff
@@ -0,0 +1,196 @@
...
+        new RunGradleBuildLaunchRequestJob(createLaunch(task, projectDir, arguments)).schedule()
+        waitForGradleJobsToFinish()
+        waitForBuildFinishedMessage()
+    }
+
+    private void waitForBuildFinishedMessage() {
+        waitFor { consoleListener.activeConsole.document.get().contains('BUILD SUCCESSFUL') }
+    }
+
+    private void waitFor(int timeout = 5000, Closure condition) {
```
oehme: Is there no better way to listen to the console?

donat: Unfortunately not. The console output changes asynchronously, so the build job finishes before the build success message is printed.

donat:
I added a few unit tests what in fact showed that the pattern is not good enough. After searching the interwebs I've found that there's a nice Android utility class that contains regexps for web URLs, which I copied over.

org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/view/execution/OpenBuildScanActionTest.groovy
donat: I found a little improvement for the waitFor condition, see the latest changes.

org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/view/execution/OpenBuildScanActionTest.groovy
donat: Fixed.

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/console/BuildScanPatternMatchListener.java
```diff
@@ -56,7 +57,7 @@ public void matchFound(PatternMatchEvent event) {
 
     @Override
     public String getPattern() {
-        return "Publishing build information...\\s+https?://[-a-zA-Z0-9+&@#/%?=~_|!:,.;]*[-a-zA-Z0-9+&@#/%=~_|]";
+        return "Publishing build information...\\s+" + PatternUtils.WEB_URL_PATTERN;
```
oehme: I'd probably not match on the three dots

donat: I replaced `...` with `.*` in the pattern.

PR #480, opened 2017-05-24
Show viewer headers by default

donat:
This change makes the headers visible in the tasks and in the executions view when Eclipse is launched with a new workspace location. As of the existing workspaces, we can't distinguish between default user-selected the visibility, so I'd not change anything that regard.


oehme:
As discussed on our meeting, we should either:

- find a solution that works without headers (i.e. a movable splitter between the data)
- remove the option and just always show the headers

donat:
There was no usable way to inject a slider between the table columns, so I enabled the headers by default and removed the related preference.

oehme:
I think this mutable "headerVisible" flag was the only reason we had our own TreeViewer implementation. Can you see if we can get rid of that class now or whether we still need it?

donat:
I don't remember this. The `TreeViewer` widget we use comes from the `JFace` plugin.

oehme:
Nevermind, I was thinking of the copy of [`FilteredTree`](https://github.com/eclipse/buildship/blob/dcfe7a3417f95a7edd8b74707328a52804f1aa53/org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/external/viewer/FilteredTree.java) that we have to hide the filter by default.

Issue #454, opened 2017-04-24
NullPointerException in DefaultPersistentModel


donat:
AERI report: https://dev.eclipse.org/recommenders/committers/aeri/v2/#!/problems/58fb7c5ae4b02cb2d5a43adf

```
Bundle: org.eclipse.buildship.core 2.0.2.v20170420-0909
Message: Synchronize Gradle projects with workspace failed due to an unexpected error.
Exception:
java.lang.NullPointerException: null
    at org.eclipse.buildship.core.preferences.internal.DefaultPersistentModel.(DefaultPersistentModel.java:41)
    at org.eclipse.buildship.core.workspace.internal.PersistentModelBuilder.build(PersistentModelBuilder.java:78)
    at org.eclipse.buildship.core.workspace.internal.SynchronizeGradleBuildOperation.synchronizeOpenWorkspaceProject(SynchronizeGradleBuildOperation.java:214)
    at org.eclipse.buildship.core.workspace.internal.SynchronizeGradleBuildOperation.addNewEclipseProjectToWorkspace(SynchronizeGradleBuildOperation.java:276)
    at org.eclipse.buildship.core.workspace.internal.SynchronizeGradleBuildOperation.synchronizeNonWorkspaceProject(SynchronizeGradleBuildOperation.java:258)
    at org.eclipse.buildship.core.workspace.internal.SynchronizeGradleBuildOperation.synchronizeGradleProjectWithWorkspaceProject(SynchronizeGradleBuildOperation.java:175)
    at org.eclipse.buildship.core.workspace.internal.SynchronizeGradleBuildOperation.access$000(SynchronizeGradleBuildOperation.java:99)
    at org.eclipse.buildship.core.workspace.internal.SynchronizeGradleBuildOperation$1.run(SynchronizeGradleBuildOperation.java:134)
    at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2240)
    at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2262)
    at org.eclipse.buildship.core.workspace.internal.SynchronizeGradleBuildOperation.synchronizeProjectsWithWorkspace(SynchronizeGradleBuildOperation.java:131)
    at org.eclipse.buildship.core.workspace.internal.SynchronizeGradleBuildOperation.run(SynchronizeGradleBuildOperation.java:115)
[...]

PR #482, opened 2017-05-25
Make synchronization more robust

donat:
The intent of this PR is to fix https://github.com/eclipse/buildship/issues/454 in two ways:
- Disallow importing and syncing projects with a subproject located at the workspace root
- Ensure that all project build directories has a non-null value.

### DIFF
--- a/org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/workspace/internal/ImportingProjectInDefaultLocation.groovy
+++ b/org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/workspace/internal/ImportingProjectInDefaultLocation.groovy
@@ -57,4 +57,37 @@ class ImportingProjectInDefaultLocation extends ProjectSynchronizationSpecificat
         1 * notification.errorOccurred(*_)
     }
 
+    def "Disallow importing the workspace root"() {
+        setup:
+        Logger logger = Mock(Logger)
+        UserNotification notification = Mock(UserNotification)
+        environment.registerService(Logger, logger)
+        environment.registerService(UserNotification, notification)
+
+        when:
+        synchronizeAndWait(workspaceDir)
+
+        then:
+        1 * logger.warn(*_)
+        1 * notification.errorOccurred(*_)
+    }
+
+    def "Disallow importing any modules located at the workspace root"() {
+        setup:
+        Logger logger = Mock(Logger)
+        UserNotification notification = Mock(UserNotification)
+        environment.registerService(Logger, logger)
+        environment.registerService(UserNotification, notification)
+        File rootProject = workspaceDir('sample2') {
+            file 'settings.gradle', "include '..'"
+        }
+
+        when:
+        synchronizeAndWait(workspaceDir)
+
+        then:
+        1 * logger.warn(*_)
+        1 * notification.errorOccurred(*_)
+    }
+
 }
--- a/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/i18n/CoreMessages.java
+++ b/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/i18n/CoreMessages.java
@@ -59,6 +59,7 @@ public final class CoreMessages extends NLS {
     public static String ErrorMessage_0_MustBeSpecified;
     public static String ErrorMessage_0_MustBeDirectory;
     public static String ErrorMessage_0_AlreadyExists;
+    public static String ErrorMessage_0_WorkspaceDirectory;
 
     public static String Preference_Label_GradleUserHome;
     public static String Preference_Label_OfflineMode;
--- a/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/util/binding/Validators.java
+++ b/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/util/binding/Validators.java
@@ -88,6 +88,21 @@ public final class Validators {
         };
     }
 
+    public static Validator<File> nonWorkspaceFolderValidator(final String prefix) {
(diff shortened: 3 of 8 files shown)


oehme:
Do we need to forbid the workspace root as a project location to make this work? I know it's unconventional and I'm fine with forbidding it. But wouldn't just fixing the NPE be enough?

donat:
Yes, that fix is unrelated from the original report. 

Based on the user reports on #454 I did some investigation and turned out that the project synchronization is broken when you import the workspace root (a wrong project is created for the build root). There's not much we can do, it's due to the Eclipse workspace implementation. Even though the use-case is weird, looks like some people still try to do this and I think it makes sense to give an explicit saying it shouldn't be done.

oehme:
Got it. LGTM then

jsteinich:
Perhaps we should start a forum discussion, but I'm curious what the general convention is for buildship workspaces. Our team has been using the project root as the workspace for many years (buildship only recently). We did notice the oddities with an extra project being created, but that was only a minor annoyance.

oehme:
@donat Based on @jsteinich's comment this seems to work in other cases. Can you please investigate if we can make this work without creating a buggy project?

donat:
Sure, I'll dig deeper into this story.

donat:
Eclipse [prohibits](https://github.com/eclipse/eclipse.platform.resources/blob/6e968c80e74a60d1f407c00e044016f882b94313/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/LocationValidator.java#L364) creating projects at the workspace root. It is only a coincidence that Buildship worked for your case: the root project was not imported and instead a new, empty project was created. 

We could fix that by just skipping the root project import, but that would mess up other parts of the synchronization logic. Currently, Buildship assumes that all projects are in the workspace. If the subprojects reference the root project then not importing it might result in an incomplete classpath. 

Even though we don't have a fully feature documentation, but we already suggest to place their Gradle projects outside of the Eclipse workspace, because of it's problematic implementation. For example, if you rename a project in the workspace the project directory is physically moved on the disk, messing up CVS, the Gradle configuration and probably more. 

All in all, I think it's better to explicitly deny importing the workspace root, but if you have a good idea how can we work around that, then don't hesitate to share with us.

oehme:
Thanks for the link @donat, I agree with your conclusion.

jsteinich:
I think the difference is that we haven't treated the root as a project, though I certainly see the value in doing so. With our workflow, we'll basically just need another directory above the gradle root.

It seems a bit hacky, but perhaps the empty project that is currently created could utilize linked resources to pull in non-sub project files and folders.

oehme:
@jsteinich you could flip it around and nest the workspace folder under your project root instead of the other way round.

jsteinich:
Just tried nesting the workspace inside the root and got the following error:

`org.eclipse.buildship.core.GradlePluginsRuntimeException: org.eclipse.core.internal.resources.ResourceException: Invalid project description.
	at org.eclipse.buildship.core.workspace.internal.DefaultWorkspaceOperations.createProject(DefaultWorkspaceOperations.java:128)
	at org.eclipse.buildship.core.workspace.internal.SynchronizeGradleBuildOperation.addNewEclipseProjectToWorkspace(SynchronizeGradleBuildOperation.java:276)
	at org.eclipse.buildship.core.workspace.internal.SynchronizeGradleBuildOperation.synchronizeNonWorkspaceProject(SynchronizeGradleBuildOperation.java:259)
	at org.eclipse.buildship.core.workspace.internal.SynchronizeGradleBuildOperation.synchronizeGradleProjectWithWorkspaceProject(SynchronizeGradleBuildOperation.java:175)
	at org.eclipse.buildship.core.workspace.internal.SynchronizeGradleBuildOperation.access$000(SynchronizeGradleBuildOperation.java:99)
	at org.eclipse.buildship.core.workspace.internal.SynchronizeGradleBuildOperation$1.run(SynchronizeGradleBuildOperation.java:134)
	at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2240)
	at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2262)
	at org.eclipse.buildship.core.workspace.internal.SynchronizeGradleBuildOperation.synchronizeProjectsWithWorkspace(SynchronizeGradleBuildOperation.java:131)
	at org.eclipse.buildship.core.workspace.internal.SynchronizeGradleBuildOperation.run(SynchronizeGradleBuildOperation.java:115)
	at org.eclipse.buildship.core.workspace.internal.SynchronizeGradleBuildsJob.synchronizeBuild(SynchronizeGradleBuildsJob.java:81)
	at org.eclipse.buildship.core.workspace.internal.SynchronizeGradleBuildsJob.runToolingApiJob(SynchronizeGradleBuildsJob.java:72)
	at org.eclipse.buildship.core.util.progress.ToolingApiJob$1.run(ToolingApiJob.java:73)
	at org.eclipse.buildship.core.util.progress.ToolingApiInvoker.invoke(ToolingApiInvoker.java:62)
	at org.eclipse.buildship.core.util.progress.ToolingApiJob.run(ToolingApiJob.java:70)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:55)
Caused by: org.eclipse.core.internal.resources.ResourceException: Invalid project description.
	at org.eclipse.core.internal.resources.Project.checkDescription(Project.java:169)
	at org.eclipse.core.internal.resources.Project.assertCreateRequirements(Project.java:53)
	at org.eclipse.core.internal.resources.Project.create(Project.java:263)
	at org.eclipse.core.internal.resources.Project.create(Project.java:247)
	at org.eclipse.buildship.core.workspace.internal.DefaultWorkspaceOperations.createProject(DefaultWorkspaceOperations.java:113)
	... 15 more`

oehme:
I see, so Eclipse doesn't like that either. That sucks :/

PR #477, opened 2017-05-23
Streamline preferences UI

donat:
**Please don't merge this**! This PR only changes the following UI elements and this only for review purposes. I'll adjust the preferences functionality as part of the remaining stories of #466

PR #484, opened 2017-05-29
Streamline preferences UI and add missing configuration options

donat:
This PR supersedes #477 as it contains the all the changes from there plus it adjusts the preferences API and usage


org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/configuration/internal/BuildConfigurationPersistence.java
```diff
@@ -108,33 +110,47 @@ public void deletePathToRoot(File projectDir) {
...
-        }
+    private static DefaultBuildConfigurationProperties readPreferences(PreferenceStore preferences, File rootDir) {
         boolean overrideWorkspaceSettings = preferences.readBoolean(PREF_KEY_OVERRIDE_WORKSPACE_SETTINGS, false);
+
+        String distributionString = preferences.readString(PREF_KEY_CONNECTION_GRADLE_DISTRIBUTION, null);
+        GradleDistribution distribution = distributionString == null
+                ? GradleDistribution.fromBuild()
+                : GradleDistributionSerializer.INSTANCE.deserializeFromString(distributionString);
+
+        String gradleUserHomeString = preferences.readString(PREF_KEY_GRADLE_USER_HOME, "");
```
oehme: I'm still not terribly happy that we're reintroducing this, but I guess it is useful for local experimentation. It might be worth putting a ⚠️ next to the filed and say "This will make your project settings non-portable". Same for using a local Gradle distro.

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/workspace/AddBuildshipNatureHandler.java
```diff
@@ -58,7 +58,10 @@ public Object execute(ExecutionEvent event) throws ExecutionException {
         for (Object element : elements) {
             IProject project = adapterFunction.apply(element);
             if (project != null && !GradleProjectNature.isPresentOn(project) && project.getLocation() != null) {
-                projects.add(project);
+                IPath location = project.getLocation();
```
oehme: Something's wrong with this diff, this can't even compile.

Also, the location null check already happens above.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/configuration/internal/BuildConfigurationPersistence.java
donat: I added the warning to the UI. I wonder where exactly should we show it? Only in the project configuration page, or also in the run config page?

oehme: I think only in the project config is enough for now, because it feels more surprising there. I'd only show it when the user actually configures a path.

oehme:
LGTM

PR #485, opened 2017-06-01
Don't delete user-defined natures and builders

donat:


org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/preferences/internal/BuildCommandConverter.java
```diff
@@ -0,0 +1,151 @@
...
+
+import org.eclipse.core.resources.ICommand;
+import org.eclipse.core.resources.IProject;
+
+import org.eclipse.buildship.core.CorePlugin;
+
+/**
+ * Transforms build commands to XML format and vica versa.
+ */
+final class BuildCommandConverter {
```
oehme: Is there no API that we could reuse, like there was for classpath entries?

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/BuildCommandUpdater.java
```diff
@@ -29,39 +29,46 @@
...
  */
 final class BuildCommandUpdater {
 
-    public static void update(IProject project, Optional<List<OmniEclipseBuildCommand>> buildCommands, IProgressMonitor monitor) throws CoreException {
+    public static void update(IProject project, Optional<List<OmniEclipseBuildCommand>> buildCommands, PersistentModelBuilder persistentModel, IProgressMonitor monitor) throws CoreException {
         IProjectDescription description = project.getDescription();
-        Set<ICommand> commands = toCommands(buildCommands, description);
-        description.setBuildSpec(commands.toArray(new ICommand[0]));
+
+        Set<ICommand> current = Sets.newLinkedHashSet(Arrays.asList(description.getBuildSpec()));
```
oehme: ImmutableSet.copyOf() would probably read better here.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/BuildCommandUpdater.java
```diff
@@ -29,39 +29,46 @@
...
+        Set<ICommand> model = toCommands(buildCommands, description);
+        PersistentModel previousModel = persistentModel.getPrevious();
+        Set<ICommand> managed = previousModel.isPresent() ? Sets.newLinkedHashSet(previousModel.getManagedBuilders()) : Sets.<ICommand>newLinkedHashSet();
+
+        Result<ICommand> result = GradleBuilderAndNatureMergingStrategy.calculate(current, model, managed);
```
oehme: This might use a better name, maybe "ManagedModelMergingStrategy"?

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/GradleBuilderAndNatureMergingStrategy.java
```diff
@@ -0,0 +1,87 @@
...
+
+    /**
+     * Calculates the updated state.
+     *
+     * @param current elements currently defined on the project
+     * @param model elements defined in the Gradle model
+     * @param managed elements managed by Buildship
+     * @return the description of the updated state
+     */
+    public static <T> Result<T> calculate(Set<T> current, Set<T> model, Set<T> managed) {
```
oehme: The `Sets` class in Guava has union/intersection/difference methods that might be useful here.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/GradleBuilderAndNatureMergingStrategy.java
```diff
@@ -0,0 +1,87 @@
...
+
+        public Result(Set<T> nextElements, Set<T> nextManaged) {
+            this.nextElements = nextElements;
+            this.newManaged = nextManaged;
+        }
+
+        /**
+         * @return the updated elements that should be set on the project
+         */
+        public Set<T> getNextElements() {
```
oehme: maybe "merged" would be a better term than "next"?

donat:
@oehme Should we introduce a workspace or project settings to enable disable this new behavior? Personally, I like it better without any additional UI, but our users might think otherwise.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/preferences/internal/BuildCommandConverter.java
donat: I did another round on this to see if I've overlooked something. Turns out we'd need to use reflection to call the platform serialization (see [`ModelObjectWriter`](https://github.com/eclipse/eclipse.platform.resources/blob/6e968c80e74a60d1f407c00e044016f882b94313/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/ModelObjectWriter.java)).

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/GradleBuilderAndNatureMergingStrategy.java
donat: `getMergedElements()` seems fine, but `getMergedManaged()` feels a bit off. Any ideas?

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/BuildCommandUpdater.java
donat: Fixed

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/BuildCommandUpdater.java
donat: Renamed

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/GradleBuilderAndNatureMergingStrategy.java
donat: Makes sense, I've changed the implementation to use Guava. Now, the `calculate` method is much more readable.

PR #491, opened 2017-06-08
Prefer --scan argument for publishing build scans

donat:


org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/invocation/InvocationCustomizerTest.groovy
```diff
@@ -37,21 +42,51 @@ class InvocationCustomizerTest extends ProjectSynchronizationSpecification {
         new InvocationCustomizerCollector().extraArguments == EXTRA_ARGUMENTS
     }
 
-    def "Run configuration contains extra arguments"() {
+    def "Build configuration use extar arguments"() {
```
oehme: Extra

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/configuration/internal/ArgumentsCollector.java
```diff
@@ -0,0 +1,66 @@
...
+     *
+     * @param baseArgs the arguments that must be in the result
+     * @param buildScansEnabled whether {@code --scan} (or {@code -Dscan} for older Gradle
+     *            distributions) should be added to the result
+     * @param offlineMode whether {@code --offline} should be added to the resolt
+     * @param buildEnvironment the {@link BuildEnvironment} model to determine the current Gradle
+     *            version
+     * @return the arguments to pass to the TAPI
+     */
+    public static List<String> collectArguments(List<String> baseArgs, boolean buildScansEnabled, boolean offlineMode, BuildEnvironment buildEnvironment) {
```
oehme: I think we need a `GradleArguments` class to encapsulate this, see below for another place where I think that would help.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/BaseLaunchRequestJob.java
```diff
@@ -137,6 +139,16 @@ private String toNonEmpty(List<String> stringValues, String defaultMessage) {
         return string != null ? string : defaultMessage;
     }
 
+    private BuildConfiguration buildConfigurationFor(RunConfiguration runConfig) {
```
oehme: This is confusing, because we are using `BuildConfiguration` as a parameter holder here, which it is not intended for. Instead we should have:

```
BuildConfiguration {
  GradleArguments getGradleArguments();
}

RunConfiguration {
  GradleArguments getGradleArguments();
}

GradleArguments {
  void describeTo(Writer)
  void applyTo(GradleConnector)
  void applyTo(LongRunningOperation, BuildEnvironment)
}
```

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/DefaultGradleBuild.java
```diff
@@ -84,13 +85,15 @@ public BuildConfiguration getBuildConfig() {
     @Override
     public BuildLauncher newBuildLauncher(RunConfiguration runConfiguration, TransientRequestAttributes transientAttributes) {
         // TODO (donat) once GradleWorkspaceManager#getGradleBuild(FixedRequestAttributes) is removed then we should only allow run config that contain the same build config
-        return ConnectionAwareLauncherProxy.newBuildLauncher(runConfiguration.toRequestAttributes(), transientAttributes);
+        BuildEnvironment buildEnvironment = this.modelProvider.reloadBuildEnvironment(transientAttributes);
```
oehme: This should happen in the `ConnectionAwareLauncherProxy` instead. `GradleBuild` and `ModelProvider` shouldn't need to know about this.

donat: So `ConnectionAwareLauncherProxy` should call `ModelProvider` to get the built environment? Seems like a stack overflow problem as `ModelProvider` will call back to `ConnectionAwareLauncherProxy`.

oehme: No, there is no need to call back into `ModelProvider`. We don't need caching/fetch strategies etc. for the `BuildEnvironment`.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/BaseLaunchRequestJob.java
donat: If we hide the `BuildEnvironment` model query behind the `ConnectionAwareLauncherProxy` and replace the fake `BuildConfiguration` with `GradleArguments` then we lose access to the `GradeEnvironment` here in the `BaseLaunchRequest` class (and won't be able to present the used Gradle version for instance). 

We could introduce the `GradleWorkspaceManager.getGradleBuild(GradleArguments)` but that seems way more work than it should be. Any ideas how to get around that?

oehme: Why would we loose access to the `GradleEnvironment`?

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/invocation/InvocationCustomizerTest.groovy
donat: Fixed

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/configuration/internal/ArgumentsCollector.java
donat: Fixed

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/BaseLaunchRequestJob.java
donat: Ignore my comment above.
GradleArguments added to BuildConfiguration and to RunConfiguration

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/DefaultGradleBuild.java
donat: Implemented

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/ConnectionAwareLauncherProxy.java
```diff
@@ -57,62 +57,54 @@ private ConnectionAwareLauncherProxy(ProjectConnection connection, LongRunningOp
...
         applyTransientAttributes(operation, transientAttributes);
     }
 
-    private static void applyConfiguration(LongRunningOperation operation, RunConfiguration runConfiguration, BuildEnvironment buildEnvironment,
-            TransientRequestAttributes transientAttributes) {
-        runConfiguration.applyTo(operation, buildEnvironment);
+    private static void applyConfiguration(LongRunningOperation operation, GradleArguments gradleArguments, BuildEnvironment buildEnvironment,
+            Writer configWriter, TransientRequestAttributes transientAttributes) {
+        gradleArguments.applyTo(operation, buildEnvironment);
+        gradleArguments.describe(configWriter, buildEnvironment);
```
oehme: This should have stayed in `BaseLaunchRequestJob`. We don't want every single build invocation to require a `Writer` and print the config.

donat: I'm a bit confused on this. I moved the `Writer` argument to where it is now because that's where we also moved the `BuildEnvironment` query. 

We can query the `BuildEnvironment` in `BaseLaunchRequest`, but for that, we need to convert the `RunConfiguration` to a `BuildConfiguration`. This is what the `buildConfigurationFor()` method did.

oehme: I see the problem now. I think as a next cleanup step, we should merge this "configWriter" and "transientRequestAttributes" into a new type (don't have a good name yet), to get rid of TransientRequestAttributes and make the argument list shorter again. The new type would contain the process streams (including the one for output, error, input and configuration messages), progress listeners etc.

donat: I created a new ticket for that: https://github.com/eclipse/buildship/issues/498

PR #492, opened 2017-06-09
Don't show error dialog when GradleConnectionException is received

donat:
The build launcher proxies used for executing builds produce
InvocationTargetException when a GradleConnectionException is
received from the TAPI. This exception is unknown to the
ToolingApiInvoker, therefore a general error dialog is displayed
for the user.

To make the exception handling behave correctly, this commit safely
unwraps and rethrows the root cause of the InvocationTargetException.


donat:
I just added the null check to comply with the Javadoc. The `invoke()` method's logic is mostly implemented in a native method, so I think it's better to play it safe and keep the null-check.

PR #503, opened 2017-06-27
Suppress error dialog when model synchronization is cancelled

donat:
When the user cancels a project synchronization, then he is presented with an error dialog. The root cause is the Guava cache we use for the model loading which throws a wrapper exception unknown by `ToolingApiInvoker`.  This pull request fixes the issue by unwrapping the exception and rethrowing the root cause.


org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/DefaultModelProvider.java
```diff
@@ -155,7 +153,12 @@ public T call() {
             U result = (U) this.cache.get(cacheKey, cacheValueLoader);
             return result;
         } catch (Exception e) {
-            throw new GradlePluginsRuntimeException(e);
+            if (e instanceof UncheckedExecutionException && e.getCause() instanceof RuntimeException) {
```
oehme: I don't think this needs further explanation, it's just the general pattern when you use a LoadingCache.

PR #507, opened 2017-07-03
Further improvements on the Executions view performance

donat:
I did some final testing for the 2.1.0 release and found that the Executions view was still not performant enough for larger projects. I've added two changes to the implementation which drastically improves the situation. The details are in the individual commits' messages.

I think we should include these changes for the release to provide a smoother experience in the release.

### DIFF
--- a/org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/view/execution/UpdateExecutionPageJob.java
+++ b/org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/view/execution/UpdateExecutionPageJob.java
@@ -11,15 +11,17 @@
 
 package org.eclipse.buildship.ui.view.execution;
 
+import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentMap;
 
 import org.gradle.tooling.events.FailureResult;
 import org.gradle.tooling.events.FinishEvent;
 import org.gradle.tooling.events.task.TaskOperationDescriptor;
 
 import com.google.common.base.Preconditions;
-import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Sets;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.MapMaker;
 
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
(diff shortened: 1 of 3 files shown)


org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/view/execution/UpdateExecutionPageJob.java
```diff
@@ -68,19 +78,12 @@ public void run() {
...
-        }
+        this.operationItems.add(operationItem);
     }
 
     public void removeOperationItem(OperationItem operationItem) {
         Preconditions.checkNotNull(operationItem.getStartEvent());
-        synchronized (this.operationItems) {
-            this.operationItems.remove(operationItem.getStartEvent().getDescriptor());
-            if (this.operationItems.isEmpty()) {
-                stop();
```
oehme: ❌ We've lost the stop call here.

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/view/execution/UpdateExecutionPageJob.java
```diff
@@ -34,15 +33,18 @@
 
     private static final int REPEAT_DELAY = 100;
 
-    private final Map<OperationDescriptor, OperationItem> operationItems;
+    private final Set<OperationItem> operationItems;
```
oehme: needs a better name, e.g. "activeItems"

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/view/execution/UpdateExecutionPageJob.java
```diff
@@ -34,15 +33,18 @@
...
+    private final Set<OperationItem> removeableItems;
```
oehme: removedItems

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/view/execution/UpdateExecutionPageJob.java
```diff
@@ -68,19 +78,12 @@ public void run() {
 
     public void addOperationItem(OperationItem operationItem) {
```
oehme: start instead of add?

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/view/execution/UpdateExecutionPageJob.java
```diff
@@ -68,19 +78,12 @@ public void run() {
...
         Preconditions.checkNotNull(operationItem.getStartEvent());
-        synchronized (this.operationItems) {
-            this.operationItems.put(operationItem.getStartEvent().getDescriptor(), operationItem);
-        }
+        this.operationItems.add(operationItem);
     }
 
     public void removeOperationItem(OperationItem operationItem) {
```
oehme: stop instead of remove?

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/view/execution/UpdateExecutionPageJob.java
```diff
@@ -34,15 +33,18 @@
...
+
     private volatile boolean running;
     private final TreeViewer treeViewer;
 
     public UpdateExecutionPageJob(TreeViewer treeViewer) {
         super("Updating duration of non-finished operations");
 
         this.treeViewer = treeViewer;
-        this.operationItems = Maps.newHashMap();
+        this.operationItems = Sets.newConcurrentHashSet();
```
oehme: A general caution about concurrency: Using two concurrent sets doesn't guarantee safety when these two sets have a combined meaning/need to be coordinated. In this special case it is probably fine because we know each object is unique and the started and end events always come from the same thread in the right order. But I haven't thought too deeply about it, so there might still be a race condition I'm missing.

Unless the synchronization overhead is actually a measurable problem, I would suggest to go with a simple synchronized block around the whole access to both of these fields.

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/view/execution/UpdateExecutionPageJob.java
donat: I deleted it because `ExecutionProgressListener` calls the stop method already when the build invocation job [finishes](https://github.com/eclipse/buildship/blob/1af4ae07845969bedfe95f24dc75e9b886a276b4/org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/view/execution/ExecutionProgressListener.java#L50).

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/view/execution/UpdateExecutionPageJob.java
donat: Fixed

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/view/execution/UpdateExecutionPageJob.java
donat: Fixed

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/view/execution/UpdateExecutionPageJob.java
donat: Fixed.

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/view/execution/UpdateExecutionPageJob.java
donat: Fixed

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/view/execution/UpdateExecutionPageJob.java
donat: Thanks for the explanation. Unfortunately, the synchronization does cause measurable overhead. I've rewritten the class such that only one `ConcurrentMap` is used and the possible race conditions are eliminated.

donat:
@oehme I've addressed your review items, please have a look.

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/view/execution/UpdateExecutionPageJob.java
oehme: Does it really cause overhead if you only synchronize the access to the two sets? I.e. keep the UI update out of the synchronized block?

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/view/execution/UpdateExecutionPageJob.java
```diff
@@ -12,75 +12,76 @@
...
  */
 public final class UpdateExecutionPageJob extends Job {
 
     private static final int REPEAT_DELAY = 100;
 
-    private final Map<OperationDescriptor, OperationItem> operationItems;
-    private volatile boolean running;
     private final TreeViewer treeViewer;
+    // if the value is true then the operation can be removed
+    private final ConcurrentMap<OperationItem, Boolean> activeItems;
```
oehme: I still feel like this could be simpler. We could have a Queue of events and then just forward those events to the UI element. The UI element would then react to:

- start events by adding them to the "active operations"
- finish events by removing them from the "active operations"
- timer by updating all active event's time

donat: Well, the code became longer with the queue, but in fact much more readable imo.

donat:
Done.

PR #529, opened 2017-07-18
Restore compatibility with Spring Tool Suite

donat:
This pull request restores the compatibility with existing STS releases by doing two things:

- Use the Guava version from STS
- Restore deleted class used by STS

I've tested the changes locally by installing STS, installing a local build of Buildship into it and creating a new Spring project and it seemed working.


donat:
I'll inform the `jdt.ls` team.

I remember that `RateLimiter` class was reported mostly, but there could be others too. If we use version ranges we should accept that tooling-commons and Buildship pick up a different Guava version, we should carefully investigate all usages of Guava in our codebase.

oehme:
That's what we have tests for :) if we test the lower bound and the upper bound and all tests pass then we're good.

Issue #520, opened 2017-07-12
Build on subproject builds complete multi-project


andilem:
After upgrading to Buildship 2.1.0, when building a subproject from the Gradle Tasks view, not only the subproject is built, but instead the complete multi-project.

When inspecting the run configuration, it seems like the working directory (which is set correctly) is **not** used during the task execution (the console output shows the root project location as working directory).

Expected behavior: Only the selected subproject is built (as in 2.0.2).

Tested with Gradle 3.5 and Gradle 4.0.

oehme:
Thank you for the detailed description! We'll fix this in the next patch release.

PR #531, opened 2017-07-19
Use correct project dir when executing tasks

donat:
This commit fixes a regression introduced in Buildship 2.1. The
new preferences API ignored the project location and always
executed all tasks on the root project. As a result, task selectors
were always executed on all projects, even if they were started from
a subproject.

The fix changes the RunConfiguration interface such that it references
a project config as opposed to a build configuration.

This commit fixes https://github.com/eclipse/buildship/issues/520


donat:
Test coverage added

PR #533, opened 2017-07-20
Project configuration can read absolute root project locations again

donat:
Until 2.1 Buildship accepted absolute root project location in the
import preferences.

### DIFF
--- a/org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/configuration/internal/ProjectConfigurationTest.groovy
+++ b/org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/configuration/internal/ProjectConfigurationTest.groovy
@@ -1,5 +1,7 @@
 package org.eclipse.buildship.core.configuration.internal
 
+import spock.lang.Issue
+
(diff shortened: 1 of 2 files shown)


org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/configuration/internal/ProjectConfigurationTest.groovy
```diff
@@ -237,6 +239,43 @@ class ProjectConfigurationTest extends ProjectSynchronizationSpecification {
...
+
+        when:
+        configurationManager.saveProjectConfiguration(projectConfig)
+
+        then:
+        configurationManager.buildConfigurationPersistence.readPathToRoot(project) == "../$rootProjectDir.name"
+    }
+
+    @Issue('https://github.com/eclipse/buildship/issues/528')
+    def "can save and load project configuration if project is clased and settings file contains absolute path"() {
```
oehme: closed

donat: fixed

PR #537, opened 2017-08-01
Ignore invalid project configs upon synchronization

donat:


org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/DefaultGradleWorkspaceManager.java
```diff
@@ -61,12 +62,18 @@ public GradleBuild getGradleBuild(BuildConfiguration buildConfig) {
...
+            try {
+                if (GradleProjectNature.isPresentOn(project)) {
+                    ProjectConfiguration projectConfiguration = CorePlugin.configurationManager().loadProjectConfiguration(project);
+                    BuildConfiguration buildConfig = projectConfiguration.getBuildConfiguration();
+                    return Optional.<GradleBuild>of(new DefaultGradleBuild(buildConfig));
+                } else {
+                    return Optional.absent();
+                }
+            } catch (RuntimeException e) {
+                CorePlugin.logger().debug("Cannot load configuration for project " + project.getName());
```
oehme: This should be encapsulated instead of spreading the error handling in multiple places. E.g. tryLoad

donat: I'm wondering if we should use Guava's `Optional` here. I can imagine that this might break yet another third party plugin that uses a more recent version of Guava.

donat: Because of that, I've implemented `tryLoadProjectConfiguration` such that it returns `null` when the config can't be loaded.

Issue #511, opened 2017-07-05
"Cannot run tasks for included builds"


sc1478:
Buildship is telling me it can't do the included builds.

![image](https://user-images.githubusercontent.com/15382550/27868919-d9205274-6163-11e7-8374-6a0c20766625.png)

Yet, if I go to the command line in the same directory and run `./gradlew clean`, gradle has no trouble doing the job.  I tried refreshing the Gradle project from the Eclipse menu, that did not help

Gradle version 4.0
Buildship version   Buildship: Eclipse Plug-ins for Gradle	2.0.1.v20170302-1039	org.eclipse.buildship.feature.group	Eclipse Buildship

Eclipse tells me no updates are available.

donat:
This is a limitation of Gradle itself, as you can't directly address tasks from the included builds.
Basically, when you execute `gradle foo` on a composite build, then those tasks will be executed from the included builds that produce artifacts required by `foo`. You'll find more information in the [user guide](https://docs.gradle.org/4.0/userguide/composite_builds.html#composite_build_executing_tasks).

sc1478:
This is not a composite build.  It is a straightforward multi-project build.

donat:
Nope, the tasks should be disabled only for included builds. I'll have a look how could you end up in this phase. Could you create a sample project exhibiting the problem so that we can use it for the investigation?

oehme:
This usually happens when a user runs `Add Gradle Nature` on a subproject. Our included build detection currently compares the connectionProjectDir against the root project dir, which fails in this case. 

Our detection could instead use the `GradleBuild` model to determine whether a project is from an included build or not. 

Also, we should really finally add proper root project detection, so we use that as the connectionProjectDir, even when the user selects a subproject when adding the nature.

sc1478:
Also, your reply that "this is a limitation of gradle itself" seems to ignore my initial statement 

> Yet, if I go to the command line in the same directory and run ./gradlew clean, gradle has no trouble doing the job. 

Do you still need a sample from me?  I'm not clear at this point.

sc1478:
Is something needed from me to make progress on this issue?

fbazbo:
I'm seeing the same issue, but it was working fine until I just upgraded from Gradle 3.5 to 4

sc1478:
Yes, it's not an issue when using 3.x gradle wrapper, only with 4.

fbazbo:
[...]

PR #539, opened 2017-08-02
Fix visual enablement and context menu messages for disabled tasks

donat:
This PR fixes the issues discussed in #511.

I'm not quite happy with the implementation, but I'd like to discuss the design in the review before doing more work.


oehme:
What makes you unhappy about the implementation? I think an enum value of the different reasons is a good idea.

donat:
I had a second look on the PR. Afterall I found that only the naming was confusing, the rest is fine.

Issue #514, opened 2017-07-11
Add editor support with basic highlighting capabilities for build scripts


cmoine:
My suggestion may seem naive, but never mind: why not integrating a fork of as a starting point for a gradle editor  such as https://github.com/Nodeclipse/GradleEditor ?
I think the MIT license shoudl allows this.

Btw: I can provide a PR  for that

donat:
I  think having a basic syntax highlight is much better than using a plain text editor for Gradle files. If the implementation is simple and robust enough then please go ahead and create the pull request.
As of the licensing, I'll handle the IP clearance with the Eclipse Foundation. The MIT license should be fine here.

cmoine:
Is there someone to integrate my last PR please ? https://github.com/eclipse/buildship/pull/559

donat:
Hi! Sorry for the slow response, I've just got back from vacation. I'll try to look into your PR in the upcoming days.

cmoine:
No pb, I was just wondering if it was ok from my side ! :) Don't hesitate. Good luck for coming back to work ;)

jimmytheneutrino:
Hi!

Awesome work on the editor! 

I wonder if it will also support some auto completion?

cmoine:
Hi, thank you very much. Probably in a second release of the editor ;) Step by step :)

jimmytheneutrino:
Sure:) Thank you for doing this.

cmoine:
No problem. I love Gradle, and love Eclipse too... But a Gradle editor was definitely missing I think :) 

I will try to address this after my PR is accepted ;)

donat:
PR merged, thank you for the contribution! We'll probably publish a post on the forum to let the users try this feature out.

cmoine:
Thank you very much. That was a pleasure. If I can help on further improvements. I would like to have a first feedback from end user before starting with basic completion, and so on.

PR #559, opened 2017-08-29
#514 Integration of https://github.com/Nodeclipse/GradleEditor

cmoine:
Signed-off-by: Christophe Moine <christophe.moine@free.fr>

### DIFF
--- a/org.eclipse.buildship.ui/META-INF/MANIFEST.MF
+++ b/org.eclipse.buildship.ui/META-INF/MANIFEST.MF
@@ -22,6 +22,7 @@ Require-Bundle: org.eclipse.buildship.core,
  org.eclipse.jface.text,
  com.ibm.icu,
  com.google.guava,
+ org.eclipse.ui.genericeditor,
  com.gradleware.tooling.model;bundle-version="[0.19.2,0.20.0)",
  com.gradleware.tooling.client;bundle-version="[0.19.2,0.20.0)",
  com.gradleware.tooling.utils;bundle-version="[0.19.2,0.20.0)"
--- a/org.eclipse.buildship.ui/plugin.xml
+++ b/org.eclipse.buildship.ui/plugin.xml
@@ -65,12 +65,13 @@
     <extension
             point="org.eclipse.ui.editors">
       <editor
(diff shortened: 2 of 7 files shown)


donat:
Hi, @cmoine! Sorry for the late reply, I was away for a few weeks. From now on I'll get back to you much faster.

One thing we discussed internally is that the same editor support can be achieved with Eclipse's [generic editor](https://www.eclipse.org/eclipse/news/4.7/M3/#generic-editor). That way Buildship wouldn't have to define yet another editor for Gradle files. The only problem is that the generic editor is available since Eclipse 4.7, so older Eclipse releases still wouldn't have any syntax highlighting. That being said, I still think your PR is valuable, I just wanted to point out the possibility.

donat:
In the meanwhile, I've contacted the Eclipse IP team to verify if we are allowed to merge this PR. Most probably we are, but I'd like to be 100% sure.

cmoine:
Wahou, Generic Editor feature looks amazing !!! :O And funny thing, the example is a Gradle editor !!!!! :D

> The only problem is that the generic editor is available since Eclipse 4.7
But my PR is on master, hence for Eclipse 4.7, isn't it ? Did you plan to merge my PR in older branches ?

I already tried to contact someone, but I don't remember who, I cannot find the message again :( (did I dream ? :) )

To be honest I would prefer the Generic Editor for the following reasons:
- We go with something fresh, and modern (I hope !) :)
- We won't have any licensing issues (even if you have confirmation, we never know what can do some other people...)
- We may bring an added value compared to the STS plugin editor: it is not just copy/paste.
- That would make it more homogeneous to provide a Gradle editor just like eclipse suggest to do it in the News  & NoteWorthy
- That is interesting from my point of view ;) (more than copy/pasting code :) )

What do you think ?

donat:
I'd also prefer an implementation with the generic editor, it's only up to you if you want to invest time into the implementation. And thumbs-up for the enthusiasm :)

Feel free to contact us if you need any help with the existing codebase. Also, in case you haven't found it, here's the [Buildship development documentation](https://github.com/eclipse/buildship/blob/master/docs/development/Setup.md).

Issue #560, opened 2017-08-30
Provide a way to "Refresh Gradle Project" directly from the Gradle editor


cmoine:
Following the #514 , we could add a button refresh in the toolbar when the editor is opened to "Refresh Gradle Project", instead of searching the project, right clicking on it, ....

A can also provide a PR for that ;)

cmoine:
I just found out we could right click on the editor to refresh. We can lower the priority then :)

PR #563, opened 2017-09-08
#514 and #560

cmoine:
I have used the genericeditor. It was almost too simple :)

Appart from the joke, I wanted to use a more sophisticated highlighter rules... For example "task" is defined as a keyword but it is not. It is related to Gradle DSL.... So if you invoke a task like for instance:
> myTask task: 'anotherTask'
task will be highlighted :( It seems we need some kind of semantic parser to do that. I tried (https://github.com/danielsun1106/groovy-parser), but I am not sure parsing the document for highlight syntax is good (I think I can imagine the answer :D ).

If you of someone else have experience on that, I am very interested with these problematic :)

I know Gradle DSL is complex, and it is probably the reason why there is not bundled Gradle editor in Eclipse :D Even if Eclipse seem to bet more on Kotlin DSL to provide an editor, I don't think it will be quickly the most used editor...

It is why I kept it simple so far :)

I also took the opportunity to add a EditorContributor for #560 :)

### DIFF
--- a/org.eclipse.buildship.ui/META-INF/MANIFEST.MF
+++ b/org.eclipse.buildship.ui/META-INF/MANIFEST.MF
@@ -22,6 +22,7 @@ Require-Bundle: org.eclipse.buildship.core,
  org.eclipse.jface.text,
  com.ibm.icu,
  com.google.guava,
+ org.eclipse.ui.genericeditor,
  com.gradleware.tooling.model;bundle-version="[0.19.2,0.20.0)",
  com.gradleware.tooling.client;bundle-version="[0.19.2,0.20.0)",
  com.gradleware.tooling.utils;bundle-version="[0.19.2,0.20.0)"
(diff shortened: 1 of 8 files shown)


donat:
Hi @cmoine! I see you close and reopen PRs. I suppose you do this to updated the changes. Why not updating the existing PRs with a force-push(`git push -f`)?

cmoine:
I am actually doing a force push, but I thought I had to recreate a PR. Also I had a problem using my professional email, but the sanity checks was failing :( I am sorry about that. I didn't know, and I will keep that PR. I don't master all subtleties of PRs yet, thank you for that :)

donat:
Don't worry, I'm just trying to help here :)
As of the sanity check, here are the rules to comply to: https://wiki.eclipse.org/Development_Resources/Contributing_via_Git#via_GitHub

PR #564, opened 2017-09-08
#514 Add Gradle editor

cmoine:
I have used the genericeditor. It was almost too simple :)

Appart from the joke, I wanted to use a more sophisticated highlighter rules... For example "task" is defined as a keyword but it is not. It is related to Gradle DSL.... So if you invoke a task like for instance:
> myTask task: 'anotherTask'

task will be highlighted :( It seems we need some kind of semantic parser to do that. I tried (https://github.com/danielsun1106/groovy-parser), but I am not sure parsing the document for highlight syntax is good (I think I can imagine the answer :D ).

If you of someone else have experience on that, I am very interested with these problematic :)

I know Gradle DSL is complex, and it is probably the reason why there is not bundled Gradle editor in Eclipse :D Even if Eclipse seem to bet more on Kotlin DSL to provide an editor, I don't think it will be quickly the most used editor...

It is why I kept it simple so far :)

I also took the opportunity to add a EditorContributor for #560 :)


cmoine:
Sorry, I setup a new environment properly with Oomph :) 

I tried also to rebase on master branch, I hope there will be no trouble with that... :O

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/preferences/GradleEditorPreferencePage.java
```diff
@@ -0,0 +1,68 @@
...
+import org.eclipse.jface.preference.IPreferenceStore;
+import org.eclipse.swt.widgets.Composite;
+import org.eclipse.ui.IWorkbench;
+import org.eclipse.ui.IWorkbenchPreferencePage;
+
+/**
+ * @author Benjamin gurok
+ * @author Paul Verest
+ */
+public class GradleEditorPreferencePage extends FieldEditorPreferencePage implements IWorkbenchPreferencePage {
```
donat: We have a simple editor and thus we should keep it simple. How about we
- remove the preference page
- reuse the colors  from the Java editor (`org.eclipse.jdt.ui.PreferenceConstants`)
- Add a preference listener to update the colors when the Java editor colors changed.
This way we can provide a consistent color scheme without adding more complexity to the UI.

cmoine:
Ok, no pb. Do you mind if I create a new PR ? I would like to create a clean new branch on my fork.

org.eclipse.buildship.ui/plugin.xml
```diff
@@ -65,12 +65,13 @@
     <extension
             point="org.eclipse.ui.editors">
       <editor
-            id="org.eclipse.buildship.ui.gradlebuildscripteditor"
-            class="org.eclipse.ui.editors.text.TextEditor"
-            icon="icons/full/obj16/gradle_file.png"
-            name="Gradle Build Script Editor"
+            class="org.eclipse.ui.internal.genericeditor.ExtensionBasedTextEditor"
```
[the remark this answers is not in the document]
cmoine: Hmm, good point, but I don't see other way to define an icon for the editor, and most important: an editorcontributor to add "Refresh Gradle Project" action in the toolbar.

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/preferences/GradleEditorPreferencePage.java
cmoine: I hesitated reusing Java Colors. I will see if that fits.

donat:
Aure, go ahead. Actually, it makes sense as changing this PR might mess up the review comments.

cmoine:
It seems that colors for Java Syntax depends on the used Theme. With dark theme, keywords are orange instead of violet. Do you know how to take the Theme into account ? I had a look yesterday, but I haven't found in a "reasonable" timing :) If you already know, I would be very interested, thanks.

donat:
You can attach a listener to the JDT UI plugin preferences. Here's an example:
```
EclipsePreferencesUtils.getInstanceScope().getNode(JavaUI.ID_PLUGIN)
    .addPreferenceChangeListener(new IPreferenceChangeListener() {

    @Override
    public void preferenceChange(PreferenceChangeEvent event) {
        if (event.getKey().equals(PreferenceConstants.EDITOR_JAVA_KEYWORD_COLOR)) {
           System.out.println(JavaUI.getColorManager()
               .getColor(PreferenceConstants.EDITOR_JAVA_KEYWORD_COLOR));
        }
    }
});
```
A small note: The `EclipsePreferencesUtils` is a utility class from Buildship.

donat:
Can we help anything regarding this PR?

cmoine:
Thanks but I haven't much time this week, but don't worry, I'm on it :)

Also the current implementation of the PresentationReconcilier don't support the addPreferenceChangeListener.
I made I short test:
```
UiPlugin.getInstance().getPreferenceStore().addPropertyChangeListener(new IPropertyChangeListener() {
            @Override
            public void propertyChange(PropertyChangeEvent event) {
                if (event.getProperty().equals(GradleEditorConstants.KEY_COLOR_KEYWORD)) {
                    GradlePresentationReconciler.this.keywordToken.setData(new TextAttribute(Display.getDefault().getSystemColor(SWT.COLOR_DARK_YELLOW), null, UiPlugin.getInstance().getPreferenceStore().getBoolean(GradleEditorConstants.KEY_BOLD_KEYWORD) ? SWT.BOLD : SWT.NORMAL));
                }
            }
        });
```
But as expected, the setData with a new TextAttribute doesn't refresh the editor...

JDT use setData with string (token ids) instead of TextAttribute, and I need to use something similar I guess.

cmoine:
Ok, I made some progress, but it seems a long way home :(

It seems that we need a "Reconcilier". The PresentationReconcilier is not sufficient. unfortunately, only genericeditor v1.1 let us define a Reconcilier... Either I need to get ride of genericeditor, or we need to migrate to genericeditor 1.1. What do you think ?

cmoine:
Hmm, sorry, but it seems that genericeditor it too far from what you expect: the editor doesn't override affectsTextPresentation which allows colors changing "on the fly"

donat:
I think it would be already fine if the editor colors would be updated after the editor is closed and reopened. Theme updates are not that frequent anyway. 

But there's yet another problem with the generic editor. Since the dependency is relatively new, we can only ship with it Buildship only for the latest version. So, if we set the `org.eclipse.buildship.ui.genericeditor` dependency to 1.1, then the editor support will be available in Eclipse 4.8+, maybe 4.7.1, I'm not 100% sure (I'd adjust the build scripts, you don't have to worry about that).

That being said, I think I'm leaning towards going back to the initial idea of having a discrete editor definition. That way we can support previous Eclipse versions and we have full control over the color preferences. How does that sound to you?

In any case, thanks for sticking with us! We are super grateful that you invest your time into this feature.

cmoine:
I totally agree. My experiments with genericeditor are not conclusive unfortunately :(

cmoine:
I love Eclipse, but... the Syntax Hightlighthing mechanism with Reconcilier/Damager/Partitioner is just a nightmare !!!
Do you mind if I use the internal org.eclipse.jdt.internal.ui.text.AbstractJavaScanner ?

cmoine:
Damned, I found the magic method to make it work with genericeditor:
`viewer.invalidateTextPresentation()`
Sorry for the n-th change, but I recommend to use generic editor: Defining our own editor/presenter/reconcilier require a LOT OF CODE, and sometimes references to internal Java Plugin if we don't want to copy paste too much :(

What do you think ?

donat:
I'd prefer genericeditor if we can use `viewer.invalidateTextPresentation()` and if we don't need the 1.1 version just to update the colors upon the theme changes.

cmoine:
omg, spent one day figuring out with Generic Editor doesn't work... It is definitely doesn't fit. The partitionning/damager/reconciler is definitely complex stuff, and Generic Editor doesn't fit. I will switch back the plain old implementation again :'-( (without generic editor) If somebody interested, the limitation comes from this (in SourceViewerConfiguration), which is not properly extendable within GenericEditor:
```
public String[] getConfiguredContentTypes(ISourceViewer sourceViewer) {
	return new String[] { IDocument.DEFAULT_CONTENT_TYPE };
}
public String getConfiguredDocumentPartitioning(ISourceViewer sourceViewer) {
	return IDocumentExtension3.DEFAULT_PARTITIONING;
}

```
Now I think I understand why there is not Gradle editor :D

cmoine:
I am almost ready to push the "non generic editor" solution, is it ok ?

donat:
Sure, please go ahead.

cmoine:
Pushed.

cmoine:
Sorry for my clumpsy usage of Git :(

donat:
Thanks for the change, I'll have a look shortly. 

In the meanwhile, here's what I usually do on the command line when  I want to produce a PR with a single commit:
    
    # count the commits I have on the local branch; say there were 3
    git rebase -i head~3 # interactive rebase for the first 3 commits
    # in the git rebase editor I leave the first entry on 'pick' and change the rest to 'squash'
    git rebase master
    git push -f

cmoine:
I find out my mistake with Git: I told that I woudl create a new PR :)

cmoine:
New PR: https://github.com/eclipse/buildship/pull/585

PR #585, opened 2017-09-29
Basic editor support for build scripts

cmoine:
Follows up https://github.com/eclipse/buildship/pull/564

### DIFF
--- a/org.eclipse.buildship.core/plugin.xml
+++ b/org.eclipse.buildship.core/plugin.xml
@@ -78,4 +78,12 @@
        </runtimeClasspathEntryResolver>
     </extension>
 
+    <extension point="org.eclipse.core.contenttype.contentTypes">
+      <content-type
+           file-extensions="gradle"
+           id="org.eclipse.buildship.core.files.gradlebuildscript"
+           name="Gradle Build Script"
+           priority="low"
+      />
+   </extension>
 </plugin>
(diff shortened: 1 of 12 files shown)


cmoine:
Hi Donat, 

Thank you very much for your careful testing. That is strange that tab doesn't work.

I think (hope) it shouldn't be too complicated.

I will rest the weekend first :)

Cheers.


----- Original Message -----

donat:
Sorry, if my previous message sounded a bit pushy. Take your time and let us know if we can help.

cmoine:
Ah, no, I didn't take it that way, no pb :)

The method TextSourceViewerConfiguration.getTabWidth was returning zero because of wrong PreferenceStore :) Don't hesitate if you find other problems.

Cheers.

cmoine:
I also changed the @author into Contributor into the header comment, because I saw it was like that in other files :)

donat:
About the license header: What you found is the old header. If you have a look at the newer source files (like [this](https://github.com/eclipse/buildship/blob/f404ed1ce72823a998f73586545b5b0e4d0fac18/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/preferences/ModelPersistence.java)) then you'll find the proper header.

cmoine:
I think it is because of my mistake when I did a merge with an old branch. I forced a push: I hope it will be much better.

cmoine:
Great ! Ouch, I forgot to replace to @author back. Do you want me to do it ?

donat:
Yes, it would be better if you do it :)

cmoine:
It would be a shame to give up at this point :) Done !

PR #596, opened 2017-10-17
Separate runtime classpath with scope information provided by Gradle

donat:


org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/launch/ClasspathSeparationTest.groovy
```diff
@@ -0,0 +1,206 @@
...
+        setup:
+        importAndWait(createSampleProject('sample-project'), GradleDistribution.forLocalInstallation(new File('/Development/git/gradle/gradle/build/distributions/gradle-4.4-20171016160000+0000')))
+
+        when:
+        launchAndWait(createJavaLaunchConfiguration('sample-project', 'pkg.Main'))
+
+        then:
+        assertConsoleOutputContains('pkg.Main available')
+        assertConsoleOutputContains('pkg.JunitTest inaccessible')
+        assertConsoleOutputContains('org.apache.commons.io.IOUtils available')
```
oehme: ❌ This shouldn't be available, it's a `compileOnly` dependency. We should probably only use the `runtimeClasspath` to compute the `gradle-used-by-scope` attribute after all.

org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/launch/ClasspathSeparationTest.groovy
```diff
@@ -0,0 +1,206 @@
...
+// TODO (donat) enable the test when there's a Gradle snapshot release that includes the classpath separation changes
+// TODO (donat) adjust the Gradle distributions used by the test methods
+@IgnoreIf({ !OperatingSystem.current.isMacOs() })
+class ClasspathSeparationTest extends SwtBotSpecification {
+
+    def cleanup() {
+        DebugPlugin.default.launchManager.launchConfigurations.each { it.delete() }
+    }
+
+    def "Gradle doesn't supply scope information"() {
```
oehme: ❌ Please give tests a name that reflects what they are testing. "I test that Gradle doesn't supply scope information" is probably not what this test should be saying.

donat:
@oehme Just to be sure, this should be deleted: https://github.com/gradle/gradle/blob/1c1cf4f644935b771d865bee72ed848d0221858e/subprojects/ide/src/main/java/org/gradle/plugins/ide/eclipse/model/internal/EclipseDependenciesCreator.java#L130
I suppose we should also remove this one:
https://github.com/gradle/gradle/blob/506c85739651292c37671f86cd1d1c78d26d5949/subprojects/ide/src/main/java/org/gradle/plugins/ide/eclipse/model/internal/SourceFoldersCreator.java#L206

oehme:
Yes, I think that would be more accurate.

org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/launch/ClasspathSeparationTest.groovy
donat: Fixed

org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/launch/ClasspathSeparationTest.groovy
donat: I adjusted the implementation in Gradle core and updated the test assertions.

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/launch/internal/GradleClasspathProviderUpdateTest.groovy
```diff
@@ -0,0 +1,129 @@
...
+        }
+
+        when:
+        importAndWait(projectDir)
+
+        then:
+        hasGradleClasspathProvider(launchConfiguration)
+    }
+
+    def "Gradle classpath provider not not added for new jon-Java project"() {
```
oehme: Typos

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/internal/GradleScopeUtils.java
```diff
@@ -0,0 +1,104 @@
...
+     * <li>the scopes cannot be obtained (e.g. the target configuration is not supported),</li>
+     * <li>the current Gradle version doesn't provide scope information, or</li>
+     * <li>an exception is thrown</li>
+     * </ul>
+     * .
+     *
+     * @param configuration the target launch configuration
+     * @return the source set names
+     */
+    public static Set<String> collectScopes(ILaunchConfiguration configuration) {
```
oehme: ❌ What if the user manually added a source folder to the model, but forgot scope information and we also have other source folders with scope information? Currently this code assumes such a source folder needs no additional dependencies. I think instead we should treat this situation as "we don't know, let's put everything on the classpath".

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/launch/internal/GradleClasspathProviderUpdateTest.groovy
donat: Fixed

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/internal/GradleScopeUtils.java
donat: This makes sense. I've adjusted the code accordingly.

PR #608, opened 2017-11-02
Make Buildship work on Java 9

donat:
This change adds initial Java 9 compatibility to Buildship by
implementing the following changes:

 - Update Tooling API to 4.3
 - Update wrapper to 4.3
 - Update Gson to 2.7
 - Update Groovy to 2.4.12
 - Launch Eclipse with --add-modules=ALL-SYSTEM option on Java 9
 - Fix broken tests
 - Ignore tests that are using older Gradle versions
 - Use e47 build that supports Java 9
 - GradleClasspathProvider filters unresolved rt classpath entries

At the moment older Gradle versions cannot be used when Buildship
is executed on Java 9. With the new JDK versioning scheme, we have
lost backward compatibility in this platform. We will try to make
it work but there's no guarantee that it will be done.


org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/internal/GradleClasspathProvider.java
```diff
@@ -56,7 +57,32 @@ public GradleClasspathProvider() {
 
     @Override
     public IRuntimeClasspathEntry[] computeUnresolvedClasspath(ILaunchConfiguration configuration) throws CoreException {
-        return super.computeUnresolvedClasspath(configuration);
+        return filterUnusedDependencies(configuration, super.computeUnresolvedClasspath(configuration));
+    }
+
+    private IRuntimeClasspathEntry[] filterUnusedDependencies(ILaunchConfiguration configuration, IRuntimeClasspathEntry[] entriesToFilter) throws CoreException {
+        // if the run configuration uses Java 9 then the library dependencies are already present in the
```
oehme: What does this mean in practice, i.e. who uses the unresolved classpath?

donat:
Yes, I've set up a TC build job: https://builds.gradle.org/viewType.html?buildTypeId=Tooling_Master_IntegrationTests_Linux_Eclipse47Java9

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/internal/GradleClasspathProvider.java
donat: JDT itself. If the target platform is Java 9 then the unresolved classpath already contains the jar dependencies.

PR #616, opened 2017-11-29
Make synchronize methods execute synchronously

donat:
This is my first working attempt to convert the synchronization methods to work synchronously. 

After a few failed attempts to refactor a large chunk of the code I've decided to one step at the time. This change keeps the synchronization jobs internally and just joins to mimic the synchronous behavior. We can move forward with removing the jobs entirely as the next step.


donat:
I've created a `release-3.0` branch to keep the changes separately at least until we release Buildship 2.2.

donat:
This is still WIP.

oehme:
Okay, let me know when you would like a review.

donat:
@oehme You can have a look now. I'm sure there are a couple of rough edges, but the more I think about it the more the branch is getting out of hand. 

Basically, the main functionality is there. The synchronize methods are executed synchronously. Also, the project import/preview is running inside the wizard container without an additional job. 

Besides, the common bits of the `Job` implementations are now merged to a single `ToolingApiJob` base class.

donat:
@oehme As we discussed in our previous meeting, I've checked if we can keep the 2.0 compatibility on the `GradleBuild` class. Unfortunately, it's a bit tricky because I've already got rid of the `AsyncHandler` interface in the code. 

I think it doesn't worth the effort. If we keep the master branch clean and keep all the 3.0-related changes on a separate branch then we can easily publish a bugfix release if needed.

oehme:
I was mostly concerned about the effort of porting fixes from master to 3.0, but I'll leave that up to you, since you'll be the one who has to deal with those issues :)

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/launch/internal/GradleClasspathProviderUpdateTest.groovy
```diff
@@ -32,7 +32,7 @@ class GradleClasspathProviderUpdateTest extends ProjectSynchronizationSpecificat
         importAndWait(projectDir)
 
         then:
-        hasGradleClasspathProvider(launchConfiguration)
+        waitFor { hasGradleClasspathProvider(launchConfiguration) }
```
oehme: ⭕️ Why did all the tests change in this way? Shouldn't importAndWait take care of the waiting?

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/GradleBuild.java
```diff
@@ -41,42 +32,15 @@
...
-     */
-    void synchronize(NewProjectHandler newProjectHandler, AsyncHandler initializer);
-
-    /**
-     * Returns {@code true} if a synchronization job is already running for the same root project.
-     *
-     * @return true if a synchronization is running.
-     */
-    boolean isSyncRunning();
+    void synchronize(NewProjectHandler newProjectHandler, CancellationTokenSource tokenSource, IProgressMonitor monitor) throws Exception;
```
oehme: ⭕️ Shouldn't this only throw `CoreException` to be consistent with other Eclipse APIs?

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/GradleBuild.java
```diff
@@ -41,42 +32,15 @@
...
-     * The synchronization happens asynchronously. In case of a failure, the user will be notified
-     * once the build has finished.
-     * <p/>
-     * This is equivalent to calling {@code synchronize(newProjectHandler, AsyncHandler.NO_OP)}
+     * The synchronization happens synchronously. In case of a failure, the method throws a
+     * {@link CoreException} which contains the necessary status and error message about the
+     * failure.
      *
      * @param newProjectHandler how to handle newly added projects
+     * @param tokenSource the cancellation token source
```
oehme: ⭕️ Can we get rid of the token source argument and construct it ourselves?

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/ProjectNatureUpdater.java
```diff
@@ -48,7 +51,15 @@ public static void update(IProject project, Optional<List<OmniEclipseProjectNatu
 
         Result<String> result = ManagedModelMergingStrategy.calculate(existingNatures, modelNatures, managedNatures);
 
-        description.setNatureIds(result.getNextElements().toArray(new String[0]));
+        String[] natureIds = FluentIterable.from(result.getNextElements()).filter(new Predicate<String>() {
```
oehme: ⭕️ How is this related to the refactoring?

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/SynchronizingBuildScriptUpdateListener.java
```diff
@@ -125,4 +132,19 @@ public static SynchronizingBuildScriptUpdateListener createAndRegister() {
     public void close() {
         ResourcesPlugin.getWorkspace().removeResourceChangeListener(this);
     }
+
+    /**
+     * Custom result handler that only logs the failure.
+     */
+    private static final class ResultHander implements ToolingApiJobResultHandler<Void> {
```
oehme: ⭕️ Why do we only log here? Before it would show an error dialog.

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/view/task/ReloadTaskViewJob.java
```diff
@@ -67,11 +69,10 @@ private TaskViewContent loadContent(IProgressMonitor monitor) {
                      faultyProjects.remove(eclipseProject.getName());
                  }
                  projects.addAll(eclipseProjects);
-             } catch (RuntimeException e) {
-                 CorePlugin.logger().warn("Tasks can't be loaded for project located at " + gradleBuild.getBuildConfig().getRootProjectDirectory().getAbsolutePath(), e);
+             } catch (RuntimeException ignore) {
+                 // faulty projects will be represented as empty nodes
```
oehme: ⭕️ Why did this change?

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/SynchronizeGradleBuildsOperation.java
```diff
@@ -0,0 +1,85 @@
...
+    private SynchronizeGradleBuildsOperation(Set<GradleBuild> builds, NewProjectHandler newProjectHandler) {
+        this.builds = ImmutableSet.copyOf(builds);
+        this.newProjectHandler = Preconditions.checkNotNull(newProjectHandler);
+    }
+
+    Set<GradleBuild> getBuilds() {
+        return this.builds;
+    }
+
+    protected void run(CancellationTokenSource tokenSource, IProgressMonitor monitor) throws Exception {
```
oehme: ⭕️ Would it make sense to have some common interface like `ToolingApiRunnable` for this? Then we could have `CorePlugin.run(tapiRunnable)`, which would take care of getting a token source for you.

donat:
Yes, I did every change by manually testing if the functionality is there.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/ProjectNatureUpdater.java
donat: This was a temporary fix for the broken `ImportingWtpProjectsTest`. The problem was already there as we set the unrecognized  WTP nature on the test project, but when we used the `Job` implementation the problem was swallowed. You can still find it in the logs: 

https://builds.gradle.org/viewLog.html?buildId=9843560&buildTypeId=Tooling_Master_IntegrationTests_Linux_Eclipse47IntegrationTests&tab=buildLog&_focus=2465#_state=61,1374

I've restored the original code and moved the fix to the `WorkspaceOperationsDelegate` test fixture.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/GradleBuild.java
donat: It makes sense, I've changed it.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/SynchronizingBuildScriptUpdateListener.java
donat: This was intentional. If the `auto-sync` feature is enabled then a new error dialog would pop up every time the user saves a faulty build script file.

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/view/task/ReloadTaskViewJob.java
donat: Reverted.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/SynchronizeGradleBuildsOperation.java
donat: I've introduced `ToolingApiOperation` both in the synchronous invocations and in the jobs.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/GradleBuild.java
donat: Well, you need the tokenSource to be available on the caller side to attach the cancellation for instance to the preview page.

PR #618, opened 2017-12-04
Remove Maven nature when Gradle project imported

donat:
An Eclipse project should be either Gradle or Maven project but
not both at the same time. If both natures were present, the plugins
would compete on how to calculate the classpath for the associated
run configurations. To resolve this situation, Buildship now
actively removes the Maven nature during the Gradle synchronization.

### DIFF
--- a/org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/workspace/internal/ProjectNatureUpdaterTest.groovy
+++ b/org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/workspace/internal/ProjectNatureUpdaterTest.groovy
@@ -1,5 +1,7 @@
 package org.eclipse.buildship.core.workspace.internal
 
+import spock.lang.Issue
+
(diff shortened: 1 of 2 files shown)


donat:
@oehme I know it's last minute but could you please have a look. The 2.2 release is due tomorrow and I'd like to include this change in it.

oehme:
What if I manually add the m2e nature or run the Maven importer? As far as I can tell from the code this will only catch the case when the user has the Maven nature in their build script. At least that's the only test we have.

donat:
Yeah, that way the stack overflow problem would still occur until a Gradle project synchronization would remove the maven nature. Which is also not guaranteed because the user could include the maven nature in `eclipse.project.natures`.

donat:
Here is my proposal. The classpath provider should only update a run configuration if it was not called recently with the same arguments.

oehme:
You could have something like a reentrant check, i.e. remember that the updater is currently being called and stop if it's called again recursively.

PR #619, opened 2017-12-07
Prevent stack overflow when Maven project is imported with Gradle wizard

donat:


org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/internal/DefaultExternalLaunchConfigurationManager.java
```diff
@@ -170,14 +170,26 @@ private void updateLaunchConfiguration(ILaunchConfigurationWorkingCopy configura
      */
     private class LaunchConfigurationListener implements ILaunchConfigurationListener, EventListener {
 
+        // If another ILaunchConfigurationListener changes the target launch configuration then
+        // a new change event is generated. That event is handled synchronously invoking this
+        // listener again and causing a stack overflow error.
+        // see https://github.com/eclipse/buildship/issues/617
+        private boolean configChangeCalled;
```
oehme: Is this called from different threads? If yes it needs to be thread safe.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/internal/DefaultExternalLaunchConfigurationManager.java
```diff
@@ -170,14 +170,26 @@ private void updateLaunchConfiguration(ILaunchConfigurationWorkingCopy configura
...
 
         @Override
         public void launchConfigurationChanged(ILaunchConfiguration configuration) {
-            updateClasspathProvider(configuration);
+            if (this.configChangeCalled) {
+                return;
+            } else {
+                this.configChangeCalled = true;
+                updateClasspathProvider(configuration);
+                this.configChangeCalled = false;
```
oehme: Needs to be in a finally block. Otherwise any error will put us into noop mode forever.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/internal/DefaultExternalLaunchConfigurationManager.java
donat: My initial assumption was that it doesn't need to be thread safe because the problem happened on a single thread. But, I've just discovered that that listener can be called from a different thread so I introduced a `ReentrantLock` to guard the execution.

TLDR; fixed.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/internal/DefaultExternalLaunchConfigurationManager.java
donat: Fixed.

donat:
@oehme I've addressed your review items. Would you mind having a second look?

donat:
I'm going to merge this PR since it's blocking the 2.2 release. The impact is minimal and all pending review items are already fixed.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/launch/internal/DefaultExternalLaunchConfigurationManager.java
```diff
@@ -170,14 +170,29 @@ private void updateLaunchConfiguration(ILaunchConfigurationWorkingCopy configura
...
+
         @Override
         public void launchConfigurationAdded(ILaunchConfiguration configuration) {
             updateClasspathProvider(configuration);
         }
 
         @Override
-        public void launchConfigurationChanged(ILaunchConfiguration configuration) {
-            updateClasspathProvider(configuration);
+        public synchronized void launchConfigurationChanged(ILaunchConfiguration configuration) {
```
oehme: We don't want to prevent multiple threads from updating at the same time, we just want to prevent them from overriding each other's state - A threadlocal would be the better choice here I think.

donat: I agree. I've fixed it on `master`: https://github.com/eclipse/buildship/commit/1668613eb93e0c9590e38bdd9fdb8249c9431d0f

PR #625, opened 2017-12-19
Make debugging be able to look up sources from dependency project

donat:
The source lookup takes the values from the resolved runtime classpath. If the dependency project is missing from there, then the source lookup from that project will fail when debugging.


org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/GradleClasspathContainerRuntimeClasspathEntryResolver.java
```diff
@@ -84,6 +84,9 @@ private void collectContainerRuntimeClasspath(IClasspathContainer container, Lis
                     if (candidate.isPresent()) {
                         IJavaProject dependencyProject = JavaCore.create(candidate.get());
                         IRuntimeClasspathEntry projectRuntimeEntry = JavaRuntime.newProjectRuntimeClasspathEntry(dependencyProject);
+                        // add the project entry itself so that the source lookup can find the classes
+                        // see https://github.com/eclipse/buildship/issues/383
+                        result.add(projectRuntimeEntry);
```
oehme: I remember vaguely that this somehow added all the reexported dependencies as well, which is why we didn't do this before. Or was that some other place in the code?

donat: The current method is anyway called recursively adding all reexported dependencies from the dependent project.

You might think about [Exclusion of transitive dependencies is ignored](https://bugs.eclipse.org/bugs/show_bug.cgi?id=473348) where we adjusted the behavior.

PR #634, opened 2018-01-05
Use error markers to display synchronization problems

donat:


org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/DefaultGradleBuild.java
```diff
@@ -46,7 +49,17 @@ public DefaultGradleBuild(BuildConfiguration buildConfig) {
...
+        try {
+            syncOperation.run(tokenSource, monitor);
+            GradleMarkerManager.clear(this);
+        } catch (Exception e) {
+            GradleMarkerManager.clear(this);
+            ToolingApiStatus status = ToolingApiStatus.from("Project synchronization" , e);
+            if (status.severityMatches(IStatus.WARNING | IStatus.ERROR)) {
+                GradleMarkerManager.addError(this, status);
+            }
+            throw e;
```
oehme: ⭕️ I don't think synchronization should throw an exception if the problem could be handled by adding an error marker. There might be some conditions where we really can't continue (like we can't even import the root project), but in most cases the error is not really in the synchronization process itself, but in the users's build.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/DefaultGradleBuild.java
```diff
@@ -46,7 +49,17 @@ public DefaultGradleBuild(BuildConfiguration buildConfig) {
     @Override
     public void synchronize(NewProjectHandler newProjectHandler, CancellationTokenSource tokenSource, IProgressMonitor monitor) throws CoreException {
         SynchronizeGradleBuildsOperation syncOperation = SynchronizeGradleBuildsOperation.forSingleGradleBuild(this, newProjectHandler);
-        syncOperation.run(tokenSource, monitor);
+        try {
+            syncOperation.run(tokenSource, monitor);
+            GradleMarkerManager.clear(this);
```
oehme: ❌ We should clear markers before, not after. We might for instance want to add warning markers which shouldn't be cleaned up.

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/marker/MarkersPropertyPage.java
```diff
@@ -0,0 +1,86 @@
...
+        stackTraceLabel.setText("Stacktrace:");
+        GridDataFactory.swtDefaults().align(SWT.LEFT, SWT.TOP).applyTo(stackTraceLabel);
+
+        Text stacktraceAreaText = new Text(parent, SWT.MULTI | SWT.READ_ONLY | SWT.BORDER | SWT.H_SCROLL | SWT.V_SCROLL);
+        GridDataFactory.fillDefaults().hint(100, 100).grab(true, true).applyTo(stacktraceAreaText);
+
+        stacktraceAreaText.setText(stacktrace);
+    }
+
+    public static int countLines(String str) {
```
oehme: There is probably something like CharMatcher.count

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/DefaultGradleBuilds.java
```diff
@@ -42,12 +41,8 @@ public DefaultGradleBuilds(Set<BuildConfiguration> buildConfigs) {
 
     @Override
     public void synchronize(NewProjectHandler newProjectHandler, CancellationTokenSource tokenSource, IProgressMonitor monitor) throws CoreException {
-        SynchronizeGradleBuildsOperation syncOperation = SynchronizeGradleBuildsOperation.forMultipleGradleBuilds(this, newProjectHandler);
```
oehme: ⭕️ This class seems unused now.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/marker/internal/ErrorMarkerLocation.java
```diff
@@ -0,0 +1,136 @@
...
+                if (persistentModel.isPresent()) {
+                    IFile buildScript = project.getFile(persistentModel.getbuildScriptPath());
+                    if (buildScript.exists()) {
+                        return buildScript;
+                    }
+                }
+            }
+            return project;
+        } else {
+            return ResourcesPlugin.getWorkspace().getRoot();
```
oehme: ⭕️ Do we ever clean these up?

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/DefaultGradleBuild.java
oehme: ⭕️ Do we clear markers when unlinking a project?

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/marker/GradleErrorMarker.java
```diff
@@ -0,0 +1,94 @@
...
+        IMarker marker = resource.createMarker(GradleErrorMarker.ID);
+
+        if (lineNumber >= 0) {
+            marker.setAttribute(IMarker.LINE_NUMBER, lineNumber);
+        }
+
+        marker.setAttribute(IMarker.MESSAGE, collectErrorMessages(exception));
+        marker.setAttribute(IMarker.PRIORITY, IMarker.PRIORITY_HIGH);
+        marker.setAttribute(IMarker.SEVERITY, IMarker.SEVERITY_ERROR);
+        marker.setAttribute(IMarker.LOCATION, "Gradle build script");
```
oehme: ⭕️ It's not necessarily a build script - only if the resource is a file.

donat: Any suggestion what should we display here? Maybe nothing?

screen shot 2018-01-08 at 13 45 40

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/marker/internal/ErrorMarkerLocation.java
donat: The cleanup starts at the workspace root: https://github.com/eclipse/buildship/pull/634/files#diff-1158f74e59739104ec0d6c6687e86ea2R38

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/DefaultGradleBuild.java
donat: I was thinking about the same thing, although I ended up calling `clear()` after the synchronization because if the synchronization takes longer then the user won't see the errors for a while. It's just a thought though, I have no problem moving the `clear()` before the synchronization.

donat: I think we do, but there's no coverage for that. Let me check.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/DefaultGradleBuild.java
donat: I added the rethrow so that the error dialog is still shown to the user. I planned to follow this order:
1. introduce error markers
2. allow importing root projects of broken builds
3. remove error dialog
The rethrow would be removed in step #3

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/DefaultGradleBuilds.java
donat: I removed the method and simplified the class accordingly.

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/marker/MarkersPropertyPage.java
donat: I've found this method only for chars but not for [strings](https://github.com/google/guava/issues/877). We could use the `Splitter` class, but that would instantiate a bunch of unused `String` objects, so I'd prefer the existing solution.

donat: Fixed

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/marker/GradleErrorMarker.java
donat: I've removed the location field.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/DefaultGradleBuild.java
donat: I've moved `clear()` before the project synchronization.

donat: I've added coverage for this usecase.

donat:
@oehme I've addressed the review items, you can have a look.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/marker/GradleErrorMarker.java
oehme: Yeah, it doesn't really add value.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/DefaultGradleBuild.java
oehme: Sounds good to me.

PR #637, opened 2018-01-10
Import root project even if project synchronization fails 

donat:
This PR ensures that the project import first imports the root project as a generic Eclipse project and then proceeds with the TAPI model loading and the project synchronization.

This change clears the way from replacing the error notification dialogs with error markers. When the import fails we can just put the error marker on the root project.

The import wizard has been adjusted accordingly: when the user clicks the `Finish` button the wizard closes even when the synchronization fails. The only exception is when the root project itself cannot be imported.


org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/ImportRootProjectOperation.java
```diff
@@ -0,0 +1,122 @@
...
+
+        if (!projectOrNull.isPresent() && this.newProjectHandler.shouldImport()) {
+            String projectName = findFreeProjectName(rootDir.getName());
+            projectOrNull = Optional.of(workspaceOperations.createProject(projectName, rootDir, ImmutableList.<String>of(), progress.newChild(1)));
+        }
+
+        if (projectOrNull.isPresent()) {
+            IProject project = projectOrNull.get();
+            project.refreshLocal(IResource.DEPTH_INFINITE, progress.newChild(1));
+            CorePlugin.workspaceOperations().addNature(project, GradleProjectNature.ID, progress.newChild(1));
```
oehme: ❌ Why are we refreshing and adding the Gradle nature here? I think this should be done by the next step (synchronization).

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/ImportRootProjectOperation.java
```diff
@@ -0,0 +1,122 @@
...
+
+    private void verifyNoWorkspaceRootIsImported(File rootDir, IProgressMonitor monitor) {
+        File workspaceRoot = ResourcesPlugin.getWorkspace().getRoot().getLocation().toFile();
+        if (rootDir.equals(workspaceRoot)) {
+            throw new UnsupportedConfigurationException(String.format("Project %s location matches workspace root %s", rootDir.getName(), workspaceRoot.getAbsolutePath()));
+        }
+    }
+
+    private void saveProjectConfiguration(BuildConfiguration buildConfiguration, File rootDir, IProgressMonitor monitor) {
+        ProjectConfiguration projectConfiguration = CorePlugin.configurationManager().createProjectConfiguration(buildConfiguration, rootDir);
```
oehme: ❌ Why do we need to duplicate this here? Shouldn't this be done by the synchronization action that comes next?

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/SynchronizeGradleBuildOperation.java
```diff
@@ -244,6 +244,7 @@ private void synchronizeJavaProjectInTransaction(final OmniEclipseProject projec
         JavaSourceSettingsUpdater.update(javaProject, project, progress.newChild(1));
         GradleClasspathContainerUpdater.updateFromModel(javaProject, project, SynchronizeGradleBuildOperation.this.allProjects, persistentModel, progress.newChild(1));
         WtpClasspathUpdater.update(javaProject, project, progress.newChild(1));
+        CorePlugin.externalLaunchConfigurationManager().updateClasspathProviders(workspaceProject);
```
oehme: ⭕️ This seems unrelated to this change. Why is it here?

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/NewProjectHandler.java
```diff
@@ -26,12 +24,12 @@
     NewProjectHandler NO_OP = new NewProjectHandler() {
 
         @Override
-        public boolean shouldImport(OmniEclipseProject projectModel) {
+        public boolean shouldImport() {
```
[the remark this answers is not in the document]
donat: Makes sense. How about `shouldImportNewProjects()` and `afterProjectImported(IProject)`?

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/ImportRootProjectOperation.java
donat: Same as above.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/SynchronizeGradleBuildOperation.java
donat: This is actually related to this change and covers a missing use-case. The Gradle classpath provider is updated upon certain events (Gradle nature added, project created, etc.). The update only considers Java projects. Now, we create a generic root project and add the Java nature later, there's a chance that some run configurations are not updated properly. 

Calling `updateClasspathProviders()` at the end of the synchronization adds minimal overhead and ensures that no matter how we adjust the project configuration, the classpath providers will be up-to-date after a project synchronization.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/NewProjectHandler.java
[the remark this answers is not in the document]
donat: Fixed

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/ImportRootProjectOperation.java
oehme: The user could use the "Add Gradle Nature" action instead, but I get what you mean.

PR #642, opened 2018-01-16
Ensure workspace project is in sync with filesystem before updating preferences

donat:
Currently, we save the project preferences before the project files are synchronized with the file system. In some cases (when importing existing projects) it's even worse: a the file syncing starts in the background, then comes the preference save. As a result we see frequent, but random `BackingStoreException`s being raised.

This pull request fixes this problem as it makes all file sync happening synchronously and before the preferences are updated.

### DIFF
--- a/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/SynchronizeGradleBuildOperation.java
+++ b/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/SynchronizeGradleBuildOperation.java
@@ -189,17 +189,21 @@ final class SynchronizeGradleBuildOperation implements IWorkspaceRunnable {
 
     private void synchronizeWorkspaceProject(OmniEclipseProject project, IProject workspaceProject, SubMonitor progress) throws CoreException {
         if (workspaceProject.isAccessible()) {
-            synchronizeOpenWorkspaceProject(project, workspaceProject, progress);
+            synchronizeOpenWorkspaceProject(project, workspaceProject, true, progress);
         } else {
             synchronizeClosedWorkspaceProject(progress);
         }
     }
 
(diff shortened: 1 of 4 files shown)


donat:
@oehme Could you please have a look? This is the missing story for Buildship 2.2.1.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/SynchronizeGradleBuildOperation.java
```diff
@@ -189,17 +183,27 @@ private void synchronizeGradleProjectWithWorkspaceProject(OmniEclipseProject pro
...
         }
     }
 
-    private void synchronizeOpenWorkspaceProject(OmniEclipseProject project, IProject workspaceProject, SubMonitor progress) throws CoreException {
+    private void synchronizeOpenWorkspaceProject(OmniEclipseProject project, IProject workspaceProject, boolean refreshNeeded, SubMonitor progress) throws CoreException {
         progress.setWorkRemaining(10);
 
         //currently lots of our synchronization logic assumes that the whole resource tree is readable.
-        CorePlugin.workspaceOperations().refreshProject(workspaceProject, progress.newChild(1));
+        if (refreshNeeded) {
```
oehme: ⭕️ Why not always do a refresh here and leave the workspace operations unchanged? Does the synchronous call not replace the background call?

donat: I don't think it does but I'll check just to be sure.

donat: I've checked again, the synchronous call doesn't replace the background call. Also, the background job starts with a 200-millisecond delay, so it's indeterministic if the background refresh finishes before or after Gradle project synchronization.

PR #644, opened 2018-01-22
Remove custom error dialog

donat:
Now we present the synchronization problems with error markers, so we can get rid of the annoying error dialogs from the UI. 

For the project preview errors, I've modified a UI such that the error and the stack trace are displayed within the wizard dialog. 

It seems that there is a problem in the SWT/JFace wizard implementation. When the preview progress starts in `IRunnableContext.run()`, the widgets are disabled and then re-enabled after completion. Unfortunately, the `Text` widget displaying the stack trace is not properly enabled and has a greyish color instead of the default black.


donat:
@oehme Rethrowing the exception in `DefaultGradleBuild` is necessary as it flags whether the synchronization fails or succeeds. Consequently, without it, the `ProjectImportWizard` wouldn't be able to tell if it should be closed at the end of the import: https://github.com/eclipse/buildship/blob/release-3.0/org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/wizard/project/ProjectImportWizardController.java#L197-L199

donat:
I see. It would be much better Buildship would do the project sync-exception handling cycle project-by-project. Looking forward to the discussion.

PR #646, opened 2018-01-22
In Tasks view don't try to show new input when the widgets are disposed

donat:


oehme:
The fact that we get new input after being disposed is worth debugging. We might be starting sync jobs when the workspace is already shutting down or something like that.

PR #649, opened 2018-02-09
Remove Omni model usages from the codebase

donat:
This PR removes the "Omni" model classes that remained in the codebase.

There are two significant changes. In the core plugin, I introduced a decorator `CompatEclipseProject` for the corresponding TAPI model, that adds sensible defaults for the older Gradle versions. This decorator is injected during the `ModelProvider` queries. 

There are times where Buildship needs to distinguish if the TAPI model supports a specific attribute. For those, the compatibility model forces to use helper methods defined in `ModelUtils`.

The UI plugin now contains a simplified version of `OmniBuildInvocations`.

I haven't migrated the tests, but I'm planning to do that in a next step. Keeping the cross-version coverage for the models would be nice.


oehme:
> There are times where Buildship needs to distinguish if the TAPI model supports a specific attribute. For those, the compatibility model forces to use helper methods defined in ModelUtils.

Depending on which model elements you are referring to, you could also use a null-object pattern. I.e. have a constant empty list called `UNSUPPORTED_BUILD_COMMANDS` and use that in your checks instead of forcing every client to use a different method. Or you could have an `OptionalList` class in your subclasses, which has an `isPresent` method. That would avoid having to look at a different utility class. 

We should definitely ensure that all functionality in Buildship has proper cross-version coverage. I think we already have this for many things, but please double-check.

Also you might be able to get rid of some of the catch-blocks, because we don't support some older versions anymore anyway.

donat:
I designed this PR so that the TAPI model clients don't see the `Compat*` classes at all (they are non-pubic classes), so I'd rather avoid introducing new methods there. The null-object pattern sounds better; I'll adjust the code.

Regarding the compatibility, we haven't explicitly declared what Gradle versions do we support. I think I'll be able to pinpoint a reasonable lower limit when I verify the cross-version coverage.

oehme:
We do have a hard lower limit - The lower limit of the TAPI.

donat:
Yes, right. The lower limit for the TAPI is Gradle 1.2 ATM.

oehme:
It'll be 2.9 once we upgrade to 5.0 in the middle of this year. I think we could already make that the lower limit for Buildship 3.0 now

donat:
I've modified the PR to use the null-object pattern: https://github.com/eclipse/buildship/compare/5210ce212b23...e943124dfe0d
@oehme can you please have a look before I merge this PR?

donat:
FYI, I'll verify the coverage as part of #650.

PR #657, opened 2018-02-19
Merge tooling-commons and Buildship utility classes

donat:
This PR contains two separate items:
- `GradleDistribution` and `GradleDistributionWrapper` are merged.
- Removed `FixedRequestAttributes`.

I've made a few adjustments regarding `GradleDistribution`.  I've extracted the type's attributes to an external `GradleDistributionInfo` class that can represent both valid and invalid distributions. These objects are the subject of validation (used by the preferences UI for instance). The info objects can be transformed into `GradleDistribution` instance, that are ensured to represent valid distributions.

I've changed the preference loading a little bit so that we fall back to the wrapper distribution in case of an invalid distribution. We already do this when the distribution string is missing from the preferences. 

Removing `FixedRequestAttributes` is a simple change, we only kept it for compatibility reasons.


donat:
I've successfully replaced `TransientRequestAttributes` with a new `GradleProgressAttributes` class that creates listeners and console streams in a uniform and more readable way.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/gradle/GradleProgressAttributes.java
```diff
@@ -0,0 +1,178 @@
...
+import org.eclipse.buildship.core.console.ProcessStreamsProvider;
+import org.eclipse.buildship.core.util.progress.CancellationForwardingListener;
+import org.eclipse.buildship.core.util.progress.DelegatingProgressListener;
+
+/**
+ * Holds attributes that are commonly used to handle progress in each Gradle invocation.
+ *
+ * @author Donat Csikos
+ */
+public final class GradleProgressAttributes {
```
oehme: Just a thought here: Do we even need this parameter class or could we extend the `ConfigurableLauncher` interface from the TAPI instead by adding some Buildship-specific convenience methods?

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/gradle/GradleProgressAttributes.java
```diff
@@ -0,0 +1,178 @@
...
+    public List<org.gradle.tooling.events.ProgressListener> getProgressEventListeners() {
+        return this.progressEventListeners;
+    }
+
+    /**
+     * Returns the cancellation token associated to this instance.
+     *
+     * @return the cancellation token
+     */
+    public CancellationToken getCancellationToken() {
```
oehme: ⭕️ Do we really need getters here? Shouldn't there be a single method to apply the settings to a `ConfigurableLauncher`?

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/gradle/GradleProgressAttributes.java
donat: I'm not sure I understand the proposal. Do you mean we should reimplement the `ConfigurableLauncher` interface so that can apply the attributes on the build/test launchers created by `GradleBuild`? In a way, `GradleProgressAttributes` already does this.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/gradle/GradleProgressAttributes.java
donat: That totally makes sense. I've just added the change.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/gradle/GradleProgressAttributes.java
oehme: I was thinking that the methods on `GradleBuild` could just return an `ExtendedConfigurableLauncher extends ConfigurableLauncher` (forgive the dumb name). Then users can modify everything they can do on the TAPI, but also get some convenience methods on top.

PR #662, opened 2018-02-23
Migrate test coverage from tooling-commons

donat:


org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/view/task/BuildInvocationsTest.groovy
```diff
@@ -0,0 +1,187 @@
...
+import org.eclipse.core.resources.IProject
+import org.eclipse.core.runtime.NullProgressMonitor
+
+import org.eclipse.buildship.core.CorePlugin
+import org.eclipse.buildship.core.util.gradle.Path
+import org.eclipse.buildship.core.workspace.FetchStrategy
+import org.eclipse.buildship.core.workspace.ModelProvider
+import org.eclipse.buildship.ui.test.fixtures.ProjectSynchronizationSpecification
+
+class BuildInvocationsTest extends ProjectSynchronizationSpecification {
```
oehme: ❌ This test is basically just testing the TAPI, which is unnecessary, as it already has tests in Gradle core.

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/workspace/internal/SingleProjectSynchronizationSpecification.groovy
```diff
@@ -270,16 +327,28 @@ abstract class SingleProjectSynchronizationSpecification extends ProjectSynchron
         }
 
         when:
-        synchronizeAndWait(projectDir)
+        importAndWait(projectDir, distribution)
+        String outputLocation = findJavaProject('sample-project').outputLocation?.toPortableString()
 
         then:
-        findJavaProject('sample-project').getOutputLocation().toPortableString() == '/sample-project/target/bin'
+        if (higherOrEqual('3.0', distribution)) {
```
oehme: All these if-statements make the tests pretty hard to follow. You might want to take the approach from Gradle core instead where we have a `@TargetGradleVersion` annotation on each test method, so we can split the different cases into different methods.

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/util/binding/PropertyTest.groovy
```diff
@@ -0,0 +1,138 @@
+package org.eclipse.buildship.core.util.binding
+
+import spock.lang.Specification
+
+import com.google.common.base.Optional
+
+class PropertyTest extends Specification {
```
oehme: ⭕️ Where do we still use Property?

donat: In [`ProjectImportConfiguration`](https://github.com/eclipse/buildship/blob/1ffb28156481e2e83cf9be768ad5e008c76fe81c/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/projectimport/ProjectImportConfiguration.java).

org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/view/task/BuildInvocationsTest.groovy
donat: Deleted.

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/workspace/internal/SingleProjectSynchronizationSpecification.groovy
donat: I'll try to copy the test fixtures.

donat: I tried to reuse `@TargetGradleVersion`. Copying over the custom test runner adds a lot of extra code which seems to be too much of an extra maintenance burden. I can still split the tests based on the version requirement.

org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/view/task/BuildInvocationsTest.groovy
oehme: Wait a second, is BuildInvocations part of Builfship? In that case please disregard my comment.

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/workspace/internal/SingleProjectSynchronizationSpecification.groovy
donat: I've refactored the tests not to have those `higherOrEqual` conditions in the test assertions.

PR #663, opened 2018-02-27
Consume Tooling API dependency directly from the build

donat:


org.gradle.toolingapi/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,78 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: Gradle Tooling API
+Bundle-Vendor: Gradle Inc.
+Bundle-SymbolicName: org.gradle.toolingapi
+Bundle-ClassPath: .
+Bundle-RequiredExecutionEnvironment: JavaSE-1.7
+Require-Bundle: org.slf4j.api;bundle-version="1.7.2"
+Export-Package: org.gradle;version="4.3.0",
```
oehme: ❌ Almost none of the packages below are part of the tooling API. We should only export the ones that are actually part of the official API.

org.gradle.toolingapi/.project
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
```
oehme: ❌ I'm a bit confused by the whole setup. Why is the TAPI jar checked in? Shouldn't that be the job of the build to set up the project?

donat: The jar is only needed to make the Eclipse workspace compile successfully. Without it, we'd need to execute Gradle tasks when setting up the dev environment.

org.gradle.toolingapi/META-INF/MANIFEST.MF
donat: I didn't change anything in that regard. I remember we had problems before when not all packages were exported. IMO we should tackle this separately.

oehme: Can you open an issue for that for 3.0? We definitely need to fix this, as these packages should not be exposed to anyone.

donat: https://github.com/eclipse/buildship/issues/671

PR #672, opened 2018-03-09
Hide internal packages in TAPI bundle

donat:


org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/util/gradle/GradleDistributionTest.groovy
```diff
@@ -73,7 +72,7 @@ class GradleDistributionTest extends Specification {
         1 * connector.useBuildDistribution()
     }
 
-    def "toString"() {
+    def "_toString"() {
```
oehme: Why not give these methods a more descriptive name?

org.gradle.toolingapi/META-INF/MANIFEST.MF
```diff
@@ -9,8 +9,6 @@ Require-Bundle: org.slf4j.api;bundle-version="1.7.2"
 Export-Package: org.gradle;version="4.3.0",
```
oehme: Is org.gradle really part of the tooling API? It might just be here because it is used by the TAPI internally. I don't remember any classes from that package that you would need to use ProjectConnection. We should hide as much as possible.

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/util/gradle/GradleDistributionTest.groovy
donat: This class was just moved from tooling-commons. The change you see here is because the groovy tooling showed some errors.

oehme: I know and my comment was that instead of making the test names even more artificial, we should make them more human friendly.

org.gradle.toolingapi/META-INF/MANIFEST.MF
donat: I've already found one exception: `import org.gradle.api.JavaVersion` is used in the tooling models.

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/util/gradle/GradleDistributionTest.groovy
donat: Done

org.gradle.toolingapi/META-INF/MANIFEST.MF
donat: I managed to remove a bunch of unused entries from the exported packages list. Only the following ones remained:
- `org.gradle.tooling.*`: obviously
- `org.gradle.tooling`: because of `JavaVersion`
- `org.gradle.util`: because of `GradleVersion`

donat:
@oehme I've removed the `org.gradle.util` and I copied over the missing `GradleVersion` (with tests) and the `DistributionLocator` classes.

donat:
I've removed the unused bits in `GradleVersion`. The rest looks fine to me.

PR #675, opened 2018-03-13
Hide Gradle test launch shortcut on production sources

donat:


org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/launch/TestLaunchShortcutValidator.java
```diff
@@ -105,6 +131,37 @@ private static boolean validateJavaElements(List<? extends IJavaElement> element
...
+        IClasspathEntry fragmentRootEntry = null;
+        try {
+            fragmentRootEntry = getPackageFragmentRoot(type).getRawClasspathEntry();
+        } catch (JavaModelException e) {
+            throw new GradlePluginsRuntimeException(e); // raw classpath entry checked in `isInSourceFolder()`
+        }
+
+        Optional<Set<String>> scopes = ClasspathUtils.scopesFor(fragmentRootEntry);
+        if (scopes.isPresent()) {
+            // if source folder provides source set information then check if the source set name contains the word `test`
```
oehme: Obvious comment is obvious ;)

Please don't use inline comments. Make your code readable without them. If there is a class or method that needs deeper explanation, add a class or method-level block comment.

donat: Makes sense. I removed the comments.

Issue #354, opened 2017-02-08
Remove test dependencies from runtime classpath


donat:
## Context

The Buildship runtime classpath provided for Eclipse is different from the one used in Gradle. This leads to problematic classpath differences when the user executes a Java application from the IDE and via a Gradle build. It makes Spring Boot application development with STS and Buildship painful.

Original post: https://bugs.eclipse.org/bugs/show_bug.cgi?id=482315

## Overview

Buildship should read the runtime classpath from the Gradle build and use it in Buildship.

In Gradle, all source folders and classpath entries should be marked with classpath attributes describing which source sets they belong.

In Eclipse the default JDT runtime classpath calculation should be overridden removing all elements from the classpath that don't belong to the same source set where the launched class is declared.

## Implementation in Gradle

The `eclipse` plugin should define the following classpath attributes:
- source folders: `<attribute name="gradle_source_sets" value="main,test"/>`
- external dependencies: `<attribute name="gradle_source_sets" value="main,test"/>`

The source folder attribute is easy because the `SourceFoldersCreator` instantiates source folders directly from `SourceSet`s.

To calculate the attribute for the external dependencies we have to collect the runtime classpath entries for all source sets. Then, `EclipseDependenciesCreator` should look up which source sets an external dependency belong and add this info as a classpath attribute.

### Step 1 - Adjust `GradleClasspathContainerRuntimeClasspathEntryResolver`

The classpath container should automatically pick up the new classpath attributes from Gradle.

The classpath entry resolver should only add an external dependency to the runtime classpath if the target source folder from the run configuration and dependency have a common source folder. If the source folder can't be determined (e.g. no run configuration is provided) then no such filtering is needed. 

### Step 2 - Provide custom JDT classpath provider
Along with external dependencies, Buildship should filter source folders that don't belong to the source set from where a JDT execution is launched. It can be done by contributing a custom runtime classpath provider via the `org.eclipse.jdt.launching.classpathProviders` extension point. Here's the required interface to implement:

    public interface IRuntimeClasspathProvider {
        IRuntimeClasspathEntry[] computeUnresolvedClasspath(ILaunchConfiguration);
        IRuntimeClasspathEntry[] resolveClasspath(IRuntimeClasspathEntry[], ILaunchConfiguration);
    }

We should implement this extension point such that an entry is not added to the runtime classpath if it's not part of the current source set. The existing `GradleClasspathContainerRuntimeClasspathEntryResolver` class should be merged with this code.

[...]

PR #692, opened 2018-04-15
Lukeu/fix/more tests off runtime classpath

lukeu:
This change resolves the 'test resources in the runtime classpath' issue I was continuing to see following issue #354. I tested with my minimal example with Java 8 & 9, and my commercial workspace with Java 8 & both work as expected with this. However I really don't feel 'confident' with Eclipse or Buildship code yet, so I think this does need a good check - there seem to be quite a number of possibilities to consider.

(I spent some time trying to write an integration test for this based on RuntimeClasspathTest.groovy, but failed. I got as far as seeing that I'd need to construct a Jdt launch (rather than a Gradle launch) but that seemed to entail too much to set up, without some form of mocking or stubbing. There didn't seem to be something like that to emulate either?)

### DIFF
--- a/org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/workspace/internal/RuntimeClasspathTest.groovy
+++ b/org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/workspace/internal/RuntimeClasspathTest.groovy
@@ -167,6 +167,8 @@ class RuntimeClasspathTest extends ProjectSynchronizationSpecification {
     @Issue("https://bugs.eclipse.org/bugs/show_bug.cgi?id=507206")
     def "Runtime classpath contains custom output folders"() {
         setup:
+        // Another non-custom source directory is required for the default-directory to be set
+        new File(location, 'a/src/test/java').mkdirs()
         buildFile << '''
             project(':a') {
                 apply plugin: 'eclipse'
--- a/org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/workspace/internal/RuntimeClasspathTest.groovy
+++ b/org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/workspace/internal/RuntimeClasspathTest.groovy
@@ -25,9 +25,9 @@ class RuntimeClasspathTest extends ProjectSynchronizationSpecification {
     def setup() {
         location = dir('sample-project') {
             file('settings.gradle') << "include 'a', 'b', 'c'"
-            dir('a/src/main/java').mkdirs()
-            dir('b/src/main/java').mkdirs()
-            dir('c/src/main/java').mkdirs()
+            dir('a/src/main/java')
+            dir('b/src/main/java')
+            dir('c/src/main/java')
             buildFile = file 'build.gradle', '''
                 subprojects {
                     apply plugin: 'java'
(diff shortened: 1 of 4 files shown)


donat:
Hi! Thanks for investing time into creating a pull request.

I had to switch to a different project lately but I'm trying to get back to may Buildship duties. Expect replies to your previous questions in the upcoming days. 

Regarding the tests, it's indeed a tricky thing to implement. We do have [coverage to execute launch configurations and inspect the runtime classpath](https://github.com/eclipse/buildship/blob/6b651347ba2e529d516a4481825f7222c9d52768/org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/launch/ClasspathSeparationTest.groovy). Unfortunately it uses some Eclipse UI components, hence it is located in the `org.eclipse.ui.plugin`.

lukeu:
> I had to switch to a different project lately (...) expect replies to your previous questions in the upcoming days.

Thanks & sure, whenever makes sense - we humans are bad at multi-tasking.  :-)

lukeu:
Hi, I'm back from travels and came back to look at this.

> We do have coverage to execute launch configurations and inspect the runtime classpath.

Thanks for the pointer. Unfortunately I've hit some blockers and can't get the existing `ClasspathSeparationTest` to run: one test timed-out, and the rest failed because of a snapshot that is no longer [available here](https://services.gradle.org/distributions-snapshots/). After updating all the links to an available gradle-snapshot version, then all tests throw a timeout exception. Increasing the timeout 5s -> 60s did not help, something seems to be genuinely not working.

Would a way to move this forward be to just do a code review? The fix is essentially just a 2 line change that might just need a knowledgeable person to confirm that it makes sense. Otherwise someone with a working build could extend `createSampleProject` to split the test project into 2 projects, one that references the other. Then I imagine the existing checks would also cover this fix.

donat:
FYI I've fixed the `ClasspathSeparationTest`.

lukeu:
Just as a quick update: I've not had much time for "hobby coding" the past few weeks (time pressure at work & home - end of school term etc). I did pull this change down one evening but still had trouble - I still couldn't get the UI tests to execute properly locally :-(

I'm super keen to see this change merged however, as it requires an elaborate workaround which would be nice not to have to cherry-pick into each of my branches.

lukeu:
Thanks. I hope to fit in some time over the weekend to wrap these 2 PRs up.

lukeu:
Hi @donat, I've resolved my problem with the UI tests.

I've updated the 7 classpath separation tests to add coverage for the transient nature of the fix, as well as ensuring it works for resources (not just sources). I confirmed 1 of these tests failed before my fix and works afterwards, with JDK-8 and 9.

Testing notes:
 * Launching from Eclipse Oxygen, I get the same number of test failures before and after my fix (5 in core tests, 2 in the ui.view tests). I already fixed one by setting UTF-8 character encoding and another seems to be a locked file resource; so these both sound like issues with running on Windows vs Linux/Mac. Not sure about the rest.
 * I can't build/test using the gradle wrapper on the command-line, as these both give the following:

        [ant:taskdef] Could not load definitions from resource org/apache/maven/artifact/ant/antlib.xml. It could not be found.

Could this be a broken web-link, or perhaps because I don't have Maven installed?

donat:
Thanks .for the update, I'll take a look. 

If something unrelated fails please ignore it, at least in the context of this PR. Let's discuss them separately, in new issues. If you have time for the fix there, that's even better :)

Regarding the `antlib.xml`, the Buildship build itself relies on a local Maven repository that is created by the [scripts](https://github.com/eclipse/buildship/blob/master/buildSrc/src/main/groovy/eclipsebuild/mavenize/DeployMavenAntTaskExecutor.groovy#L32) in `buildSrc` which depends on `org.apache.maven:maven-ant-tasks`. Please let me know if you find out why it's not working for you.

donat:
@lukeu Thank you for the contribution!

lukeu:
Thanks a lot, very happy to see this one go in.

PR #704, opened 2018-05-29
resolve lock the workbench resources when using GradleLaunchConfigura…

simonjhy:
…tion to run task with "--continuous"

### DIFF
--- a/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/operation/internal/DefaultToolingApiOperationManager.java
+++ b/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/operation/internal/DefaultToolingApiOperationManager.java
@@ -13,6 +13,7 @@ import java.util.concurrent.TimeUnit;
 import org.gradle.tooling.CancellationTokenSource;
 import org.gradle.tooling.GradleConnector;
 
+import org.eclipse.core.resources.IWorkspace;
 import org.eclipse.core.resources.IWorkspaceRunnable;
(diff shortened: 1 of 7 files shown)


org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/operation/internal/DefaultToolingApiOperationManager.java
```diff
@@ -38,7 +38,7 @@ public void run(ToolingApiOperation runnable, IProgressMonitor monitor) throws C
     @Override
     public void run(ToolingApiOperation runnable, CancellationTokenSource tokenSource, IProgressMonitor monitor) throws CoreException {
         IProgressMonitor efficientMonitor = new RateLimitingProgressMonitor(monitor, 500, TimeUnit.MILLISECONDS);
-        ResourcesPlugin.getWorkspace().run(new WorkspaceRunnableAdapter(runnable, tokenSource), efficientMonitor);
+        new WorkspaceRunnableAdapter(runnable, tokenSource).run(efficientMonitor);
```
donat: ❌ This makes this use case work incidentally. The proper approach would be to define `ToolingApiOperation#getRule()` that returns an `ISchedulingRule` instance. That rule can be passed to `IWorkspace#run()` and to `ToolingApiJob#setRule()`as well.

simonjhy: thanks donat

PR #706, opened 2018-05-31
resolve lock the workbench resources when using GradleLaunchConfiguration to run task with "--continuous"

simonjhy:
signed by simon.jiang


donat:
Why did you close this PR?

simonjhy:
I found my IP validation didn't pass, so I closed it and resent pull

PR #707, opened 2018-06-01
resolve lock the workbench resources when using GradleLaunchConfiguration to run task with "--continuous"

simonjhy:


org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/operation/ToolingApiJob.java
```diff
@@ -91,7 +97,7 @@ protected void canceling() {
     /**
      * Default handler for the target operation.
      *
-     * @param <T> the result type
+     * @param <T> the result typeToolingApiJobToolingApiJob
```
donat: This looks like an accidental change.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/operation/ToolingApiOperation.java
```diff
@@ -28,6 +29,14 @@
      */
     public String getName();
 
+    /**
+     * Returns the operation Scheduling rule.
```
donat: > Returns the scheduling rule that is applied when the operation is executed synchronously or asynchronously.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/operation/ToolingApiOperation.java
```diff
@@ -28,6 +29,14 @@
...
+     *
+     * @return the operation Scheduling rule
```
donat: ` @return the operation's scheduling rule`

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/operation/ToolingApiOperations.java
```diff
@@ -31,7 +33,7 @@ private ToolingApiOperations() {
      * If the first operation is cancelled then the second operation is not executed.
      *
      * @param op1 the operation to execute first
-     * @param op2 the operation to execute last
+     * @param op2 the operation to execute last        // TODO Auto-generated method stub
```
donat: This looks like an accidental change.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/operation/ToolingApiOperations.java
```diff
@@ -56,6 +58,11 @@ public void runInToolingApi(CancellationTokenSource tokenSource, IProgressMonito
 
                 op2.runInToolingApi(tokenSource, progress.newChild(1));
             }
+
+            @Override
+            public ISchedulingRule getRule() {
+                return ResourcesPlugin.getWorkspace().getRoot();
```
donat: `MultiRule` should be returned.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/operation/internal/DefaultToolingApiOperationManager.java
```diff
@@ -38,7 +39,7 @@ public void run(ToolingApiOperation runnable, IProgressMonitor monitor) throws C
     @Override
     public void run(ToolingApiOperation runnable, CancellationTokenSource tokenSource, IProgressMonitor monitor) throws CoreException {
         IProgressMonitor efficientMonitor = new RateLimitingProgressMonitor(monitor, 500, TimeUnit.MILLISECONDS);
-        ResourcesPlugin.getWorkspace().run(new WorkspaceRunnableAdapter(runnable, tokenSource), efficientMonitor);
+        ResourcesPlugin.getWorkspace().run(new WorkspaceRunnableAdapter(runnable, tokenSource), runnable.getRule(), IWorkspace.AVOID_UPDATE, efficientMonitor);
```
donat: Can you elaborate explain why the `AVOID_WORKSPACE` is necessary here?

simonjhy:
After change I should push a new commit or still use same commit?

simonjhy:
hi, donat. I added an integration test in RunGradleBuildLaunchRequestJobTest. I think it isn't need to add UI SWT test to check this fix. Is this ok?

donat:
> After change I should push a new commit or still use same commit?

It's up to you, the GH reviews can handle both. Personally, I reuse the commits and use force-push to rewrite remote history when I'm working on a feature branch and I want to avoid having lots of small commits.

donat:
> I think it isn't need to add UI SWT test to check this fix. Is this ok?

simonjhy:
ok, I will merge two commits into one and sent again.

donat:
PR merged. Thank you for the contribution!

simonjhy:
Thanks you, happy to see it merged

PR #717, opened 2018-07-12
Project nodes should always top than taskgroup nodes

Camork:


org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/view/task/TaskNodeViewerSorter.java
```diff
@@ -78,25 +82,30 @@ public int compare(Viewer viewer, Object leftNode, Object rightNode) {
...
 
-    private static Ordering<ProjectNode> createProjectNodeOrdering() {
+    private static Ordering<ProjectNode> createProjectNodeOrdering(final boolean isHierarchyFlattened) {
         return new Ordering<ProjectNode>() {
 
             @Override
             public int compare(ProjectNode left, ProjectNode right) {
-                EclipseProject leftRoot = HierarchicalElementUtils.getRoot(left.getEclipseProject());
-                EclipseProject rightRoot = HierarchicalElementUtils.getRoot(right.getEclipseProject());
-                if (leftRoot == rightRoot) {
```
Camork: The root project which `HierarchicalElementUtils.getRoot` returned isn't the same instance, but actual is the same gradle project.

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/view/task/TaskNodeViewerSorter.java
```diff
@@ -78,25 +82,30 @@ public int compare(Viewer viewer, Object leftNode, Object rightNode) {
...
-                    // do not change sorting of projects that belong to the same root
-                    return 0;
-                } else {
-                    // projects that do not belong to the same root should be grouped by the name of
-                    // their root projects
-                    return leftRoot.getName().compareTo(rightRoot.getName());
+                if (!isHierarchyFlattened) {
+                    return left.toString().compareTo(right.toString());
```
donat: ❌Should be `leftRoot.getName().compareTo(rightRoot.getName())`.

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/view/task/TaskNodeViewerSorter.java
donat: That's correct, thanks for the explanation.

donat:
Here's a sample test that should be part of `TaskViewContentTest`:
```
    def "Project nodes should come before task group nodes if hierarchy is not flattened"() {
        setup:
        view.state.projectHierarchyFlattened = false
        
        when:
        def project = dir("root") {
            file 'build.gradle', """
                task foo() {
                    group = 'b'
                }
            """
            file 'settings.gradle', """
                include 'z'
                include 'a'
            """
            dir('z')
            dir('a')
        }
        importAndWait(project)
        waitForTaskView()

        then:
        waitFor { taskTree.root.collect { k, v -> k }.findAll { ['a', 'b', 'z'].contains(it) } == ['a', 'z', 'b'] }
    }
```

I'd expect a similar test verifying that the root comparison is now correct.

donat:
Here's how you run the tests in Eclipse: https://github.com/eclipse/buildship/blob/master/docs/development/Setup.md#running-the-tests-in-eclipse

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/view/task/TaskNodeViewerSorter.java
Camork: Hi @donat , I am curious about this point. Since we grouped child nodes into parent node folder. Child project will only compare to the node that belongs to the same parent. Why we need to sort them by their parent (nothing to be sorted)?

donat: My mistake. What I meant to write is to sort the nodes by the displayed name. I'd do the following:
- Define `ProjectNode.getDisplayName()` and move the implementation there from `TaskNameLabelProvider.getProjectText()`
- Here, return `left.getDisplayName().compareTo(right.getDisplayName())`

donat:
@Camork Thank you for the contribution!

PR #719, opened 2018-07-18
[DO NOT MERGE] Import API prototype

donat:
Here's a first take on the import API. I share this wor as I'd like to discuss the possible paths we can go down. 

Currently, the entry point is the `GradleWorkspace` class:

```
public interface GradleWorkspace {
    BuildIdentifier newBuildIdentifier(File projectLocation, GradleDistribution gradleDistribution);
    void performImport(BuildIdentifier id, IProgressMonitor monitor, Class<? extends GradleProjectConfigurator> configuratorClass) throws GradleException;
}
```

The caller side:
```
try {
    GradleWorkspace gradleWorkspace = GradleCore.workspace();
    BuildIdentifier id = gradleWorkspace.newBuildIdentifier(loc);
    gradleWorkspace.performImport(id, mon, MyConfigurator.class);					
} catch (Exception e) {
    ...
}
```

The `GradleConfigurator` class is defined via an extension point. It is needed because:
- external plugins want to adjust the configuration after each synchronization, and 
- Buildship wouldn't know about the configurators when the synchronization is called from the inside. 

The things to consider:
- the `performImport()` method returns `void`. Maybe we should return the references of the added/removed/renamed projects.
- Do we need the `GradleConfigurator` extension point? I did this because it's required for the STS plugin to work. What they do is quite simple atm: they ensure that the Spring nature is applied on all projects. Is there another way how we can do that?
- Job rules. Although I haven't implemented it yet, I think the `performImport` should internally acquire the workspace root rule, just like the internal `SynchronizationJob` and `SynchronizeOperation` does. This way the client doesn't have to worry about the job rules at all.
- `GradleProjectConfigurator.configure()` is called for each for each open project. There the client doesn't have the context of what project was added/modifed/deleted. The `configure()` method should be called once, and the `context` object should hold a reference to the added/removed projects. 
- A small addition to the previous point. If we also add the TAPI models to the `context` object then we can dogfood this API into Buildship. One configurator would set up the `IProject`s  (builders, natures, etc.) and another one would do the Java projects.


donat:
Another thing occurred to me. There are two concerns in my proposal:
- API to import a project
- API to participate in the synchronization logic.
I think we should have two separate APIs for that matter. The import API should just invoke the project synchronization logic and return the result. The synchronization participant (`GradleProjectConfigurator ` in the prototype) should be independent. A client can always 1) import the project, 2) investigate and mark the imported projects based on the results, and 3) interact with only the marked projects in the synchronization participant extension point.

oehme:
I agree, the configurator should not be in the import method signature, but called transparently. I would also separate the two concerns and focus on just the import API for now.

oehme:
The `performImport` method should return a `Status` containing all the problems that potentially occured. The method should only throw an error when some fatal problem occurs. Everything else should be handled with error markers on the project and the returned status object.

oehme:
Yes, both the configurator and some other things like task launching should be part of 3.0. I just want to go one step at a time. Can we first do a Google doc laying out the use cases we want to support and the APIs we want to provide for them?

donat:
Yes, this is my plan. The prototype was mainly used to have an initial feeling what it takes to implement such an API, plus to check how can I integrate with STS locally.

donat:
https://docs.google.com/document/d/1tZ8kAVayZ1fnoQjmcgKOV63c-L0h-x21OTJCyT5Waxo/edit#
I'll share the contents with the community once we discussed it internally.

gamerson:
Hey @donat  and @oehme  thanks for your efforts.  We are excited to see this Project Configurator type extension points coming as API.  It will help us solve some headaches we have as it comes to "configuring Liferay module projects" that use gradle as the build.  It fills a very much needed gap that the m2e-project-configurators have filled nicely over the years.  We can be some testers for you once you feel you want some outside testing.

donat:
@gamerson Thanks for the interest! We'll let you know if we have something to show. FYI we'll post the API spec at https://github.com/eclipse/buildship/issues/724 once we discussed it internally.

Issue #621, opened 2017-12-11
Add Eclipse Source Reference Headers to MANIFEST.MF


fbricon:
Can you please add Eclipse Source Reference Headers (`Eclipse-SourceReferences` attribute) to Buildship's various MANIFEST.MFs, as described in https://wiki.eclipse.org/PDE/UI/SourceReferences, so that we can easily reference the source code associated with the various binaries.

donat:
@fbricon Is it possible to specify a commit hash in the header entry? Something like 

```
Eclipse-SourceReferences: scm:git:https://github.com/eclipse/buildship.git;tag=2e93d866714a6274bd27f50d7bca322597f2b73
```

If so then we can associate all releases with the sources.

donat:
Sorry, found my answer. FTR: 
``` 
Eclipse-SourceReferences: scm:git:git://git.eclipse.org/gitroot/platfo
 rm/eclipse.platform.team.git;path="bundles/org.eclipse.compare";tag="
 I20180418-2000";commitId=b771accd19e5806552c113be35e86e4f8bd9429b
```

donat:
Fixed by https://github.com/eclipse/buildship/commit/4330972ee476ae203d3df81535064ee5668cb2f3 on master and https://github.com/eclipse/buildship/commit/aca9a98b3eb3b4a6e98308aad0a6292ef0074f48 on release.

PR #720, opened 2018-07-20
issue-697

moley:
Hi, I have fixed the issue, can you please review?
If you have any suggestions please call

### DIFF
--- a/org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/workspace/internal/ImportingWtpProjects.groovy
+++ b/org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/workspace/internal/ImportingWtpProjects.groovy
@@ -11,8 +11,9 @@ import org.eclipse.jdt.core.IAccessRule
 import org.eclipse.jdt.core.IClasspathAttribute
 import org.eclipse.jdt.core.IClasspathEntry
 import org.eclipse.jdt.core.JavaCore
-
+import org.eclipse.buildship.core.CorePlugin
 import org.eclipse.buildship.core.Logger
+import org.eclipse.buildship.core.notification.UserNotification
 import org.eclipse.buildship.core.test.fixtures.ProjectSynchronizationSpecification
 import org.eclipse.buildship.core.workspace.GradleClasspathContainer
 import org.eclipse.buildship.core.workspace.WorkspaceOperations
@@ -23,6 +24,62 @@ class ImportingWtpProjects extends ProjectSynchronizationSpecification {
     private static final String NON_DEPLOYED = "org.eclipse.jst.component.nondependency"
     private static final String DEPLOYED = "org.eclipse.jst.component.dependency"
     private static final String WTP_COMPONENT_NATURE = "org.eclipse.wst.common.modulecore.ModuleCoreNature"
+
+    def "Do not check mixed deployment paths without WTP installed"() {
+        setup:
+        File root = dir("project") {
+            file 'build.gradle', """
+                apply plugin: 'java'
+                apply plugin: 'ear'
+                ${jcenterRepositoryBlock}
+                dependencies {
+                    earlib 'org.assertj:assertj-core:3.10.0'
+                    deploy "log4j:log4j:1.2.17"
+                }
+            """
+        }
+
+        UserNotification notification = Mock(UserNotification)
+        registerService(UserNotification, notification)
+
+        WorkspaceOperations operations = new WorkspaceOperationsDelegate(recognizeWtpComponentNature: false, wtpInstalled:false)
+        registerService(WorkspaceOperations, operations)
+
+        when:
+        importAndWait(root)
+
+        then:
+        0 * notification.errorOccurred(*_)
+
+    }
+    
+    def "Check mixed deployment paths with WTP installed"() {
+        setup:
+        File root = dir("project") {
+            file 'build.gradle', """
+                apply plugin: 'java'
+                apply plugin: 'ear'
+                ${jcenterRepositoryBlock}
+                dependencies {
+                    earlib 'org.assertj:assertj-core:3.10.0'
+                    deploy "log4j:log4j:1.2.17"
+                }
+            """
+        }
+
+        UserNotification notification = Mock(UserNotification)
+        registerService(UserNotification, notification)
+
+        WorkspaceOperations operations = new WorkspaceOperationsDelegate(recognizeWtpComponentNature: false, wtpInstalled:true)
+        registerService(WorkspaceOperations, operations)
+
+        when:
+        importAndWait(root)
+
+        then:
+        1 * notification.errorOccurred(*_)
+
+    }
 
     def "The eclipseWtp task is run before importing WTP projects"() {
         setup:
@@ -381,6 +438,11 @@ class ImportingWtpProjects extends ProjectSynchronizationSpecification {
     class WorkspaceOperationsDelegate {
         @Delegate WorkspaceOperations delegate = new DefaultWorkspaceOperations()
         boolean recognizeWtpComponentNature
+        boolean wtpInstalled = false
+
+        boolean isWtpInstalled () {
+        	wtpInstalled
+        }
 
         boolean isNatureRecognizedByEclipse(String nature) {
             nature == WTP_COMPONENT_NATURE ? recognizeWtpComponentNature : delegate.isNatureRecognizedByEclipse(nature)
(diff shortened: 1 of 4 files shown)


donat:
I'll talk a look today.

donat:
In any case, can you please fix the IP validation failure? Your commit needs a signed-off-by footer with a user that has signed the ECA.

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/workspace/internal/ImportingWtpProjects.groovy
```diff
@@ -23,6 +24,34 @@ class ImportingWtpProjects extends ProjectSynchronizationSpecification {
     private static final String NON_DEPLOYED = "org.eclipse.jst.component.nondependency"
     private static final String DEPLOYED = "org.eclipse.jst.component.dependency"
     private static final String WTP_COMPONENT_NATURE = "org.eclipse.wst.common.modulecore.ModuleCoreNature"
+	
+	def "Do not check mixed deployment paths without WTP installed"() {
```
donat: ❌ The test fails in my local eclipse as I have WTP installed. I'd fix it the following way:
- Introduce the `isWtpInstalled()` method on the `WorkspaceOperations` interface.
- Adjust `WorkspaceOperationsDelegate` to return a configurable result for `isWtpInstalled`. 
- Duplicate this test to cover both the use cases when WTP is and is not installed.

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/workspace/internal/ImportingWtpProjects.groovy
```diff
@@ -23,6 +24,34 @@ class ImportingWtpProjects extends ProjectSynchronizationSpecification {
...
+		registerService(UserNotification, notification)
+
+		WorkspaceOperations operations = new WorkspaceOperationsDelegate(recognizeWtpComponentNature: false)
+		registerService(WorkspaceOperations, operations)
+
+		when:
+		importAndWait(root)
+
+		then:
+        0 * notification.errorOccurred(*_)
```
donat: Wrong indentation

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/WtpClasspathUpdater.java
```diff
@@ -42,13 +43,15 @@
     private static final String NON_DEPLOYMENT_ATTRIBUTE = "org.eclipse.jst.component.nondependency";
 
     public static void update(IJavaProject javaProject, OmniEclipseProject project, SubMonitor progress) throws JavaModelException {
+      if (Platform.getBundle("org.eclipse.wst.common.core") != null) {
```
donat: ❌ move the check to the `WorkspaceOperations.isWtpInstalled()`  method in order to make it testable.

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/workspace/internal/ImportingWtpProjects.groovy
Frontrider: @moley for this project, set up your formatter to be the same as this repository. If you don't use one, I can recommend it.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/WtpClasspathUpdater.java
moley: ok thanks for you suggestions, I will proceed on monday. 
The formatter is not automatically set with oomph? can you please give me a link to the correct formatter?

Frontrider: https://help.eclipse.org/neon/index.jsp?topic=%2Forg.eclipse.jdt.doc.user%2Freference%2Fpreferences%2Fjava%2Fcodestyle%2Fref-preferences-formatter.htm
https://www.jetbrains.com/help/idea/configuring-code-style.html

donat: Yes, the Buildship Oomph setup contains the formatting configuration.

moley: Sorry - I don't have it yet... The link of @frontrider leads me to the eclipse formatter preferences mechanism - ok. There are three options I could choose (Eclipse [built-in], Eclipse 2.1 [built-in] and Java-Conventions [built-in]. I think none of this tree options is the right one? 
I have set up my workspace with oomph. Which formatter do I have to choose? Is it somewhere commited to git in the projects?

donat: Oomph should set up a new Java formatter called `Buildship`
screen shot 2018-07-24 at 15 27 59
If it doesn't work for you, then just ignore this for this PR. I'll look into the Oomph setup why it's not working for you.

moley: Sorry, doesn't work, I don't have the buildship formatter in place. I tried to adapt the format manually best possible, don't now if it is 100% exact... So I changed the isWtpInstalled thing and added a second test...
If you have any requests, please call. 
If you need more info to fix the formatter thing, also call. 

Cheers
Markus

moley:
no problem, I have done these commands.

donat:
It seems that only your last commit has the `signed-off-by` footer. Can you please make sure that all commits have that?

moley:
Hi again, I think I've done, please check

moley:
Do you still wait for me to do anything?

donat:
@moley Sorry for the late response, I was on holiday. You don't have to do anything, I'll merge your PR soon.

donat:
@moley Thank you for the contribution!

moley:
@donat No problem, you're welcome. Thank you for merging it, should I delete my branch moley:moley/issue-697?

moley:
@donat 
Hi, 
Did you already build a snapshot, which contains this fix. Can you please give me the url?

moley:
Ok, thank you so far

moley:
@donat 
I have downloaded this versions and the error still occurs :-( Did you check if my change was working on your machine? Did I miss something?

moley:
Different question: I didn't find an uploaded source jar. 
How can I check if my change is contained?

Frontrider:
I woukd like to test the change.

donat:
@moley The change was merged on 7th August, all [snapshot releases](http://download.eclipse.org/buildship/updates/e48/snapshots/2.x/) should contain your change after that.

donat:
Also, we recently [added](https://github.com/eclipse/buildship/issues/621) source reference headers to the jars which contain the commit hash from which the update site was generated. You can use it to trace the history whether your change is included. Or you can use the PDE import feature (referenced at #621), but AFAIC it only works for the Buildship 2.x versions, because the 3.x development is is done on a feature branch.

donat:
@Frontrider You can recreate the sample projects from the [integration tests](https://github.com/eclipse/buildship/pull/720/files#diff-65516a651c52091d6edb472bf08ce11b). Create a project, copy-paste the configuration to your `build.gradle` file and call `Gradle > Refresh Gradle Project`. If everything is right you should not see an exception mentioning mixed WTP deployment paths.

moley:
@donat, @Frontrider Hi, any new insights on this?

Frontrider:
We ended up not using it because the project also has some RCP bits, that will not import at all if we use gradle.

moley:
Hi again @donat 

I retestet my pullrequest and noticed, that the fix was not enough, because it seems that eclipse photon comes with org.eclipse.wst.common_core.feature and org.eclipse.wst.common_ui.feature by default :-( 
So I would suggest to replace Exception by gradle log error or warning. May I open another pullrequest to change implementation like this? Do you have other suggestions to solve this problem?

moley:
Yes, that's the point - as soon as `org.eclipse.wst.common.core` plugin is available, the error is shown. I wanted to avoid this error thinkin that this plugin is not installed by default. But - not using the target platform and downloading the real eclipse - I noticed it is installed. So I think I will try your idea of using release-3.0 branch

Thank you

moley:
Hi again, 

I have tried with branch release-3.0, which did not work. Now I have removed the lines: 
                  ```
  if (deploymentPath != null && !deploymentPath.equals(attribute.getValue())) {
                        throw new UnsupportedConfigurationException("WTP currently does not support mixed deployment paths.");
                    }
```
and created a company internal updatesite, were I installed this version from, which worked. So - what about removing this exception?

PR #735, opened 2018-08-22
Reorganize packages part 1/2

donat:
This pull request adjusts the project's package layout so that the upcoming public API can be placed in the `org.eclipse.buildship.core` package.

The following changes have been made:
- all packages are moved into the `org.eclipse.buildship.<plugin_id>.internal` base package (except the invocation customizer in `org.eclipse.buildship.core.invocation` which is already a public API).
- the packages with `.internal` suffix have been renamed to `.impl`
- the `org.eclipse.buildship.ui.external.viewer` package has been renamed to `org.eclipse.buildship.ui.internal.extviewer`.

The plugin IDs, extension IDs, etc. have remained unchanged.

The changeset is so big it must have hit sort of limit in GitHub. I had to split this PR (Part 2: https://github.com/eclipse/buildship/pull/736), otherwise the unchanged files would be displayed as a new and a deleted file, effectively making the diff unusable. Please handle the two reviews as one.


oehme:
> the packages with .internal suffix have been renamed to .impl

Why two different suffixes? 

>the org.eclipse.buildship.ui.external.viewer package has been renamed to org.eclipse.buildship.ui.internal.extviewer

internal.external.viewer? ;) Why does it have the word "external" in it anyway?

donat:
The original `internal` packages were meant to be hidden from the `ui` plugin. We can replace it with package-private classes.

The `external` package contains a custom widget implementation (`FilteredTree`) that was copied over from another Eclipse plugin.

oehme:
So there's really no reason to call it "external". It doesn't matter where the code came from, that's covered in the license header.

donat:
That's correct. I guess the `org.eclipse.buildship.ui.internal.util.widget` package would be a better location for the classes.

PR #746, opened 2018-08-22
Create entry interface to access upcoming API services

donat:


org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/GradleCoreTest.groovy
```diff
@@ -0,0 +1,11 @@
+package org.eclipse.buildship.core
+
+import spock.lang.Specification;
+
+class GradleCoreTest extends Specification {
+
+    def "All API services are available"() {
+        expect:
+        GradleCore.workspace.configurationManager
```
oehme: This test is meaningless, it's basically just testing "does Javac still work"? Let's just remove it until there is actually something to test.

PR #747, opened 2018-08-23
Raise minimum source and runtime version to Java 8 

donat:
This change removes all support for Java versions <= 1.7.

I had to drop support for Eclipse 4.2 as it can't run on Java 8. Also, Eclipse I added the installable Java 8 support plugin to the Eclipse 4.3 target platform. The build pipeline has been adjusted accordingly: https://github.com/eclipse/buildship/compare/b9ab6b79aacefca04f7040c0a1dc8c04170387ad...e9686aca407b667b39fd908d305dfd08b88c50d8

The source and target compatibility are also set to `1.8` and there's a sample Java 8 snippet in the code to verify that we can use Java 8 constructs after the merge.


org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/workspace/internal/GradleClasspathContainerUpdater.java
```diff
@@ -84,13 +82,7 @@ private void updateClasspathContainer(PersistentModelBuilder persistentModel, IP
...
         List<IClasspathEntry> projectDependencies = collectProjectDependencies();
 
-        boolean hasExportedEntry = FluentIterable.from(externalDependencies).anyMatch(new Predicate<IClasspathEntry>() {
-
-            @Override
-            public boolean apply(IClasspathEntry entry) {
-                return entry.isExported();
-            }
-        });
+        boolean hasExportedEntry = externalDependencies.stream().anyMatch(IClasspathEntry::isExported);
```
oehme: It's gonna be a relief to finally code like it's the 2000s ;)

Issue #718, opened 2018-07-12
Raise minimum execution environment to Java 8


donat:
Java 7 reached EOL 3 years ago already. Buildship has Java 7 builds because few older Eclipse releases specify Java 7 as a minimal execution environment. Supporting Java 7 has become harder and harder. The latest example is that the build broke because Maven Central and Bintray [discontinued the support for older TLS protocols](https://blog.gradle.org/unable-to-download-maven-central-bintray). Also, this prevents us from using newer Java features, such as lambdas.

We raise the minimum Java version to Java 8. Starting with Eclipse 4.6 until the latest release that's also the minimal required version. For the older Eclipse distros, we can simply require higher versions in the plugin manifest.

This is a breaking change, but the 3.0 release seems like the best time to do it.

PR #749, opened 2018-08-24
[WIP] Move GradleDistribution to public API

donat:
The PR requires #718 to be merged.


org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/GradleDistributionInfo.java
```diff
@@ -0,0 +1,89 @@
...
+
+import org.eclipse.buildship.core.internal.DefaultGradleDistributionInfo;
+
+/**
+ * Represents a valid or invalid {@link GradleDistribution}.
+ *
+ * @author Donat Csikos
+ * @since 3.0
+ */
+public abstract class GradleDistributionInfo {
```
oehme: ❌ All of this is part of our preference infrastructure and not really interesting to other clients. We can cast to an internal interface to get the information we want. The public API should only have `GradleDistribution`

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/launch/JavaHomeTab.java
```diff
@@ -130,9 +130,9 @@ public boolean isValid(ILaunchConfiguration launchConfig) {
             return false;
         }
 
-        File javaHome = FileUtils.getAbsoluteFile(javaHomeResolved).orNull();
+        File javaHome = FileUtils.getAbsoluteFile(javaHomeResolved).orElse(null);
```
oehme: This shows that we shouldn't be using `Optional` here, because we just want null anyway ;) We should just return null when there is no error. 

Optional is only meant for cases where returning null has big potential for confusion, e.g. `stream.findFirst` (where null could mean "the first element is null" or "there is no first element").

PR #754, opened 2018-08-28
Define initial API to synhronize existing projects

donat:
This PR adds a public facade to the internal synchronization API to synchronize existing projects. The test coverage has been adjusted so that existing projects are also synchronized via the API.


org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/GradleWorkspaceTest.groovy
```diff
@@ -0,0 +1,50 @@
...
+        IProject project = newClosedProject("GradleWorkspaceTest")
+
+        when:
+        GradleCore.workspace.getBuild(project)
+
+        then:
+        thrown(IllegalArgumentException)
+    }
+
+    def "Cannot get reference to Gradle build if target project is not a Gradle project"() {
```
oehme: ❌ Using an exception for this feels wrong, especially since this method would essentially be the only way to test whether something is indeed a Gradle project. Should probably return null instead or use an Optional return type.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/GradleBuild.java
```diff
@@ -0,0 +1,28 @@
...
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+
+package org.eclipse.buildship.core;
+
+import org.eclipse.core.runtime.IProgressMonitor;
+
+/**
+ * Represents a Gradle build that can be imported into or synchronized with the workspace.
```
oehme: ...and many other things. I don't think this adds much to the description, since I can see what I can do with this by looking at the methods. If there is nothing more to say than "Represents a Gradle build", that's fine.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/GradleBuild.java
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * @author Donat Csikos
+ * @since 3.0
+ */
+public interface GradleBuild {
+
+    /**
+     * Synchronizes this build with the workspace.
```
oehme: ❌ I think it's the other way round. This definitely needs more documentation. What does synchronization mean? Are there any special conditions for calling this? Does it run synchronously or in the background etc...

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/GradleBuild.java
```diff
@@ -0,0 +1,28 @@
...
+     *
+     * @param monitor the progress monitor
```
oehme: I find comments like these unhelpful. They're just noise, repeating what the types are saying.  If there is nothing else to say about them, why not leave them out?

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/DefaultGradleBuild.java
```diff
@@ -0,0 +1,68 @@
...
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.Status;
+
+import org.eclipse.buildship.core.GradleBuild;
+import org.eclipse.buildship.core.SynchronizationResult;
+import org.eclipse.buildship.core.internal.workspace.NewProjectHandler;
+
+/**
+ * Default implementation for {@link GradleBuild}.
```
oehme: Obvious comment is obvious :)

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/DefaultGradleBuild.java
```diff
@@ -0,0 +1,68 @@
...
+ */
+public final class DefaultGradleBuild implements GradleBuild {
+
+    private final IProject project;
```
oehme: ❌ A gradle build is not bound to a project, but to a root directory

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/DefaultGradleBuild.java
```diff
@@ -0,0 +1,68 @@
...
+        return new SynchronizationResult() {
+
+            @Override
+            public IStatus getStatus() {
+                return result;
+            }
+        };
+    }
+
+    public static GradleBuild from(IProject project) {
```
oehme: Unnecessary indirection

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/GradleWorkspaceTest.groovy
donat: Replaced with `Optional`.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/GradleBuild.java
donat: I've added some documentation to the interace.

donat: Makes sense. I've removed the unnecessary parts.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/GradleBuild.java
donat: I added extra javadoc there.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/GradleBuild.java
donat: It's kind of obvious what it does, although I'd still prefer to have an entry for all params for a public API.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/DefaultGradleBuild.java
donat: This is a long-standing request of yours. We have a checkstyle rule in place that enforces to have minimal comment on the classes.  I guess it's the right time to get rid of it.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/DefaultGradleBuild.java
donat: Or more precisely it's bound to a `BuildConfiguration`. It will be changed in the subsequent PRs.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/DefaultGradleBuild.java
donat: Deleted.

donat:

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/GradleBuild.java
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * Synchronizes the workspace with this Gradle build.
+     *
+     * <p>
+     * The method loads the Gradle build configuration and updates the workspace based on the
+     * retrieved information.The algorithm is as follows:
+     * <ul>
+     *   <li>Synchronize all Gradle projects of the Gradle build with the Eclipse workspace
+     *       project counterparts. If there is a project in the workspace at the location of the
```
oehme: "If there is a project in the workspace at the location of theGradle project..."

And if there isn't, we import it.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/GradleWorkspace.java
```diff
@@ -0,0 +1,34 @@
...
+
+    /**
+     * Returns a reference to a Gradle build containing the target project.
+     * <p>
+     * If the target project is not accessible or not part of a Gradle build then the method returns
+     * {@code Optional#empty()}.
+     *
+     * @param project the target project
+     * @return the Gradle build or {@code Optional#empty()} for non-Gradle projects.
+     * @throws NullPointerException if project is null
```
oehme: I wonder if we should return Optional.empty instead.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/GradleBuild.java
donat: Added clarification to the section.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/GradleWorkspace.java
donat: I like the `Optional.empty` approach better. Fixed.

Issue #732, opened 2018-08-16
Implement hierarchical preferences API


donat:
# Overview

This is a pre-requisite for the import API implementation. In order to import new projects into the workspace, clients need to be able to specify build configuration for the project. We already have an internal API for hierarchical preferences but it has some modeling and redundancy problems and  that should be cleaned up before making it public.

# Design

The API should support three layers of configuration:
- The first is the workspace level, which declares configuration for all Gradle projects in the workspace. 
- The second is the build configuration which applies to a collection of workplace project that belongs to the same Gradle build. A build configuration can either inherit workspace settings or override it completely. 
- The last one is the run configuration. It stores the properties to execute a Gradle build. It can either inherit or override the project configuration settings.

# Classes
The entry point for the API services should be the `org.eclipse.buildship.core.GradleCore` class. It should provide a static reference to the Gradle workspace service.
```
class GradleCore {
    static GradleWorkspace getWorkspace()
}
```

The `GradleWorkspace` service should contain only a reference to the configuration manager object.

New instances can be created with builders through the `new[Workspace|Build|Run]ConfigurationBuilder()` methods.

The workspace configuration is quite simple as only one instance belongs to this scope. The existing configuration can be loaded from the current workspace and can be modified with the save method.

The build configuration is associated with a set of workspace projects. A client can enumerate all build configuration available in the workspace or can read a configuration from an external location. A workspace project is associated with the build when the `saveBuildConfiguration` is called.

A run configuration contains all properties to launch a Gradle build. It is linked with a(n external) build configuration and persisted into a platform launch configuration.

```
interface ConfigurationManager {
    WorkspaceConfigurationBuilder newWorkspaceConfigurationBuilder()
    WorkspaceConfiguration loadWorkspaceConfiguration()
    void saveWorkspaceConfiguration(WorkspaceConfiguration configuration)

    BuildConfigurationBuilder newBuildConfigurationBuilder(File rootProjectDirectory)
    Collection<BuildConfiguration> loadAllBuildConfigurations()
    BuildConfiguration loadExternalBuildConfiguration(File projectDirectory)
    BuildConfiguration loadBuildConfiguration(IProject project)
    void saveBuildConfiguration(IProject project, BuildConfiguration configuration)
    void deleteBuildConfiguration(IProject project)

    RunConfigurationBuilder newRunConfigurationBuilder(BuildConfiguration configuration)
    Collection<RunConfiguration> loadAllRunConfigurations()
[...]

PR #757, opened 2018-08-29
Add initial API to import a new Gradle project

donat:


org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/DefaultGradleBuild.java
```diff
@@ -22,28 +21,37 @@
...
+import org.eclipse.buildship.core.configuration.BuildConfiguration;
 import org.eclipse.buildship.core.internal.workspace.NewProjectHandler;
 
 /**
  * Default implementation for {@link GradleBuild}.
  */
 public final class DefaultGradleBuild implements GradleBuild {
 
-    private final IProject project;
+    private final Optional<org.eclipse.buildship.core.internal.workspace.GradleBuild> gradleBuild;
```
oehme: ❌ Never use Optional as a field type. We have `null` for that. This just creates unnecessary noise and garbage.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/configuration/BuildConfiguration.java
```diff
@@ -0,0 +1,25 @@
...
+
+package org.eclipse.buildship.core.configuration;
+
+/**
+ * Describes a configuration of a Gradle build.
+ *
+ * @author Donat Csikos
+ *
+ */
+public interface BuildConfiguration {
```
oehme: I think this should be a public value object. We can handle the connection between workspace config, build config and run configurations behind the scenes. No extra public service necessary.

donat: How exactly should we create new configuration objects then? With the [current design](https://github.com/eclipse/buildship/issues/732), the configuration manager object is responsible for the object creation/loading/storing. Shall we expose the builder objects as separate classes?

donat:
@oehme I wonder what should we do when `GradleWorkspace.createBuild()` is called on a nonexistent directory? Shall we keep the existing behavior and throw an exception, or do the same as with `getBuild()`: return an `Optional` to signal that we can't create a build there.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/DefaultGradleBuild.java
donat: Fixed.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/configuration/BuildConfiguration.java
oehme: Yep: `BuildConfiguration.forRootDirectory().withGradleDistribution().build`

oehme:
I don't think there's anything wrong with creating a GradleBuild object holding a currently non-existing directory. It will fail later when trying to synchronize it. But if the user creates that directory in the meantime, everything is fine. The same issue happens in the other direction - Some creates a `GradleBuild` object, but then deletes the underlying directory. So doing the check at object creation time doesn't really help much.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/configuration/BuildConfiguration.java
donat: Done.

donat:
Makes sense. Removed the precondition checking for the existence of the root project directory.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/configuration/BuildConfigurationBuilder.java
```diff
@@ -0,0 +1,20 @@
...
+
+package org.eclipse.buildship.core.configuration;
+
+/**
+ * Creates {@link BuildConfiguration} instances.
+ *
+ * @author Donat Csikos
+ * @since 3.0
+ */
+public interface BuildConfigurationBuilder {
```
oehme: ❌ This doesn't need to be an interface, you can just make the Builder a nested class of BuildConfiguration and give it a private constructor.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/configuration/BuildConfiguration.java
```diff
@@ -0,0 +1,35 @@
...
+
+import org.eclipse.buildship.core.internal.configuration.DefaultBuildConfigurationBuilder;
+
+/**
+ * Describes a configuration of a Gradle build.
+ *
+ * @author Donat Csikos
+ * @since 3.0
+ */
+public abstract class BuildConfiguration {
```
oehme: ❌ Shouldn't be abstract. Instead give it a private constructor to prevent instantiation.

donat: I want to build a proper builder pattern where all the implementation details are hidden in the concrete class. How about adding the `@noimplement` annotation and a protected constructor?

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/configuration/BuildConfigurationBuilder.java
donat: I see, but I'd prefer to hide all implementation details in a separate class.

oehme: For what purpose? What is there to hide? Let's not make it unnecessarily complex :)

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/configuration/BuildConfiguration.java
oehme: This is all unnecessary. The builder pattern does not require an interface. It just requires private constructors.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/configuration/BuildConfigurationBuilder.java
oehme: By that explanation `java.lang.Integer` should be an interface and there should be `IntegerImpl` and a factory method. But ask yourself: Did you *ever* have any problem with the fact that Integer has all the code it needs inside it?

donat: Sure, I understand that. I just have the implementation already and I quite liked it the way it is. I'll change that soon.

oehme: > I'll soon adjust the implemementation.

I.e. you'll remove the unnecessary indirection?

donat: Yes.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/configuration/BuildConfiguration.java
donat: I know how it works, I only wanted add a twist to it :)

donat: Fixed.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/configuration/BuildConfigurationBuilder.java
donat: Fixed.

donat:

PR #758, opened 2018-08-29
Let clients declare Gradle distribution for imported projects

donat:


org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/GradleDistribution.java
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * @author Donat Csikos
+ * @since 3.0
+ */
+public abstract class GradleDistribution {
+
+    /**
+     * The available Gradle distributions types.
+     */
+    public enum Type {
```
oehme: ❌ Should not be public, as previously discussed. What would be the purpose of making this public?

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/GradleDistribution.java
```diff
@@ -0,0 +1,109 @@
...
+         */
+        VERSION
+    }
+
+    /**
+     * Returns the distribution type.
+     *
+     * @return the distribution type
+     */
+    public abstract Type getDistributionType(); // TODO (donat) rename to getType() and merge public and internal enums
```
oehme: ❌ Should not be public.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/GradleDistribution.java
```diff
@@ -0,0 +1,109 @@
...
+     * <p>
+     * For {@link GradleDistributionType#VERSION} the version is returned. For
+     * {@link GradleDistributionType#LOCAL_INSTALLATION} the installation location path is returned.
+     * for {@link GradleDistributionType#REMOTE_DISTRIBUTION} the remote URI is returned. For
+     * {@link GradleDistributionType#WRAPPER} and for {@link GradleDistributionType#INVALID} an
+     * empty string is returned.
+     *
+     * @return the configuration string
+     */
+    public abstract String getConfiguration();
```
oehme: ❌ Should not be public

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/GradleDistribution.java
```diff
@@ -0,0 +1,109 @@
...
+import java.io.File;
+import java.net.URI;
+
+/**
+ * Represents a Gradle distribution.
+ *
+ * @author Donat Csikos
+ * @since 3.0
+ */
+public abstract class GradleDistribution {
```
oehme: ❌ Should be marked as `@noimplement`. (this applies to other interfaces we have too).

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/GradleDistribution.java
donat: Removed.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/GradleDistribution.java
donat: Added `@noimplement` to thge public types.

donat:
All review items have been resolved.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/GradleDistribution.java
```diff
@@ -0,0 +1,64 @@
...
+import java.net.URI;
+
+/**
+ * Represents a Gradle distribution.
+ *
+ * @author Donat Csikos
+ * @since 3.0
+ * @noimplement this interface is not intended to be implemented by clients
+ */
+public abstract class GradleDistribution {
```
oehme: We should have another chat about this class on Monday. Right now it is basically useless for our end users, so we might es well hide it completely (like ProjectConnection does).

On the other hand, if we want to allow users to build their own configuration UI for instance, we should fully expose the different subtypes.

PR #759, opened 2018-08-29
BuildConfigurationBuilder can configure all attributes

donat:


org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/configuration/BuildConfiguration.java
```diff
@@ -60,6 +68,68 @@ public File getRootProjectDirectory() {
...
     }
 
+    /**
+     * Returns Gradle user home for this configuration.
+     * <p>
+     * If {@link #isOverrideWorkspaceConfiguration()} returns true then the Gradle user home from
+     * this object is returned. Otherwise, the Gradle user home for the workspace configuration is
+     * returned.
+     * <p>
+     * If no Gradle user home is specified then this method returns {@code null}.
```
oehme: ❌ We should decide whether our public API uses null or Optional for optional return values. A mix is confusing.

donat: Fixed.

donat:

PR #761, opened 2018-08-30
Merge GradleDistribution declaration and usages

donat:


org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/GradleDistributions.java
```diff
@@ -0,0 +1,67 @@
...
+import java.net.URI;
+
+/**
+ * Contains factory methods for all supported Gradle distribution types.
+ *
+ * @author Donat Csikos
+ * @since 3.0
+ * @noimplement this interface is not intended to be implemented by clients
+ */
+public abstract class GradleDistributions {
```
oehme: ❌ Not needed, can be part of `GradleDistribution`

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/GradleDistribution.java
```diff
@@ -8,57 +8,22 @@
...
-     * Creates a reference to a specific version of Gradle. The appropriate distribution is
-     * downloaded and installed into the user's Gradle home directory.
-     *
-     * @param version the Gradle version to use
-     * @return a new distribution instance
-     */
-    public static GradleDistribution forVersion(String version) {
-        return org.eclipse.buildship.core.internal.util.gradle.GradleDistribution.forVersion(version);
-    }
+public interface GradleDistribution {
```
oehme: ❌ Should be an abstract class with 4 nested concrete classes, a.k.a a sealed class hierarchy.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/configuration/GradleArguments.java
```diff
@@ -93,7 +98,17 @@ private String toNonEmpty(List<String> stringValues, String defaultMessage) {
     public void applyTo(GradleConnector connector) {
         connector.forProjectDirectory(this.rootDir);
         connector.useGradleUserHomeDir(this.gradleUserHome);
-        this.gradleDistribution.apply(connector);
+        if (this.gradleDistribution instanceof LocalGradleDistribution) {
```
oehme: ❌ `GradleDistribution` should have `applyTo(Connector)`

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/GradleDistributionInfo.java
```diff
@@ -18,27 +18,64 @@
...
 import org.eclipse.buildship.core.internal.i18n.CoreMessages;
 import org.eclipse.buildship.core.internal.util.binding.Validator;
 
 /**
- * Describes a valid or invalid {@link GradleDistribution}.
+ * Describes a valid or invalid {@link BaseGradleDistribution}.
  *
  * @author Donat Csikos
  */
 public final class GradleDistributionInfo {
```
oehme: ⭕️ Why do we still have this class, now that the distribution types expose their information?

donat: `GradleDistributionInfo` can represent invalid states. It's needed for example when the user configures an invalid URL on the preferences pages. Besides the info object is there to serialize and deserialize distributions.

oehme: Got it, thanks! A common term for this is `ViewModel`, maybe call it like that?

donat: `GradleDistributionViewModel`?

oehme: Yeah, exactly

oehme: And it can live in the UI plugin

donat: Well, the serialization part is used in the core plugin to persist build configurations.

oehme: Shouldn't that be part of the `GradleDistribution` class (or BuildConfiguration)?

donat: That might make sense. The serialization is used in different configuration (workspace/build/run). Besides, a few dependent classes are in `core` but they should be in `ui`. I'll see what can I do about that.

oehme: Since we need to keep the serialized form backwards compatible anyway, I think making it part of the API is fine. I.e. have a `toString()` and `fromString()`.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/GradleDistributions.java
donat: Fixed.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/GradleDistribution.java
donat: Fixed.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/configuration/GradleArguments.java
donat: Isn't it a problem if we expose those methods? I aimed for a minimal API.

donat: Fixed

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/GradleDistributionInfo.java
donat: I've added `fromString()` and `toString()` to the pubic API, renamed `GradleDistributioninfo` to `GradleDistributionViewModel` and moved it to the `ui` plugin.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/GradleDistribution.java
```diff
@@ -81,4 +84,50 @@ public static FixedVersionGradleDistribution forVersion(String version) {
...
+            return forLocalInstallation(new File(configuration));
+        }
+
+        String remoteDistributionPrefix = "GRADLE_DISTRIBUTION(REMOTE_DISTRIBUTION(";
+        if (distributionString.startsWith(remoteDistributionPrefix) && distributionString.endsWith("))")) {
+            String configuration = distributionString.substring(remoteDistributionPrefix.length(), distributionString.length() - 2);
+            return forRemoteDistribution(createUrl(configuration));
+        }
+
+        String versionPrefix = "GRADLE_DISTRIBUTION(VERSION(";
```
oehme: These prefixes should be constants in the corresponding subclasses.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/GradleDistribution.java
```diff
@@ -81,4 +84,50 @@ public static FixedVersionGradleDistribution forVersion(String version) {
...
+    /**
+     * Deserializes a distribution.
+     * <p>
+     * This method recognizes strings returned by {@link #toString()}.
+     *
+     * @param distributionString the string representation of a distribution
+     * @return the parsed distribution
+     * @throws RuntimeException if the deserialization fails
+     */
+    public static GradleDistribution fromString(String distributionString) {
```
oehme: I think our general code style is to have static members first.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/LocalGradleDistribution.java
```diff
@@ -32,11 +31,23 @@
...
-        this.location = Preconditions.checkNotNull(location);
-        Optional<String> error = GradleDistributionInfo.validateLocalInstallationLocation(location);
+        this.location = Preconditions.checkNotNull(location).getAbsoluteFile();
+        Optional<String> error = validate(location);
         Preconditions.checkArgument(!error.isPresent(), error.or(""));
     }
 
+    private static  Optional<String> validate(File location) {
+        if (location == null) {
+            return Optional.of(NLS.bind(CoreMessages.ErrorMessage_0_MustBeSpecified, CoreMessages.GradleDistribution_Label_LocalInstallationDirectory));
```
oehme: Just throw here instead of doing an Optional dance :)

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/LocalGradleDistribution.java
```diff
@@ -32,11 +31,23 @@
...
+        } else if (!location.exists()) {
```
oehme: ❌ We shouldn't be doing IO here. This is a value object and shouldn't care about whether this file actually exists when it is constructed. This fact can change over time.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/LocalGradleDistribution.java
oehme: We don't need localized messages here, these are programming errors.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/LocalGradleDistribution.java
```diff
@@ -32,11 +31,23 @@
     private final File location;
 
     LocalGradleDistribution(File location) {
-        this.location = Preconditions.checkNotNull(location);
-        Optional<String> error = GradleDistributionInfo.validateLocalInstallationLocation(location);
+        this.location = Preconditions.checkNotNull(location).getAbsoluteFile();
```
oehme: Calling `getAbsoluteFile` does't really make sense this deep down in the code. If we can't deal with relative files, we should just assert that it's absolute. But actually I think we can deal with relative files just fine and shouldn't do anything surprising here.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/GradleDistribution.java
donat: Rearranged.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/LocalGradleDistribution.java
donat: I removed the whole validate method.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/LocalGradleDistribution.java
donat: Removed.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/LocalGradleDistribution.java
donat: Changed to make it work with relative paths.

donat:
Thank you for the thorough review @oehme!

PR #763, opened 2018-09-11
Describe different Gradle distribution types with inheritance

donat:


donat:
@oehme I guess we should simplify this even more by merging the subinterfaces and the concrete classes into public final classes.

Issue #752, opened 2018-08-28
Support debug Gradle test


Camork:
We can only lanuch a gradle tester in 'run' mode now, is there any plan to support debug it? Also there should be able to set the run configuration to customize the jvm args.

donat:
Yes, we have the intention to support Gradle invocations in debug mode. To give you a short overview, the plan is to first finish 3.0 with the new stable APIs, then implement task execution upon synchronization, and then comes the debug support. Prioritization might change over time, but the main indicator we use what should come next is the number of 👍 on the issues.

FYI, you can already set arguments and JVM arguments in the run configurations dialog.

donat:
Duplicate of https://github.com/eclipse/buildship/issues/249.

PR #765, opened 2018-09-17
Use public sync API in production code

donat:
The last missing piece for #752. The PR changes Buildship such that all production code executes project synchronization via the public `GradleWorkspace` service.


donat:
@marcphilipp It seems like I can't request a review from you if you are not part of the `eclipse` organization. Let's see if it changes after I mention you :)

donat:
It does not work. Let's figure something together tomorrow.

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/internal/wizard/project/ProjectImportConfiguration.java
```diff
@@ -62,7 +62,7 @@ public void setProjectDir(File projectDir) {
     }
 
 
-    public Property<Boolean> getOverwriteWorkspaceSettings() {
+    public Property<Boolean> overrideWorkspaceConfiguration() {
```
marcphilipp: I think this should be called `getOverrideWorkspaceConfiguration()`

donat: Fixed.

donat:
I was thinking about removing the interfaces but decided to keep it simple. But since you are also noticed it, I've removed the `GradleBuilds` and replaced the only usage with `Set<GradleBuild>`.

PR #775, opened 2018-09-24
Introduce GradleBuild.withConnection()

donat:
This commit adds a basic API to access Gradle via Tooling API. The
implementation doesn't do anything special yet, as the configuration of
the operations will be done in subsequent stories.

### DIFF
--- a/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/GradleBuild.java
+++ b/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/GradleBuild.java
@@ -8,6 +8,10 @@
 
 package org.eclipse.buildship.core;
 
+import java.util.function.Function;
+
+import org.gradle.tooling.ProjectConnection;
+
 import org.eclipse.core.runtime.IProgressMonitor;
 
 /**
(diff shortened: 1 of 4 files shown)


org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/GradleBuild.java
```diff
@@ -55,4 +59,21 @@
...
+     * TODO (donat) document exception handling
+     * TODO (donat) document scheduling rules
+     * TODO (donat) document how to load custom models (here or link to external documentation)
+     *
+     * @param action the action to execute
+     * @param monitor the monitor to report progress on, or {@code null} if progress reporting is not desired
+     * @return the result of the action
+     * @throws Exception when the action fails
+     */
+    <T> T withConnection(Function<ProjectConnection, ? extends T> action, IProgressMonitor monitor) throws Exception;
```
marcphilipp: This might be a personal preference, but I prefer "lambda parameters" as the last parameter.

```java
build.withConnection(connection -> {
    ...
}, monitor);
```

vs.

```java
build.withConnection(monitor, connection -> {
    ...
});
```

WDYT?

marcphilipp: Should we explicitly give the `action` access to the `monitor` by using a `BiFunction` or would we recommend just passing it the same `IProgressMonitor` instance?

```java
IProgressMonitor monitor = ...;
build.withConnection(connection -> {
    for (...) {
        monitor.worked(1);
    }
}, monitor);
```

Moreover, who's responsible for calling `monitor.done()`? The `action` or the code that executes the `action`?

donat: Re: argument order: the order is the same for most API methods in Eclipse. Here, I prefer consistency.

donat: Re: monitor object used by action: I think it's cleaner to use a simple `Function` and recommend the users to reuse `monitor` if needed within the action. Besides, we don't have to deal with calling `monitor.done()`. From the javadoc: _callers that pass a monitor are no longer allowed to rely upon the callee invoking done() and must do so themselves if necessary_.

donat:
Thanks, @marcphilipp, that was fast. ⚡️

PR #782, opened 2018-10-08
Describe `withConnection` use-cases in javadoc

donat:


org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/GradleBuild.java
```diff
@@ -70,9 +70,59 @@
...
+     * <p>
+     * <pre><code>
+     *     IProject project = ResourcesPlugin.getWorkspace().getRoot().getProject("project-name");
+     *     GradleBuild build = GradleCore.getWorkspace().getBuild(project).get();
+     *
+     *     GradleProject model = build.withConnection(connection -> connection.getModel(GradleProject.class), monitor);
+     *     System.out.println(model.getBuildDirectory());
+     * </code></pre>
+     * Clients can load the <a href="https://docs.gradle.org/current/javadoc/org/gradle/tooling/ProjectConnection.html#model-java.lang.Class-"> default models</a>.
+     * Also, custom model loading possible with using {@code --init-script}:
```
oehme: You can load custom models provided by Gradle plugins or even inject your own with an init script:

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/GradleBuild.java
```diff
@@ -70,9 +70,59 @@
...
+     * <pre><code>
+     *     connection.model(CustomModel.class).withArguments("--init-script", "/path/to/init-script/with/model").get();
+     * </code></pre>
+     * An example project with custom model loading is available at <a href="https://github.com/donat/buildship-custom-tapi-model">https://github.com/donat/buildship-custom-tapi-model</a>.
```
oehme: This should live in the gradle organization

oehme: Some nicer names in that sample would be great. "CustomModel" and "DefaultModel" etc. don't say very much.

oehme: Also, the model implementation belongs to the plugin, not the model project. The client only depends on the model interface.

donat: Should I request a new public repo? Or maybe we could store it within a new `samples` folder ?

oehme: Samples in Buildship sounds like a good idea.

donat: > Also, the model implementation belongs to the plugin, not the model project. The client only depends on the model interface.

That's correct, I just wanted to reuse as much from https://github.com/bmuschko/tooling-api-custom-model as possible. If you think it's important I can extract the interface from the `model` project and have only that included in the plugin.

oehme: >  I can extract the interface from the model project and have only that included in the plugin.

You mean the implementation? Yes, I think that's important, because clients should not know about the implementation.

donat: - Sample project moved in `samples/custom-tapi-model` directory.
- Custom model renamed to `PluginApplication`, implementation class to `PluginApplicationImpl`
- `PluginApplicationImpl` is moved from `model` to `plugin`.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/GradleBuild.java
donat: Fixed.

PR #787, opened 2018-10-12
Declare project configurator extension pont

donat:
This PR adds a simple extension point to hook into the Buildship project synchronization. The extension point requires a simple `ProjectConfigurator` implementation and has no feature for the ordering.


org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/workspace/SynchronizeGradleBuildOperation.java
```diff
@@ -240,11 +249,74 @@ private IProject addNewEclipseProjectToWorkspace(EclipseProject project, SubMoni
...
+            }
+        };
+
+        SubMonitor progress = SubMonitor.convert(monitor);
+        progress.setWorkRemaining(this.configurators.size());
+        for (ProjectConfigurator c : this.configurators) {
+            try {
+                c.init(context, progress.newChild(1));
+            } catch (Exception e) {
+                CorePlugin.logger().warn("Project configurator " + c.getClass().getName() + " failed to initialize" , e);
```
oehme: Shouldn't we use the logger associated with the plugin that contributed the configurator? Otherwise bugs will be submitted to our bug tracker instead of theirs.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/workspace/SynchronizeGradleBuildOperation.java
```diff
@@ -240,11 +249,74 @@ private IProject addNewEclipseProjectToWorkspace(EclipseProject project, SubMoni
...
+            }
+        };
+
+        SubMonitor progress = SubMonitor.convert(monitor);
+        progress.setWorkRemaining(this.configurators.size());
+        for (ProjectConfigurator c : this.configurators) {
+            try {
+                c.configure(context, progress.newChild(1));
+            } catch (Exception e) {
+                CorePlugin.logger().warn("Project configurator " + c.getClass().getName() + " configuration failed" , e);
```
oehme: Same as above.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/workspace/SynchronizeGradleBuildOperation.java
oehme: We'll need some user-visible indication, e.g. an error marker on the project or similar.

donat: Isn't it enough if we log the problem? I'm not really sure if we want to use error markers for this. In the end, it's the contributors' responsibility to provide exception handling imo.

oehme: If we get an exception here it's because the contributor didn't handle exceptions correctly. We don't want to just silently pretend that everything is alright. Users won't look in the warning log, they'll just wonder why some things don't quite work as expected.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/ProjectConfigurator.java
```diff
@@ -12,8 +12,28 @@
 
 /**
  * Interface to hook into Buildship's project synchronization.
+ *
+ * <p>
+ * The primary intent of this interface is to let external plugins set up project configuration for
+ * tools that are unknown by Buildship. An implementation can be registered via the
+ * {@code projectconfigurators} extension point. An implementation is not limited to use any APIs
```
oehme: An implementation is free to use any APIs to change the project. It is recommended that configurators only add additional configuration and not remove anything previously added (e.g. don't delete projects or dependencies)

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/workspace/ProjectConfigurators.java
```diff
@@ -0,0 +1,132 @@
...
+                contribution.getContributorPluginId(),
+                operationName);
+        Status status = new Status(IStatus.WARNING, contribution.getContributorPluginId(), message, exception);
+        CorePlugin.getInstance().getLog().log(status);
+
+        try {
+            IResource targetLocation = markerLocation();
+            GradleErrorMarker.create(targetLocation, this.gradleBuild, message, exception, 0);
+        } catch (CoreException e) {
+            CorePlugin.getInstance().getLog().log(e.getStatus());
```
oehme: ❌ This is still logging using our logger, thus AERI reports will be sent to Buildship. We should log using the logger of the faulty configurator.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/ProjectConfigurator.java
donat: Thanks for the proposal, I'll adjust the javadoc before merging.

PR #795, opened 2018-11-08
Prevent executing multiple synchronization upon import

donat:


donat:
The idea is to detect if a synchronization is running on another thread regardless of how the `GradleBuild` objects are created. So if there are two instances representing the same build and one is synchronizing then the calling `isSynchronizing()` on the other object will still return `true`.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/DefaultGradleBuild.java
```diff
@@ -58,16 +63,24 @@ public SynchronizationResult synchronize(IProgressMonitor monitor) {
     }
 
     public SynchronizationResult synchronize(NewProjectHandler newProjectHandler, CancellationTokenSource tokenSource, IProgressMonitor monitor) {
+        synchronizingBuilds.add(this);
```
marcphilipp: Is it possible that this method is called from different threads, e.g. when synchronization is requested synchronously after the job has just been scheduled? If so, should check the result of calling `add()` and abort the synchronization if one is already in progress?

donat: It's can happen but the other way around. The synchronization is scheduled synchronously, which triggers Eclipse JDT to initialize the classpath container with a new sync job.
It makes sense, although I just realized that the proper solution might be to implement scheduling rules. I'm going to look into that tomorrow.

donat: The scheduling rule solution was a big dead-end. I had the assumption that I can amend existing scheduling rules which turned out to be not true as there's no way to extend the workspace root rule with new functionality (`GradleBuildRule`).

I managed to solve the situation with your proposal: the synchronize operations are skipped within `DefaultGradleBuild` if there's already running on a different thread.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/DefaultGradleBuild.java
```diff
@@ -59,15 +64,29 @@ public SynchronizationResult synchronize(IProgressMonitor monitor) {
...
         monitor = monitor != null ? monitor : new NullProgressMonitor();
-        SynchronizeOperation operation = new SynchronizeOperation(newProjectHandler);
+
+        SynchronizeOperation runningOperation = syncOperations.get(this);
+        if (runningOperation != null && (newProjectHandler == NewProjectHandler.NO_OP || Objects.equals(newProjectHandler, runningOperation.newProjectHandler))) {
+            return newSynchronizationResult(Status.OK_STATUS);
+        }
+
+        SynchronizeOperation operation = new SynchronizeOperation(this.gradleBuild, newProjectHandler);
+        syncOperations.put(this, operation);
```
marcphilipp: I think there's still a race condition here: if this method is called from multiple threads, both could get past the initial check and to this line. Instead, I'd call `putIfAbsent` and check if the return value is `null` in line 68.

donat: Good catch! I've adjusted the code.

Issue #786, opened 2018-10-08
Extract Java project configurator


donat:

moley:
Hi donat, so as we talked about in eclipsecon, please keep in mind that when an error (like this wtp thing) comes along, please do not break reading/configuring project, because this leeds to an incomplete project setup, but only add an error marker. 
Thanks a lot

PR #798, opened 2018-11-13
Extract project synchronization to separate project configurator

donat:


donat:
@oehme I think we should not extract create another [configurator for Java projects](#786) as it would only introduce additional complexity. WDYT?

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/internal/workspace/ImportingWtpProjects.groovy
```diff
@@ -62,17 +61,17 @@ class ImportingWtpProjects extends ProjectSynchronizationSpecification {
...
 
         when:
         SynchronizationResult result = tryImportAndWait(root)
+        IProject project = findProject('project')
 
         then:
-        result.status.severity == IStatus.WARNING
-        result.status.exception instanceof UnsupportedConfigurationException
-        result.status.exception.message == "WTP currently does not support mixed deployment paths."
+        result.status.severity == IStatus.OK
```
oehme: ❌ Shouldn't be OK if there were errors.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/workspace/BaseConfigurator.java
```diff
@@ -0,0 +1,128 @@
...
+        progress.worked(1);
+
+        BuildScriptLocationUpdater.update(model, persistentModel, progress.newChild(1));
+
+        LinkedResourcesUpdater.update(project, ImmutableList.copyOf(model.getLinkedResources()), persistentModel, progress.newChild(1));
+        GradleFolderUpdater.update(project, model, persistentModel, progress.newChild(1));
+        ProjectNatureUpdater.update(project,  ImmutableList.copyOf(model.getProjectNatures()), persistentModel, progress.newChild(1));
+        BuildCommandUpdater.update(project, ImmutableList.copyOf(model.getBuildCommands()), persistentModel, progress.newChild(1));
+
+        if (isJavaProject(model)) {
```
oehme: ⭕️ Should probably be in its own Java configurator.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/workspace/BaseConfigurator.java
```diff
@@ -0,0 +1,128 @@
...
+        //old Gradle versions did not expose natures, so we need to add the Java nature explicitly
+        CorePlugin.workspaceOperations().addNature(project, JavaCore.NATURE_ID, progress.newChild(1));
+        IJavaProject javaProject = JavaCore.create(project);
+        OutputLocationUpdater.update(javaProject, model.getOutputLocation(), progress.newChild(1));
+        SourceFolderUpdater.update(javaProject, ImmutableList.copyOf(model.getSourceDirectories()), progress.newChild(1));
+        LibraryFilter.update(javaProject, model, progress.newChild(1));
+        ClasspathContainerUpdater.update(javaProject, model, progress.newChild(1));
+        JavaSourceSettingsUpdater.update(javaProject, model, progress.newChild(1));
+        GradleClasspathContainerUpdater.updateFromModel(javaProject, model, this.locationToProject.values(), persistentModel, progress.newChild(1));
+        WtpClasspathUpdater.update(javaProject, model, this.internalGradleBuild, progress.newChild(1));
```
oehme: ⭕️ Should probably be in its own wtp configurator, just to illustrate the point of configurators.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/workspace/ModelProviderUtil.java
```diff
@@ -0,0 +1,38 @@
...
+import org.eclipse.core.runtime.IProgressMonitor;
+
+import org.eclipse.buildship.core.internal.util.gradle.HierarchicalElementUtils;
+
+public class ModelProviderUtil {
+
+    /**
+     * @return a flattened list of Eclipse projects for all projects included in the Gradle build
+     */
+    public static Set<EclipseProject> fetchAllEclipseProjects(GradleBuild build, CancellationTokenSource tokenSource, FetchStrategy fetchStrategy, IProgressMonitor monitor) {
```
oehme: ⭕️ I think we shouldn't use a different API than our users are using. If we think that caching is important, it should be part of the official API.

A possible solution: Decorate the `withConnection` method when invoked from project synchronization so that it automatically caches models that were requested during that synchronization before. This may help reduce the number of TAPI calls.

An even more efficient alternative would be an API where configurators can tell us what models they are interested in and we fetch them all in one go, including the ones we need.

This is not something we need to solve in this PR, but I wanted to point it out, because the current API encourages inefficiency (configuring the build many times by calling `withConnection` in different configurators).

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/internal/workspace/ImportingWtpProjects.groovy
donat: How should we solve this? This change is there because configurators are now responsible for handling their own errors. I guess we should introduce a `ProjectConfiguratorResult` and use it to compute the actual result. WDYT?

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/workspace/BaseConfigurator.java
donat: I'll add a TODO for this and implement it as part of #786.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/workspace/BaseConfigurator.java
donat: 👍

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/workspace/ModelProviderUtil.java
donat: https://github.com/eclipse/buildship/issues/800

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/internal/workspace/ImportingWtpProjects.groovy
oehme: The context objects we pass to the configurators should have an error-handling method. It should take a message and optionally a throwable and 

- add a Gradle error marker at the most specific location
- add the problem to the aggregate Status we return at the end of synchronization

oehme: We don't want a return object, because we don't want contributors to instantiate stuff themselves. Also we may want additional stuff returned in the future. Using a single context object for all interactions is the most extensible way to write an SPI.

donat: Makes sense, thanks for he elaboration!

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/InitializationContext.java
```diff
@@ -21,4 +21,13 @@
...
     GradleBuild getGradleBuild();
+
+    /**
+     * Registers a problem during initialization. Instead of implementing separate error handling,
+     * project configurators should use this method to report issues.
+     *
+     * @param message the error message describing the problem
+     * @param exception The exception to report; can be {@code null}
+     */
+    void onError(String message, Exception exception);
```
oehme: Should just be `error`. `onError` would be a name for a method that configures error handling.

oehme: Similarly we should probably also have `warning()`

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/ProjectContext.java
```diff
@@ -23,4 +23,13 @@
...
     IProject getProject();
+
+    /**
+     * Registers a problem during project synchronization. Instead of implementing separate error
+     * handling, project configurators should use this method to report issues.
+     *
+     * @param message the error message describing the problem
+     * @param exception The exception to report; can be {@code null}
+     */
+    void onError(String message, Exception exception);
```
oehme: Same as above, should share a common interface.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/DefaultGradleBuild.java
```diff
@@ -131,26 +161,59 @@ public boolean equals(Object obj) {
...
+            GradleMarkerManager.clear(this.gradleBuild);
+            DefaultSynchronizationResult result;
+            try {
+                CorePlugin.operationManager().run(this, tokenSource, monitor);
+                this.failures.forEach(f -> GradleErrorMarker.create(f.getResource(), this.gradleBuild, f.getMessage(), f.getException(), 0));
+                result = DefaultSynchronizationResult.from(getFailures());
+            } catch (CoreException e) {
+                ToolingApiStatus status = ToolingApiStatus.from("Project synchronization" , e);
+                if (status.severityMatches(IStatus.WARNING | IStatus.ERROR)) {
+                    GradleMarkerManager.addError(this.gradleBuild, status);
```
oehme: Should make a difference between warning and error. Or maybe the method name is just misleading?

donat: No, it actually adds a marker with error severity. I'll add a separate method for defining warning markers.

PR #801, opened 2018-11-13
Make new project wizard work with init task from Gradle 5.0+

donat:
Gradle 5.0 added extra required parameters to the init task to specify
attributes like the script language, package name, etc. If this info
is not specified then Gradle reads them from the standard input. This
made the new project wizard to hang as it provides no input.


donat:
The cross-version coverage would catch that use-case, but I agree, it still would be better to make it forward-compatible. I'll look into disabling the input stream.

donat:
I've disabled the input for the `init` task.

PR #805, opened 2018-11-20
[WIP] Define configurator model queries in extension point

donat:


donat:
@oehme instead of a `Function` maybe we should require implementing a`GradleBuildAction` interface
```
interface GradleBuildAction {
     Object execute(GradleBuildActionContext context) throws Exception;
}

public GradleBuildActionContext {
     GradleBuild getGradleBuild();
}
```

donat:
I think we should deliver this as part of Buildship 3.0. I don't want to do a release without fully finish the project configurator story. Caching seems the way to go. I can imagine it can be done transparently within one synchronization. I'll look into this.

PR #810, opened 2018-11-26
Add missing options to workspace and project configuration

donat:
This pull request lets the user configure the following properties in the workspace and build configuration:
 - Java home,
 - program arguments
 - JVM arguments
 - show console view
 - show executions view

To implement that, I've added the UI controls to the project preference page that is reused by the workspace, project, and run configuration dialogs. Consequently, I've removed arguments and java home tabs from the run configuration dialog as it provided duplicate functionality. 

I've also adjusted the project import as it now respects the `show console view` configuration.


oehme:
> To implement that, I've added the UI controls to the project preference page that is reused by the workspace, project, and run configuration dialogs. Consequently, I've removed arguments and java home tabs from the run configuration dialog as it provided duplicate functionality.

Some screenshots would be nice to make this clearer. Does this mean that I have to repeat all the configuration I did for the project, even if I only wanted to change the build arguments?

donat:
That is correct. Which makes me think that we should initialize the property pages such that they are pre-filled with the inherited settings. That way, changing a single argument will be much simpler.

oehme:
Pre-filling doesn't help when I later change the workspace/project settings. Are you sure it's best to override them all together? What if we did it more fine grained, as in 

```
JDK
  O Location
  x Use Project Setting
Gradle distribution
  O Location
  x Version
  O Wrapper
  O Use Project Setting
```

donat:
I thought about that, but it would make the UI a bit cluttered. For the JDK and the Gradle distribution it's fine, but how about the boolean flags like the offline mode of for the build scans enablement? Shall we have duplicated flags for that too?

donat:
Besides this level of granularity is fine if you ask me. Nobody complained that you can either inherit all or none of the preference so far.

oehme:
Well, we can always change it later if our users don't like it. So I'm fine with going forward with this solution.

PR #813, opened 2018-11-28
Cache Tooling API results during synchronisation

donat:


org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/CachingProjectConnection.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @Override
+    public <T> BuildActionExecuter<T> action(BuildAction<T> buildAction) {
+        return new CachingBuildActionExecuter<>(this.delegate.action(buildAction), buildAction, this.cache);
+    }
+
+    @Override
+    public Builder action() {
+        return this.delegate.action();
```
oehme: We should make these cacheable too.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/DefaultGradleBuild.java
```diff
@@ -201,13 +203,18 @@ SynchronizationResult run(CancellationTokenSource tokenSource, IProgressMonitor
...
+            try {
+                SubMonitor progress = SubMonitor.convert(monitor, 5);
+                progress.setTaskName((String.format("Synchronizing Gradle build at %s with workspace", this.gradleBuild.getBuildConfig().getRootProjectDirectory())));
+                new ImportRootProjectOperation(this.gradleBuild.getBuildConfig(), this.newProjectHandler).run(progress.newChild(1));
+                Set<EclipseProject> allProjects = ModelProviderUtil.fetchAllEclipseProjects(this.gradleBuild, tokenSource, FetchStrategy.FORCE_RELOAD, progress.newChild(1));
+                new ValidateProjectLocationOperation(allProjects).run(progress.newChild(1));
+                new RunOnImportTasksOperation(allProjects, this.gradleBuild.getBuildConfig()).run(progress.newChild(1), tokenSource);
+                this.failures = new SynchronizeGradleBuildOperation(allProjects, this.gradleBuild, this.newProjectHandler, ProjectConfigurators.create(this.gradleBuild, CorePlugin.extensionManager().loadConfigurators())).run(progress.newChild(1));
+            } finally {
+                this.gradleBuild.projectConnectionCache.clear();
```
oehme: ❌ Should not be stateful.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/DefaultGradleBuild.java
```diff
@@ -290,6 +297,10 @@ public GradleConnectionOperation(Function<ProjectConnection, ? extends T> action
         public void runInToolingApi(CancellationTokenSource tokenSource, IProgressMonitor monitor) throws Exception {
             // TODO (donat) use AutoCloseable once we update to Tooling API 5.0
             ProjectConnection connection = IdeAttachedProjectConnection.newInstance(tokenSource, getGradleArguments(), monitor);
+            if (isSynchronizing()) {
```
oehme: ❌ Should not be stateful.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/DefaultGradleBuild.java
```diff
@@ -157,11 +159,11 @@ public boolean equals(Object obj) {
      */
     private static class SynchronizeOperation extends BaseToolingApiOperation {
 
-        private final InternalGradleBuild gradleBuild;
+        private final DefaultGradleBuild gradleBuild;
         private final NewProjectHandler newProjectHandler;
         private List<SynchronizationProblem> failures;
 
-        public SynchronizeOperation(InternalGradleBuild gradleBuild, NewProjectHandler newProjectHandler) {
+        public SynchronizeOperation(DefaultGradleBuild gradleBuild, NewProjectHandler newProjectHandler) {
```
oehme: ❌ A CachingGradleBuild decorator should be used here, so it is naturally scoped to the sync operation instead of mutating state inside the GradleBuild object. The DefaultGradleBuild should have no caching logic.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/IdeAttachedProjectConnection.java
```diff
@@ -63,7 +63,6 @@ public TestLauncher newTestLauncher() {
     private <T extends LongRunningOperation> T configureOperation(T operation) {
         BuildEnvironment buildEnvironment = this.delegate.getModel(BuildEnvironment.class);
         this.gradleArguments.applyTo(operation, buildEnvironment);
-        this.gradleArguments.describe(this.progressAttributes, buildEnvironment);
```
oehme: Unrelated change?

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/ProjectConnectionCache.java
```diff
@@ -0,0 +1,36 @@
...
+
+    public Optional<Object> get(Object key) {
+        return Optional.ofNullable(this.cache.getIfPresent(key));
+    }
+
+    public void put(Object key, Object value) {
+        this.cache.put(key, value);
+    }
+
+    public Map<Object, Object> asMap() {
```
oehme: Dead code

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/ProjectConnectionCache.java
```diff
@@ -0,0 +1,36 @@
...
+
+public class ProjectConnectionCache {
+
+    private final Cache<Object, Object> cache = CacheBuilder.newBuilder().build();
+
+    public void clear() {
+        this.cache.invalidateAll();
+    }
+
+    public Optional<Object> get(Object key) {
```
oehme: We don't need optional here, we're just unwrapping it 1 line later again. This is an internal helper, not a public API.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/DefaultGradleBuild.java
donat: Do you mean that we should do use caching even if the project is not being synchronized? How can we tell that the results are up-to-date?

donat: Ah, I understand now. I should have read the comments below before commenting here.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/IdeAttachedProjectConnection.java
donat: It's unrelated but valid: I didn't notice before that at some point we started printing the connection properties into the background console. We only want this behavior for the task execution. This change fixes that problem.

PR #814, opened 2018-12-03
Allowed refresh gradle project with shortcut key

Camork:
As a high frequency action when developing gradle projects. It would be better if we have default shortcut key.


donat:
I don't think we should add hotkeys for any actions. We contribute Buildship to a list of Eclipse packages by default. A new hotkey can potentially mess up the setup of thousands of users who have custom key bindings. 

If you need to persist your custom key bindings, I recommend checking out [Oomph's Preference Recorder](https://www.eclipse.org/community/eclipse_newsletter/2014/november/article2.php) feature.

Issue #858, opened 2019-03-04
Dark Theme: fix Window > Preferences > Gradle > Advanced Options background color


eric-milles:
The "Advanced Options" group control has a background color set that is incompatible with the Dark Theme.

![image](https://user-images.githubusercontent.com/18193802/53753572-3142c300-3e77-11e9-9d98-0c0b16051269.png)

donat:
I thought we fixed this. Do you use the latest version?

eric-milles:
Still appears this way in 3.0.1.v20181217-1554.  This is what update site for Eclipse 4.10 provides.

donat:
Strange. I'll take a look.

donat:
@eric-milles On osx it seemed fine, at least for the first glance:

<img width="738" alt="Screenshot 2019-03-19 at 13 52 34" src="https://user-images.githubusercontent.com/419883/54607399-87eef600-4a4e-11e9-848d-84a1c583dcec.png">

I've just noticed that the `Advanced Options` label has the wrong color. My guess is that this is the bug. I'll provide a patch soon.

PR #862, opened 2019-03-19
Fix dark theme for preferences UI

donat:
After adding UI for all Gradle configuration options we forgot to
adjust the class names in the dark theme css file.

Resolves https://github.com/eclipse/buildship/issues/858


donat:
Before
Screenshot 2019-03-19 at 13 52 34

After
image

PR #866, opened 2019-03-20
Workspace composites

kuzniarz:
Reopened pull request due to missing signed-off-footer. Implemented workspace composite wizard and preference UIs.


donat:
Can you please do a rebase against the `workspace-composites` branch so that I don't see my commits:

image

donat:
@kuzniarz I have yet another request. The diff page is quite messed up because we use a different OS for development. For example, the `plugin.xml` file shows complete replacement. 

Could you please adjust the edited files to use `LF` for line endings instead of the Windows-style `CRLF`? Also, please adjust your Eclipse workspace to do the same to avoid future problems.

PR #871, opened 2019-04-08
Run tasks upon synchronization and on auto-build

donat:
This PR makes use the synchronization task declaration in the  `EclipseProject` model introduced in Gradle 5.4. Additionally, I've implemented the same thing for auto-build tasks too.

To make things a bit cleaner, I've moved the compatibility model injection from the `ModelProvider` `ProjectConnection` level. 

It's a WIP as w should merge it only with Gradle 5.4-rc-1 at least.


org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/CachingIntermediateResultHandler.java
```diff
@@ -0,0 +1,37 @@
...
+        this.buildAction = buildAction;
+        this.delegate = handler;
+        this.cache = cache;
+    }
+
+    @Override
+    public void onComplete(T result) {
+        if (result != null) {
+            CacheKey cacheKey = new CacheKey();
+            cacheKey.setBuildAction(this.buildAction);
```
oehme: ❌ cache keys should never be mutable

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/DefaultGradleBuild.java
```diff
@@ -214,11 +218,17 @@ public void runInToolingApi(CancellationTokenSource tokenSource, IProgressMonito
...
+                Set<EclipseProject> allProjects = collectAll(rootProjects);
                 new ValidateProjectLocationOperation(allProjects).run(progress.newChild(1));
                 new RunOnImportTasksOperation(allProjects, this.gradleBuild.getBuildConfig()).run(progress.newChild(1), tokenSource);
                 this.failures = new SynchronizeGradleBuildOperation(allProjects, this.gradleBuild, this.newProjectHandler,
                         ProjectConfigurators.create(this.gradleBuild, CorePlugin.extensionManager().loadConfigurators())).run(progress.newChild(1));
+
+                // The root projects are no longer loaded via GradleBuild.getModelProvider(). To avoid unnecessary
+                // model loads (e.g. when somebody opens the test sources from the Execution view) we add the result
+                // to the cache here.
+                ((DefaultModelProvider)this.gradleBuild.getModelProvider()).getCache().put(EclipseProject.class, rootProjects);
```
oehme: ❌ Why do we have to do this here, when you've added the caching to the phased actions too?

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/util/gradle/IdeFriendlyClassLoading.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    private static final ClassLoader getIdeFriendlyClassLoader() throws IOException {
+        ClassLoader coreClassloader = DefaultModelProvider.class.getClassLoader();
+        ClassLoader tapiClassloader = ProjectConnection.class.getClassLoader();
+        URL actionRootUrl = FileLocator.resolve(coreClassloader.getResource(""));
+        if (ideFriendlyClassLoader == null) {
+            ideFriendlyClassLoader = new URLClassLoader(new URL[] { actionRootUrl }, tapiClassloader);
+        }
+        return ideFriendlyClassLoader;
```
oehme: ❌ Needs to be closed after use

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/workspace/ModelProviderUtil.java
```diff
@@ -9,30 +9,61 @@
...
+            } else if (supportsCompositeBuilds(gradleVersion)) {
+                return queryCompositeEclipseModel(connection);
+            } else {
+                return queryEclipseModel(connection);
+            }
+        }, monitor);
+
+    }
+
+    private static boolean supportsTaskExecution(GradleVersion gradleVersion) {
```
oehme: ⭕️ You mean supportsPhasedActions?

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/workspace/WtpConfigurator.java
```diff
@@ -61,9 +62,14 @@
     public void init(InitializationContext context, IProgressMonitor monitor) {
         // TODO (donat) add required model declarations to the project configurator extension point
         GradleBuild gradleBuild = context.getGradleBuild();
-        this.gradleBuild = (DefaultGradleBuild) gradleBuild;
-        Collection<EclipseProject> eclipseProjects = ModelProviderUtil.fetchAllEclipseProjects(this.gradleBuild, GradleConnector.newCancellationTokenSource(), FetchStrategy.LOAD_IF_NOT_CACHED, monitor);
-        this.locationToProject = eclipseProjects.stream().collect(Collectors.toMap(p -> p.getProjectDirectory(), p -> p));
+        try {
```
oehme: ⭕️ Why this change?

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/CachingIntermediateResultHandler.java
donat: It's been like this in the code, I've followed the existing pattern.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/DefaultGradleBuild.java
donat: Yeah, now we have two types of model caches. One that kicks in only during synchronization and happens automatically and another one that is long-lived (i.e. `ModelProvider`). Because we moved away from `ModelProvider` usage for loading the Eclipse model, other code pieces like the `Open Test Sources` functionality would have to load the same Eclipse model as it was retrieved during the sync.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/util/gradle/IdeFriendlyClassLoading.java
donat: It's not a new thing, I only renamed the class. But I can fix it of course.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/workspace/ModelProviderUtil.java
donat: No, phased actions and task execution were introduced in Gradle 4.8, and 5.4 respectively. For Gradle <=4.8 <5.4 versions the model builder is not available for the `RunEclipseSynchronizationTasks` model.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/CachingIntermediateResultHandler.java
oehme: Let's fix it then, keys should always be immutable.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/workspace/WtpConfigurator.java
donat: It seemed better to use a public API to load models in all configurators. Besides, it hid the problem with model caching: the original code filled the `ModelProvider` cache by accident.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/workspace/ModelProviderUtil.java
oehme: I just want to avoid confusion with "being able to execute tasks", which the TAPI could do from day1.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/DefaultGradleBuild.java
oehme: Any way we could unify them and have everything be long-lived? I really don't like exposing the internals here like this.

donat: I don't think that's feasible, at least not with a lot of effort. Alternatively, we could handle the caching with implementing a `ModelProvider.queryEclipseModelAndExecuteTasks()` method. 

`ModelProvider` and `GradleBuild.withConnection` has duplicate functionality, that's the real issue here.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/CachingIntermediateResultHandler.java
donat: Done.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/util/gradle/IdeFriendlyClassLoading.java
donat: Done.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/workspace/ModelProviderUtil.java
donat: Done.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/DefaultGradleBuild.java
oehme: If it would be a lot of effort then I'd be in favor of a new method instead.

donat: I've moved the query to the `ModelProvide` interface.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/util/gradle/IdeFriendlyClassLoading.java
donat: Closing the class loader immediately caused problems in the tests: the class loader is needed on the daemon side when accessing the remote classes.

I've moved the cleanup code to the end of the synchronization.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/DefaultGradleBuild.java
```diff
@@ -213,24 +217,18 @@ SynchronizationResult run(CancellationTokenSource tokenSource, IProgressMonitor
...
                 this.failures = new SynchronizeGradleBuildOperation(allProjects, this.gradleBuild, this.newProjectHandler,
                         ProjectConfigurators.create(this.gradleBuild, CorePlugin.extensionManager().loadConfigurators())).run(progress.newChild(1));
-
-                // The root projects are no longer loaded via GradleBuild.getModelProvider(). To avoid unnecessary
-                // model loads (e.g. when somebody opens the test sources from the Execution view) we add the result
-                // to the cache here.
-                ((DefaultModelProvider)this.gradleBuild.getModelProvider()).getCache().put(EclipseProject.class, rootProjects);
             } finally {
                 this.gradleBuild.projectConnectionCache.invalidateAll();
+                IdeFriendlyClassLoading.cleanup();
```
oehme: I feel like this is the wrong place for this - Shouldn't this be done when our plugin is unloaded instead?

donat: Let's do that.

PR #872, opened 2019-04-09
Use package import for SLF4J

donat:
Having an explicit dependency to the `org.slf4j.api` bundle can be restricting as it might pull in the dependency if slf4j is packaged in a different way (see the [corresponding m2e bug](https://bugs.eclipse.org/bugs/show_bug.cgi?id=506473)).  To leverage that, this PR replaces the bundle dependency with a package import.

Our build scripts are not prepared for resolving this type of dependencies so I added an explicit dependency to the `build.gradle` files.

### DIFF
--- a/org.eclipse.buildship.core/build.gradle
+++ b/org.eclipse.buildship.core/build.gradle
@@ -5,3 +5,7 @@ tasks.processResources.doFirst {
     def buildshipRunimeProperties = file('src/main/resources/org/eclipse/buildship/core/internal/gradle/buildship-runtime.properties')
     buildshipRunimeProperties.text = "ToolingApiVersion=$toolingApiVersion"
 }
+
+dependencies {
+    compile 'org.slf4j:slf4j-api:1.7.2'
+}
--- a/org.eclipse.buildship.kotlin/build.gradle
+++ b/org.eclipse.buildship.kotlin/build.gradle
@@ -1,6 +1,9 @@
 apply plugin: eclipsebuild.BundlePlugin
 
-dependencies { compile 'org.gradle:gradle-kotlin-dsl-tooling-models:+' }
+dependencies {
+    compile 'org.gradle:gradle-kotlin-dsl-tooling-models:+'
+    compile 'org.slf4j:slf4j-api:1.7.2'
+}
 
 task updateKotlinModelJar(type: Copy) {
     into '.'
--- a/org.eclipse.buildship.core/META-INF/MANIFEST.MF
+++ b/org.eclipse.buildship.core/META-INF/MANIFEST.MF
@@ -16,11 +16,11 @@ Require-Bundle: org.eclipse.core.expressions,
  org.eclipse.jdt.junit.core,
  org.eclipse.jdt.launching,
  org.eclipse.debug.core,
- org.slf4j.api;bundle-version="1.7.2",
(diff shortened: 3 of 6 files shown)


org.eclipse.buildship.core/build.gradle
```diff
@@ -4,4 +4,8 @@ apply plugin: eclipsebuild.BundlePlugin
 tasks.processResources.doFirst {
     def buildshipRunimeProperties = file('src/main/resources/org/eclipse/buildship/core/internal/gradle/buildship-runtime.properties')
     buildshipRunimeProperties.text = "ToolingApiVersion=$toolingApiVersion"
+}
+
+dependencies {
+    compile 'org.slf4j:slf4j-api:1.7.2'
```
big-guy: should this be implementation or api?

org.eclipse.buildship.kotlin/build.gradle
```diff
@@ -1,6 +1,9 @@
 apply plugin: eclipsebuild.BundlePlugin
 
-dependencies { compile 'org.gradle:gradle-kotlin-dsl-tooling-models:+' }
+dependencies {
+    compile 'org.gradle:gradle-kotlin-dsl-tooling-models:+'
```
big-guy: Should this be limited to a version range? Is this still necessary with the repackaging of Kotlin DSL inside Gradle?

org.eclipse.buildship.kotlin/build.gradle
```diff
@@ -1,6 +1,9 @@
...
+    compile 'org.slf4j:slf4j-api:1.7.2'
```
big-guy: same as above (api or impl?)

org.eclipse.buildship.core/build.gradle
donat: The same as the tooling-api requires: slf4j-api.

org.eclipse.buildship.kotlin/build.gradle
donat: This is just an evaluation code we did when kotlin-eclipse was announced. We don't actually ship it with Buildship, not until JB fixes their plugin.

org.eclipse.buildship.kotlin/build.gradle
donat: api

org.eclipse.buildship.core/build.gradle
big-guy: Sorry, I meant the configuration.  Instead of compile, should it be the api or implementation configuration?

org.eclipse.buildship.kotlin/build.gradle
big-guy: Ah OK

org.eclipse.buildship.core/build.gradle
donat: Ah, that was a silly comment from me. 
So, the build script predates api-implementation separation and simply [uses](https://github.com/eclipse/buildship/blob/86fe8c98417d38bc421c02578ab06685f3f0da79/buildSrc/src/main/groovy/eclipsebuild/BundlePlugin.groovy#L141) the `compile` scope. Improving that is out of scope at this point.

donat:
@big-guy Let me know if you have further comments.

Issue #829, opened 2018-12-17
Import should not require all Gradle projects names to be free in the workspace


donat:

donat:
I've linked the solution to the AERI report: https://bugs.eclipse.org/bugs/show_bug.cgi?id=547149

PR #878, opened 2019-04-24
Inform gradle about used project names

fraenkelc:
By default gradle makes sure that project names are unique inside a build.
Since Eclipse supports multiple projects inside a workspace this can still
lead to conflicts. With this change we send the list of non-build related
project names to gradle so it can exclude those during name generation.

This change fixes the import of gradle projects that have no existing
conflicting .project files

Issue: #829


donat:
I've created a Gradle nightly snapshot (5.5-20190430153457+0000) and tested this PR with it. It works quite nicely!

@fraenkelc You can incorporate copy my [changes](https://github.com/eclipse/buildship/commit/a21d078297d459eb8c704c8fb77905ac3d386b68) over your branch. Also, a few integration tests are required before we can merge the PR.

donat:
@fraenkelc I've created a PR for your changes that fixes the caching issue you reported https://github.com/fraenkelc/buildship/pull/1

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/internal/workspace/SynchronizingRenamedProject.groovy
```diff
@@ -47,7 +49,7 @@ class SynchronizingRenamedProject extends ProjectSynchronizationSpecification {
         SynchronizationResult result = trySynchronizeAndWait(sample)
 
         then:
-        result.status.exception instanceof UnsupportedConfigurationException
+        result.status.exception instanceof BuildActionFailureException
```
donat: ❌ So we can't deduplicate root projects, but why? In this case, why not rename the project to `already-there-2`? If that's a hard limitation that we can't fix for Gradle 5.5 then the thrown exception should be `UnsupportedConfigurationException`.

donat:
@fraenkelc Please note that the headless build is also failing. The`IdeFriendlyClassLoader` probably fails for the non-development case.

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/internal/workspace/SynchronizingRenamedProject.groovy
fraenkelc: `HierarchicalElementDeduplicator` in its current form doesn't support it:
https://github.com/gradle/gradle/blob/e975872ae2cf1d521b73f1107cae772ca09c63cd/subprojects/ide/src/main/java/org/gradle/plugins/ide/internal/configurer/HierarchicalElementDeduplicator.java#L40-L43
I guess we can check the Exception if it contains the `IllegalArgumentException` and throw an `UnsupportedConfigurationException`

fraenkelc:
@donat  I've rebased my branch and cherry-picked your changes into the pr. I've also added some more tests and tried to fix the failing tests in the "Basic Test Coverage (Windows, Eclipse 4.6, Java 8)" build.

donat:
It was a mistake. I meant to do it on the eclipse/buildship repository.

donat:
Thank you @fraenkelc!

PR #884, opened 2019-05-16
Let UI specify arguments and JVM arguments containing spaces

donat:
Previously, Buildship relied on a simple text field to display and update program/JVM arguments. The values are stored by splitting the text field content on space characters. This meant that users could not specify arguments that contained spaces.

This PR fixes the problem by replacing the text field with a table where each entry can be edited independently.

New run configurations page:
Screenshot 2019-05-16 at 12 11 00

New workspace/project configuration page (same as above but it doesn't have the `Variables..` button):
Screenshot 2019-05-16 at 12 12 50

New arguments can be added with the `Add` and `Variables..` button (obviously). The editor automatically deletes entries with empty content.


big-guy:
I can't speak much about the Eclipse UI code itself, but LGTM

donat:
Thanks for the review @big-guy. I think it's nice that we get Gradle people more involved in this project.

Issue #376, opened 2017-02-13
Buildship / Oomph Support


rwinch:
It would be good to create [Oomph](https://projects.eclipse.org/projects/tools.oomph) / Buildship support. This was originally discussed on [discuss](https://discuss.gradle.org/t/buildship-tasks-for-oomph/21104/6), but I crated this ticket to track the progress.

I work on Spring framework and I view Oomph support critical to get more users contributing and making it as easy as possible to run sample applications.

oehme:
As mentioned on the forum, this probably needs to be done in Oomph, not in Buildship. If Oomph has some way to contribute a custom importer, then this should be pretty simple.

rwinch:
Thanks for the fast reply. Wouldn't Oomph still be waiting on Public APIs or is that now available?

Perhaps this issue need to be create public APIs for Oomph?

oehme:
If oomph can call a command on a project then it could use the "Add Gradle Nature" command which is public. If it needs a programmatic import API, then you are right, we would need a public and stable one for that.

fraenkelc:
I wrote https://github.com/fraenkelc/buildship3_setuptask a while ago and it works and has some downloads (~80 for the most recent version, I didn't actively advertise it anywhere). Since there looks to be at least some interest in this still would it make sense to add a oomph setup task to buildship?

fraenkelc:
I've got a pull request for this in the works, based partially on my already existing plugin

donat:
Implemented in #892 .

PR #892, opened 2019-06-15
Add an Oomph setup task for import and sync

fraenkelc:
This PR adds an Oomph setup task that can be used to import and synchronize Gradle projects.

This PR implements https://github.com/eclipse/buildship/issues/376


fraenkelc:
This should be functionally complete but for the following bit:
Oomph setup tasks can have enablement annotations. These are used to install the setup task and  required plugins and features for the setup task in a setup model that contains the task. 
The site has to be specified in the ecore model and requires a genmodel run to update (it gets hardcoded into the java source files). I'm not sure what update site to add there for buildship. Oomph itself points at "http://download.eclipse.org/oomph/updates/milestone/latest" - is there a corresponding site for Buildship? 
If not we could skip the annotation (this would require the user to add buildship and import task features to their setup model) or maybe try to filter in the relevant values during the gradle build and run the genmodel generation (or keep placeholders in the generated genmodel file which we replace in the sources before compilation ...)

donat:
@fraenkelc Fortunately, we have a similar update site. For milestones we can use https://download.eclipse.org/buildship/updates/e411/milestones/3.x. For releases, there's https://download.eclipse.org/buildship/updates/e411/releases
Any reason not to use latest release?

fraenkelc:
I've asked for input on the [oomph forums](https://www.eclipse.org/forums/index.php/t/1099618/) and the advice I got was the following: To support updates in the future It would be best to add a composite update site (e.g. https://download.eclipse.org/buildship/updates/latest ) that we can then point to the most recent update site ( https://download.eclipse.org/buildship/updates/e411/releases ).
@donat: would you mind adding another composite to the update sites / do you think it would cause issues ?

donat:
No, it's not that hard, I'll create the update site.

donat:
Finally, I have a bit of time to look at your PR. Impressive work. I'll try to dig into the details and check the classes one by one. 
One quick note: the oomph extension needs to work with the latest Eclipse version, right? It's not a problem now, but if Oomph introduces any breaking changes then we can still ditch your Oomph extension plugins from the builds using older target platforms.

Btw I've created the 'latest' update site for you: https://download.eclipse.org/buildship/updates/latest.

fraenkelc:
Great, I've updated the site and regenerated the relevant files.

Yeah, by using a single update site shared by all target platforms breaking changes in oomph (that we have to follow in the code) might break older platforms.
I'm not sure how likely that is though - the current code builds and unit-tests correctly against all oomph versions starting from the one shipped with eclipse mars. I admit that I haven't gone as far as installing it by hand into all those versions though.

donat:
I think it's reasonable to support only the latest oomph version. Also, if the recent oomph version will stop working with Old eclipse versions then I can just add a condition to exclude it from our build. I think we are good on this front.

buildship.setup
```diff
@@ -269,6 +273,8 @@
       <requirement
           name="com.google.guava"
           versionRange="27.1.0"/>
+      <requirement
```
donat: ❌ The `ProjectsImportTask` setup task  should contain the new projects.

settings.gradle
```diff
@@ -7,4 +7,8 @@ include ':org.eclipse.buildship.branding'
 include ':org.eclipse.buildship.core.test'
 include ':org.eclipse.buildship.ui.test'
 include ':org.eclipse.buildship.stsmigration'
-include ':org.eclipse.buildship.stsmigration.test'
\ No newline at end of file
+include ':org.eclipse.buildship.stsmigration.test'
+include ':org.eclipse.buildship.oomph'
+include ':org.eclipse.buildship.oomph.edit'
+include ':org.eclipse.buildship.oomph.feature'
+include ':org.eclipse.buildship.oomph.test'
```
donat: As mentioned above, how about that we include the oomph plugins when a system property is present?

org.eclipse.buildship.oomph.feature/feature.properties
```diff
@@ -0,0 +1,147 @@
+# Copyright (c) 2019 the original author or authors.
+# All rights reserved. This program and the accompanying materials
+# are made available under the terms of the Eclipse Public License v1.0
+# which accompanies this distribution, and is available at
+# http://www.eclipse.org/legal/epl-v10.html
+# 
+
+featureName =\
+Buildship: Eclipse Plug-ins for Gradle Oomph Import Setup Task
```
donat: Buildship project import support for Oomph

org.eclipse.buildship.oomph.test/gradle.properties
```diff
@@ -0,0 +1,2 @@
+# set SWTBot to put screenshots from failed tests under the build folder
+systemProp.org.eclipse.swtbot.screenshots.dir=build/screenshots
```
donat: Does it works if we move it to the root project's gradle.properties file?

org.eclipse.buildship.oomph.test/src/main/groovy/org/eclipse/buildship/oomph/internal/test/fixtures/GradleVersionProvider.java
```diff
@@ -0,0 +1,77 @@
+/*
+ * Copyright (c) 2018 the original author or authors.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+
+package org.eclipse.buildship.oomph.internal.test.fixtures;
```
donat: ✅ I guess it's on me now to extract duplicate test fixtures to a  base test project.

org.eclipse.buildship.oomph/.classpath
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<classpath>
+	<classpathentry kind="src" path="src/main/java-generated"/>
+	<classpathentry kind="src" path="src/main/resources"/>
+	<classpathentry kind="src" path="src/main/java"/>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.8"/>
+	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins">
+		<accessrules>
+			<accessrule kind="accessible" pattern="org/eclipse/oomph/**"/>
```
donat: Remove, maybe?

org.eclipse.buildship.oomph/model/GradleImport-1.0.ecore
```diff
@@ -0,0 +1,61 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
donat: ❌ Why do we need two ecore files?

org.eclipse.buildship.oomph/src/main/resources/org/eclipse/buildship/oomph/messages.properties
```diff
@@ -0,0 +1,3 @@
+GradleImportTaskImpl_importing=Importing / Synchronizing location {0}
```
donat: Synchronizing location {0}

org.eclipse.buildship.oomph/src/main/resources/org/eclipse/buildship/oomph/messages.properties
```diff
@@ -0,0 +1,3 @@
...
+GradleImportTaskImpl_found_existing=Found existing project {0} for location.
```
donat: Found existing project {0}
(remove dot the end)

org.eclipse.buildship.oomph/src/main/resources/org/eclipse/buildship/oomph/messages.properties
```diff
@@ -0,0 +1,3 @@
...
+GradleImportTaskImpl_import_new=Importing new project from folder.
```
donat: Importing new project from folder 
(without dot at the end)

org.eclipse.buildship.site/build.gradle
```diff
@@ -37,6 +37,11 @@ dependencies {
     localPlugin project(':org.eclipse.buildship.branding')
     localPlugin project(':org.eclipse.buildship.stsmigration')
     localFeature project(':org.eclipse.buildship.feature')
+    if (isOomphSupported()) {
```
donat: Use same mechanism as in the settings.gradle file

org.eclipse.buildship.site/category.xml
```diff
@@ -6,6 +6,10 @@
    <feature url="features/org.eclipse.buildship.kotlin_0.0.0.jar" id="org.eclipse.buildship.kotlin" version="0.0.0">
       <category name="BuildshipExtras"/>
    </feature>
+   <feature url="features/org.eclipse.buildship.oomph_0.0.0.jar" id="org.eclipse.buildship.oomph" version="0.0.0">
+      <category name="Oomph"/>
+   </feature>
    <category-def name="Buildship" label="Buildship: Eclipse Plug-ins for Gradle"/>
    <category-def name="BuildshipExtras" label="Buildship Extras"/>
+   <category-def name="Oomph" label="Buildship: Oomph Setup Task"/>
```
donat: Use the same (or similar) name as the proposed feature name above.

org.eclipse.buildship.oomph.edit/plugin.properties
```diff
@@ -0,0 +1,46 @@
+# Copyright (c) 2019 the original author or authors.
```
donat: For some reason, I can't comment on binary files. This one is for ` org.eclipse.buildship.oomph.edit/icons/full/obj16/GradleImportTask.gif`:

❌ Where did you get this icon? All of our our images in the `org.eclipse.buildship.ui/icons/full` directories are pngs. Can we use them here as well? Also, can we make use of the `@2x` pngs?

build.gradle
```diff
@@ -460,6 +541,13 @@ subprojects {
             url gradleRemoteRepositoryUrl
         }
     }
+    // disable oomph projects on targets without oomph
+    if (['43', '44'].contains(eclipsebuild.Config.on(project).targetPlatform.eclipseVersion) && name.startsWith("org.eclipse.buildship.oomph")) {
```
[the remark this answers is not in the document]
fraenkelc: Sure, we can tie the generation to a project property. Since oomph only needs to be disabled for two of the target platforms I propose defaulting the switch to true.

org.eclipse.buildship.oomph.test/gradle.properties
fraenkelc: ah, this is a relic from copying the framework from `org.eclipse.buildship.ui.test`. I'll remove the whole file.

org.eclipse.buildship.oomph.edit/plugin.properties
fraenkelc: The standard generated class was looking for a gif, so I converted one of the pngs. I should be able to specify pngs by editing `org.eclipse.buildship.oomph.provider.GradleImportTaskItemProvider.getImage(Object)`  and excluding the method from the generation. I'm not sure if the EMFPlugin handles the high dpi images.

org.eclipse.buildship.oomph/.classpath
fraenkelc: Hmm, the initial project creation added that one, not sure why or what for. I've removed those access rules from the .classpath files.

org.eclipse.buildship.oomph/model/GradleImport-1.0.ecore
fraenkelc: The `GradleImport.ecore` is the master where changes are made. `GradleImport-1.0.ecore` is an output file generated by the genmodel step and is used as the schema (and pointed to from the schemaLocation). I guess this separation allows for increasing the version on incompatible model changes and still supporting older versions.

org.eclipse.buildship.oomph/.classpath
fraenkelc: Turns out these access rules are necessary so that the generated code can access oomph classes and methods without warnings.

build.gradle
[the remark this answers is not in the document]
donat: 👍

org.eclipse.buildship.oomph.edit/plugin.properties
donat: Shame, we need to use low-quality images.

org.eclipse.buildship.oomph/.classpath
donat: Thanks for checking. I just wanted to make sure it's necessary to have those rules.

org.eclipse.buildship.oomph/model/GradleImport-1.0.ecore
donat: Ah, so the 1.0.ecore is generated! It's been a while since I've worked with EMF.

org.eclipse.buildship.oomph.edit/plugin.properties
donat: I think the high dpi image handling happening in the SWT layer. It worths a shot to try.

donat:
@fraenkelc Thank you for the contribution!

PR #900, opened 2019-07-09
Align dependency versions of the 411 target platform

fraenkelc:
These versions where required to make the 411 target build for me again.


donat:
Thank you!

PR #905, opened 2019-07-31
Use "test" classpath attribute to separate main and test classpath

donat:
This PR makes use of Eclipse's new test sources feature. Since Gradle 5.6, the `EclipseProject` model [contains](https://github.com/gradle/gradle/pull/9484) the `test` classpath attribute on all source folders and dependencies that belong to a test configuration. 

Gradle 5.6 still [provides](https://github.com/gradle/gradle/pull/9789) the custom scope information, which is used by previous Buildship releases. Buildship falls back to that information if an older Gradle version is being used or if the Host Eclipse version doesn't support test sources (< [4.8](https://www.eclipse.org/eclipse/news/4.8/jdt.php#jdt-test-sources)).

### Testing the feature

1. Install the Eclipse installer: go to https://www.eclipse.org/downloads/ and click the `Download 64` button and proceed with the installer installation.
2. Provision the Buildship IDE
  - Open the Eclipse installer

  - On the product page select Eclipse IDE for Eclipse Committers and change the product version to the latest release

  - On the variables page set the git branch to `donat/test-sources`. Also, set the installation / workspace / git clone location to your liking. 

  - Click finish and wait for the provisioning to finish. Once done, you should see a fully configured IDE with no compilation errors.

3. Launch development version of Buildship to verify changes

 - Create a new Gradle project, try to use a test class in the main sources.
 - You can verify the classpath content by enabling tracing in the `Launch Buildship` configuration.

Screenshot 2019-07-31 at 15 04 18

In the development IDE if you right-click on a Gradle project and select `Gradle > Refresh Gradle project` then you should see the classpath attributes in the host IDE.

### DIFF
--- a/org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/internal/preferences/DefaultModelPersistenceTest.groovy
+++ b/org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/internal/preferences/DefaultModelPersistenceTest.groovy
@@ -9,6 +9,8 @@ import org.eclipse.jdt.core.JavaCore
 
 import org.eclipse.buildship.core.internal.CorePlugin
 import org.eclipse.buildship.core.internal.test.fixtures.WorkspaceSpecification
+import org.eclipse.buildship.core.internal.util.gradle.GradleVersion
+import org.eclipse.buildship.core.internal.util.gradle.GradleVersionTest
 
 class DefaultModelPersistenceTest extends WorkspaceSpecification {
 
@@ -53,8 +55,9 @@ class DefaultModelPersistenceTest extends WorkspaceSpecification {
         command.setBuilderName('custom-command')
         def managedBuilders = [command]
         def hasAutoBuildTasks = true
+        def gradleVersion = GradleVersion.version('5.6')
 
-        PersistentModel model = new DefaultPersistentModel(project, buildDir, buildScriptPath, subProjectPaths, classpath, derivedResources, linkedResources, managedNatures, managedBuilders, hasAutoBuildTasks)
+        PersistentModel model = new DefaultPersistentModel(project, buildDir, buildScriptPath, subProjectPaths, classpath, derivedResources, linkedResources, managedNatures, managedBuilders, hasAutoBuildTasks, gradleVersion)
 
         when:
         CorePlugin.modelPersistence().saveModel(model)
@@ -83,8 +86,9 @@ class DefaultModelPersistenceTest extends WorkspaceSpecification {
         command.setBuilderName('custom-command')
         def managedBuilders = [command]
         def hasAutoBuildTasks = false
+        def gradleVersion = GradleVersion.version('5.6')
 
-        PersistentModel model = new DefaultPersistentModel(project, buildDir, buildScriptPath, subProjectPaths, classpath, derivedResources, linkedResources, managedNatures, managedBuilders, hasAutoBuildTasks)
+        PersistentModel model = new DefaultPersistentModel(project, buildDir, buildScriptPath, subProjectPaths, classpath, derivedResources, linkedResources, managedNatures, managedBuilders, hasAutoBuildTasks, gradleVersion)
         CorePlugin.modelPersistence().saveModel(model)
 
         when:
@@ -108,8 +112,9 @@ class DefaultModelPersistenceTest extends WorkspaceSpecification {
         command.setBuilderName('custom-command')
         def managedBuilders = [command]
         def hasAutoBuildTasks = true
+        def gradleVersion = GradleVersion.version('5.6')
 
-        PersistentModel model = new DefaultPersistentModel(project, buildDir, buildScriptPath, subProjectPaths, classpath, derivedResources, linkedResources, managedNatures, managedBuilders, hasAutoBuildTasks)
+        PersistentModel model = new DefaultPersistentModel(project, buildDir, buildScriptPath, subProjectPaths, classpath, derivedResources, linkedResources, managedNatures, managedBuilders, hasAutoBuildTasks, gradleVersion)
         CorePlugin.modelPersistence().saveModel(model)
 
         when:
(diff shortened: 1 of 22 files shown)


org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/preferences/DefaultPersistentModel.java
```diff
@@ -53,6 +56,7 @@ public DefaultPersistentModel(IProject project, IPath buildDir, IPath buildScrip
         this.managedNatures = ImmutableList.copyOf(managedNatures);
         this.managedBuilders = ImmutableList.copyOf(managedBuilders);
         this.hasAutoBuildTasks = hasAutoBuildTasks;
+        this.gradleVersion = gradleVersion;
```
big-guy: I don't know this all very well, but hopefully this makes some sense...

If we're persisting the Gradle version used, what happens when this is out-of-sync with the actual Gradle build? e.g., I import with Gradle 5.0.  I manually change gradle-wrapper.properties to 5.6.  Are there cases where we'll interpret answers from a Gradle 5.6 version as if we were using Gradle 5.0 (or ask different questions because we think the version of Gradle is older) because the persisted value is different from the real value?

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/workspace/BaseConfigurator.java
```diff
@@ -70,6 +76,8 @@ private void configure(ProjectContext context, IProject project, IProgressMonito
         EclipseProject model = lookupEclipseModel(project);
         progress.worked(1);
 
+        persistentModel.gradleVersion(this.gradleVersion);
```
big-guy: Is configure called if init reports an error to the InitializationContext?  Could this.gradleVersion be null in some cases when init has failed/never called?

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/util/gradle/GradleVersion.java
```diff
@@ -234,6 +233,30 @@ public boolean isValid() {
         return this.versionPart != null;
     }
 
+    public boolean supportsCompositeBuilds() {
```
big-guy: I like this

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/workspace/GradleClasspathContainerRuntimeClasspathEntryResolver.java
```diff
@@ -84,17 +110,62 @@ private void collectContainerRuntimeClasspath(IClasspathContainer container, Lis
...
+            Method method = JavaRuntime.class.getMethod("resolveRuntimeClasspathEntry", IRuntimeClasspathEntry.class, IJavaProject.class, boolean.class);
+            return (IRuntimeClasspathEntry[]) method.invoke(null, projectRuntimeEntry, dependencyProject, excludeTestCode);
+        } catch (Exception e) {
+            throw new GradlePluginsRuntimeException("JavaRuntime.resolveRuntimeClasspathEntry() should not be called when Buildship is installed for Eclipse 4.8", e);
+        }
+    }
+
+    private boolean hasTestAttribute(IClasspathEntry entry) {
+        for (IClasspathAttribute a : entry.getExtraAttributes()) {
+            if ("test".equals(a.getName()) && "true".equals(a.getValue())) {
```
big-guy: WDYT  `Boolean.valueOf(a.getValue())`?  That works ignoring case

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/internal/preferences/DefaultModelPersistenceTest.groovy
```diff
@@ -53,8 +55,9 @@ class DefaultModelPersistenceTest extends WorkspaceSpecification {
         command.setBuilderName('custom-command')
         def managedBuilders = [command]
         def hasAutoBuildTasks = true
+        def gradleVersion = GradleVersion.version('5.6')
```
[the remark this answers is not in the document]
donat: The version is not really important here, I picked 5.6 randomly.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/preferences/DefaultPersistentModel.java
donat: That's correct, Buildship gets out of sync if you change the build (or import) configuration and don't call synchronize manually. This is a conscious decision: we don't want to keep track of files and eagerly call synchronize as it can be a very expensive operation that potentially blocks the user from continue working. The tasks view is a good example. If you add a new task, you have to explicitly click the refresh button to load the model to see the changes.

There's one problematic use-case though: updating from a previous version. We introduced a new entry in `PersistentModel` which will be incomplete if a user updates to the latest version (resulting in a runtime error). To overcome that, I've added a new method to the plugin startup to call synchronization for projects with an incomplete model (see `CorePlugin`).

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/workspace/BaseConfigurator.java
donat: Yes, it is called, although it could only fail if the `BuildEnvironment` can't be loaded (for example for an invalid build). In that case, we already fail earlier, when loading the `EclipseProject` model.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/workspace/GradleClasspathContainerRuntimeClasspathEntryResolver.java
donat: Changed.

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/internal/preferences/DefaultModelPersistenceTest.groovy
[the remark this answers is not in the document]
donat: I've simplified it a bit by using `GradleVersion.current()`

PR #907, opened 2019-08-06
Implement test debugging feature

donat:
This pull request implements test debugging functionality to Buildship. Users can define test launch configuration that can be launched in debug mode. 

The easiest way to try this feature is to import a Gradle project, select a test type or method, then from the context menu select `Debug As > Gradle Test`.

### Launching a test instance

1. Install the Eclipse installer: go to https://www.eclipse.org/downloads/ and click the `Download 64` button and proceed with the installer installation.
2. Provision the Buildship IDE
  - Open the Eclipse installer

  - On the product page select Eclipse IDE for Eclipse Committers and change the product version to the latest release

  - On the variables page set the git branch to `donat/test-debugging`. Also, set the installation / workspace / git clone location to your liking.

  - Click finish and wait for the provisioning to finish. Once done, you should see a fully configured IDE with no compilation errors.

3. Launch development version of Buildship to verify changes

 - Create a new Gradle project, try to use a test class in the main sources.
 - You can verify the classpath content by enabling tracing in the `Launch Buildship` configuration.


org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/configuration/DefaultConfigurationManager.java
```diff
@@ -159,6 +162,40 @@ public void deleteProjectConfiguration(IProject project) {
...
+                  attributes.isShowConsoleView(),
+                  attributes.isShowExecutionView(),
+                  attributes.isOverrideBuildSettings(),
+                  attributes.isBuildScansEnabled(),
+                  attributes.isOffline());
+        return new DefaultRunConfiguration(projectConfiguration, runConfigProperties);
+    }
+
+    @Override
+    public TestRunConfiguration loadTestRunConfiguration(ILaunchConfiguration launchConfiguration) {
```
big-guy: I think I get what's going on here... Before this change, we had a launch configuration for Gradle tasks that could be customized in all sorts of ways and since test execution does something very similar to task execution, you may need that same configuration.

Would it make more sense for some of this to be configured on a project level (console view, execution view, build scans, offline, gradle user home, etc) and only the things that are likely to be different (java, jvm arguments, arguments) be overrideable here? I think it's fine for this to be the same as task execution initially, but reducing the amount of configuration here seems easier for users to understand.  I can't think of a situation where you'd want to run your tests with a different version of Gradle.  I'm not sure that would even work in some cases because we use the project's gradle version to determine what features are available.

donat:
> Is there a button for Gradle synchronization that does the equivalent of Gradle -> Refresh Gradle project?

Only when you open the build script then a new icon appears on the toolbar.

donat:
> It would be nice if we didn't show the "Gradle Test" option at all or complained a little more loudly if test debugging was unsupported by the version of Gradle. When you try debugging with <5.6 and nothing happens, it's a little annoying to figure out why.

Working on it. I'll implement it before merging the PR.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/configuration/DefaultConfigurationManager.java
donat: The current state is in line with the UI. Each time you right-click on a test and choose `Debug As... > Gradle Test` a new launch configuration is persisted. By default, they all inherit the project settings, but you have the option to change it. I think it's useful to let the user configure all project properties for single test execution. They can play around with a single run configuration, without modifying the project config (which would have an affect on all other run configs).

donat:
I'll merge the current state and implement the improvements in a separate PR.
This way users can try out this feature earlier.

PR #913, opened 2019-08-27
Test debugging improvements

donat:
This pull request fixes the remaining TODOs from the original test debugging pull request:
- Only show Debug As.. > Gradle Test context menu item if the selected project is open and uses Gradle 5.6+
- Merge test launch validation to a TestExecutionTarget class


big-guy:
LGTM.  

I had an initial problem when switching between an older version of Gradle to 5.6.1. 

This is what I did:
- I generated a library Java project with JUnit tests with init and imported it with Gradle 5.1.1.  I added a `println gradle.gradleVersion` in the build file so I could tell which version was being used. The Gradle Test option didn't show up under Debug As (as expected). 
- I switched the version of Gradle to 5.6.1 in the wrapper properties file
- I refreshed the Gradle project (right click on project, Gradle -> Refresh)
- I checked the Debug As menu and Gradle Test was still missing (this was wrong).
- I tried refreshing more times and verified that 5.6.1 was being used (based on console output).
- I deleted any kind of run configurations for the tests.  Gradle Test was still missing from Debug As...
- I switched back to 5.1.1 and refreshed. Then I switched to 5.6.1 and refreshed.  Still not working.
- After that, I restarted Eclipse and the Gradle Test option showed up under Debug As when using 5.6.1.
- I tried switching between 5.1.1/5.6.1 and things worked as expected (disappeared/reappeared).

I tried starting over and reproducing the problem, but I couldn't get it to occur. I'm not sure the initial state of everything because I had used this same project to verify the earlier PR.

donat:
@big-guy I think  I know what's going on here. When you say `imported it with Gradle 5.1.1` you actually selected the Gradle 5.1.1 version in the import dialog. This is basically equivalent to calling [GradleConnector.useGradleVersion(String)](https://docs.gradle.org/current/javadoc/org/gradle/tooling/GradleConnector.html#useGradleVersion-java.lang.String-). Using this option ignores the version in the `gradle.properties` file. You need to import the project with the `Gradle Wrapper` option to make your steps working.

That being said, this functionality could be confusing for others too. I have a couple of ideas on how to improve the situation.

Issue #908, opened 2019-08-12
Support for multi-composite builds


Kameecoding:
Buildship should properly support multi-composite builds, e.g. when multiple projects use a shared library and want to include it.

There is already a few issue reported here that relate to this.

Let's imagine a scenario where you have a library that you are developing that is used by multiple projects.

`libraryProject`

then you have two projects A and B

    A - settings.gradle
    includeBuild "../libraryProject"

 and also an application where you use both A and the library:

    B - settings.gradle
    includeBuild "../libraryProject"
    includeBuild "../B"

This not an unreasonable scenario if you want to have a dev environment where you want to see all the changes that are being made, yet the projects themselves are built separately for release.

However when you try to import the `app` using buildship 3.+ it fails with the message that multiple projects share the same directory, in this case it's cause because the `libraryProject` gets duplicated and buildship detects it as two different projects.

In buildship 2.0 this worked, with some minor inconveniences:
e.g. https://github.com/eclipse/buildship/issues/360

Similar case reported here: https://github.com/eclipse/buildship/issues/877

euschn:
I have exactly the problem as described above with my project. The project has multiple composite build dependencies and works fine in the command line and in Netbeans, but not with Eclipse/Buildship.

Please fix this, as it keeps me from developing my project in Eclipse. If there is some workaround to this problem, I would also greatly appreciate it.

donat:
Related issue: https://github.com/gradle/gradle/issues/8031

donat:
I've created a [WIP PR](#962) for solving this issue and I'm planning to publish a snapshot release based on it. @Kameecoding (and anybody interested): would you be interested trying it out and/or reviewing my changes?

Kameecoding:
@donat I can test on our older codebase, we have switched away from composite builds since because we weren't able to update to Buildship 3+.

donat:
The update site of the snapshot is available here: https://download.eclipse.org/buildship/updates/e414/snapshots/3.x/3.1.4.v20200108-1545-s

AndreasFagschlunger:
@donat Just wanted you to know that in my setup it works. I used eclipse 2019-12 with the latest Buildship plugin version and within eclipse the project had an error saying something about two projects using the same path. Then I installed the snapshot version, restarted eclipse, refreshed gradle project and the error was gone. Love it, thanks!

ams-tschoening:
[...]

Issue #877, opened 2019-04-23
Composite Build Import Fails With UnsupportedConfigurationException in version 3.0.1


davisbri:
I have a project that was developed using Buildship version 2.2.1.v20180125-1441 and Gradle version 5.2.1.. After updating Buildship to version 3.0.1.v20181217-1554, importing the project into eclipse fails with an UnsupportedConfigurationException: Multiple projects in the build are located in the same directory. 

This project builds using Gradle outside of Eclipse, and we are able to continue using Buildship if we downgrade the plugin to version 2.2.1 or 2.2.2. Is there a reason this configuration became unsupported in version 3.0.1?

I've created a sample/dummy project to reproduce this issue: [CompositeBuildTest](https://github.com/davisbri/CompositeBuildTest).
To reproduce: 
Clone repository above. 
Using Buildship version 3.0.1, import the project into eclipse by setting the Project Root Directory to the CoreProject and by using gradle version 5.2.1 or above (see screenshots below).

My environment consists of Windows 10 Pro (v1809), Eclipse  2019-3(4.11.0), Buildship 3.0.1.v20181217-1554

![Wizard1](https://user-images.githubusercontent.com/5470231/56592712-70a0ac80-65a8-11e9-8167-a5a197ac295f.PNG)
![Wizard2](https://user-images.githubusercontent.com/5470231/56592721-726a7000-65a8-11e9-8976-f2dd4d73d3e3.PNG)
![Wizard3](https://user-images.githubusercontent.com/5470231/56592728-75656080-65a8-11e9-91de-14fba0e7de66.PNG)
![Eclipse](https://user-images.githubusercontent.com/5470231/56592737-78605100-65a8-11e9-8832-68018f6222c8.PNG)
![ImportError](https://user-images.githubusercontent.com/5470231/56592740-79917e00-65a8-11e9-987f-82bee2bab876.PNG)

hanct:
Have u solved this issue? Cos I encounter the same issue previously. If I dont recall wrongly, i think i resolve it by ensureing the settings.gradle file's rootProject.name is set to the name of the project folder u import. For your case, CoreProject.

davisbri:
No, I have not solved the issue yet. Currently, the settings.gradle file for the imported project folder has the rootProject.name property properly set to 'CoreProject', and it's still an issue.

catonymous:
This is a really annoying bug. The only way to work it around is to comment composite builds in settings.gradle, to import/synchronize and then uncomment them back.

[...]

PR #962, opened 2020-01-08
Let users import multi-composite builds

donat:
It is a valid use-case in Gradle to have a composite build that includes the same build multiple times. Unfortunately, Buildship doesn't handle this scenario well. Previous versions showed duplications in the Tasks view, and recent versions simple disallowed the import of those projects. For more context see https://github.com/eclipse/buildship/issues/908 and https://github.com/eclipse/buildship/issues/877.

The root cause of all those problems that the Tooling API depends on the same classes as the `eclipse` Gradle plugin and creates a new `EclipseProject` model for each sub-project. For included builds it ends up having duplicate models with the same project directory. For the `eclipse` plugin it's not a problem, because at the end it just writes the same `.project` and `.classpath` files multiple times. 

This PR solves the situation by eliminating the duplicate models when collecting the `EclipseProject` instances based on the project directory. It might not be an optimal solution, but it unblocks users and works in a backward-compatible way (works with all Gradle versions).

Overall, it's another example of why Buildship should have its own separate TAPI model builder that doesn't depend on the `eclipse` plugin.

### DIFF
--- a/org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/internal/workspace/GradleClasspathContainerUpdaterTest.groovy
+++ b/org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/internal/workspace/GradleClasspathContainerUpdaterTest.groovy
@@ -47,7 +47,7 @@ class GradleClasspathContainerUpdaterTest extends WorkspaceSpecification {
         PersistentModelBuilder persistentModel = persistentModelBuilder(project.project)
 
         when:
-        Set allProjects = HierarchicalElementUtils.getAll(gradleProject).toSet()
+        Collection allProjects = collectAllProjects(gradleProject)
         GradleClasspathContainerUpdater.updateFromModel(project, gradleProject, allProjects, persistentModel, null, null)
 
         then:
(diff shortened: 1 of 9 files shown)


ams-tschoening:
#511 is related as well. Testing this PR right now and things look good for some projects already, the former error message has been replaced by a warning. What's the reason for issuing the warning at all? Isn't that unnecessary if [multi-composite builds are a supported use case](https://github.com/gradle/gradle/issues/8031#issuecomment-447118695)?

deanorderly:
Anyway I can download this build version? I am ok with the duplicates in gradle task view(I never use that anyways).  I want to test out eclipse and the potential for switching our team over.

Another question regarding this pull request is multiple microservices in a monorepo sharing libraries.  (ie. can I import microserviceA and then import microserviceB such that I can start both up locally and develop).

We have 5 microservices that all work and talk together locally so we frequently have to boot up the auth service, customer api service and incomingdb service OR auth service, customer portal and admin portal.  etc. etc.  We have quite a few situations where we boot up 3 services on a certain use case that we want to debug end to end.  The issue though is a second import WILL bring in the same libraries again used by the other services.  Well, I hope that is not an issue with this fix?

**thanks for the work on this fix btw!!!!  This will rock once it's in...or at least get this one in and iterate AS RIGHT NOW, we can't use eclipse at all :(.**

donat:
Closing in favor of a more general solution implemented in #1036 /

PR #997, opened 2020-06-10
Adds external dependencies to also honor .rar and other archives

### LINKED ISSUE #996: Archives using other extensions then .jar or .zip not added to the classpath
### Expected Behavior
When adding a dependency that is an accepted Eclipse classpath entry it should be taken into account as the general .

### Current Behavior
Dependencies as resource archive files (`.rar`) being Java `.jar` files using only the `.rar` extension or mixed case archive extensions as `.Zip` or `.ZIP`  originating from some Windows based Tools are ignored when `GradleClasspathContainerUpdater` is calculating the classpath entries.

### Steps to Reproduce 
Adding a dependency to a ZIP archive resource using an extension other than `.zip` or `.jar`

### Your Environment
Linux, Gradle 6.5, Eclipse 4.15.0

reinhapa:
Fixes #996

### Context
Relaxes filter of taken external dependencies into account. Here also `.rar` and case sensitive versions of extensions will be accepted. Also lastly a existing file must be at least an ZIP archive.

### DIFF
--- a/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/workspace/GradleClasspathContainerUpdater.java
+++ b/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/workspace/GradleClasspathContainerUpdater.java
@@ -104,7 +104,7 @@ final class GradleClasspathContainerUpdater {
                 String dependencyName = dependencyFile.getName();
                 // Eclipse only accepts folders and archives as external dependencies (but not, for
                 // example, a DLL)
-                if (dependencyFile.isDirectory() || isZipArchiveSuffix(dependencyName)) {
+                if (dependencyFile.isDirectory() || hasAcceptedSuffix(dependencyName)) {
                     IPath path = org.eclipse.core.runtime.Path.fromOSString(dependencyFile.getAbsolutePath());
                     File dependencySource = dependency.getSource();
                     IPath sourcePath = dependencySource != null ? org.eclipse.core.runtime.Path.fromOSString(dependencySource.getAbsolutePath()) : null;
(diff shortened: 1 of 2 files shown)


org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/workspace/GradleClasspathContainerUpdater.java
```diff
@@ -117,6 +119,21 @@ private void updateClasspathContainer(PersistentModelBuilder persistentModel, IP
         return result.build();
     }
 
+    private boolean isZipArchiveSuffix(String dependencyName) {
+       String dependencyNameLwc = dependencyName.toLowerCase();
+        return dependencyNameLwc.endsWith(".jar") || dependencyNameLwc.endsWith(".rar") || dependencyNameLwc.endsWith(".zip");
+    }
+
+    private boolean isZipArchive(File dependencyFile) {
```
donat, line 127: ❌ I think we should avoid doing extra IO operations here. Filtering based on the extensions should be fine. Is there a non-accidental use case where we cannot expect the proper file extension?

reinhapa: @donat in our case we do only have .rar at the moment, so that could be left out. Eventually we should make those extensions configurable to catch additional ones if feasible...

donat: That makes sense. We could have a workspace-scoped configuration with a list of accepted extensions. I'm happy to help if you are interested in contributing this feature.

donat: That makes sense. We could have a workspace-scoped configuration with a list of accepted extensions. I'm happy to help if you are interested in contributing this feature.

donat:
Running CI build: https://builds.gradle.org/viewQueued.html?itemId=35972227

donat:

reinhapa:
@donat you're welcome 😎

PR #1000, opened 2020-06-18
Put Gradle error marker on the build scripts

### LINKED ISSUE #999: Attach unresolved dependency marker to build file instead of project
I tried to use the latest Buildship Snapshot in jdt.ls, following https://github.com/eclipse/buildship/pull/994#issuecomment-645321645. It kinda works, but the marker is attached to the project, instead of the build.gradle file. And vscode really doesn't like it:

Screen Shot 2020-06-17 at 7 06 13 PM

I'd like to avoid adding more hacks to jdt.ls, so it'd be really helpful if such errors were attached to the build descriptor instead of the project.

donat:
Fixes #999


org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/workspace/ProjectConfigurators.java
```diff
@@ -112,17 +140,38 @@ private String configuratorFailedMessage(InternalProjectConfigurator contributio
...
+            int lineNumber = 0;
+            if (exception instanceof UnresolvedDependencyException) {
+                String coordinates = ((UnresolvedDependencyException)exception).getCoordinates();
+                InputStream is = buildScript.getContents(true);
+                try (BufferedReader reader = new BufferedReader(new InputStreamReader(is))) {
+                    String line;
+                    int num = 0;
+                    while((line = reader.readLine()) != null) {
+                        num++;
+                        if (line.contains(coordinates)) {
```
leonardobsjr, line 165: It should be possible to try to find the line even if the dependency string is malformed by replacing the colons by spaces on the line. Don't how how hackish is this tho.

donat: It's already ridiculously hackish. The coordinates should come from Gradle. Your upcoming PR should make this colon replacement obsolete.

donat: I think the line number solution is not good enough. I'll create a simpler PR that puts the error marker on the build scripts without the line number and add that capability to the configurator API separately.

PR #1010, opened 2020-07-27
Show test output in Console view

donat:
### Context

Right now the test output is only visible in the Executions view. This is not something a user would expect.

Screenshot 2020-07-29 at 17 00 20

This PR filters out the test output events from the Executions view and prints it to the Gradle console that corresponds to the Gradle build:

Screenshot 2020-07-29 at 17 02 10

The forwarding works for task and test execution as well as for tests executed during project synchronization.


donat:
This is not correct. We should make the executions view more user-friendly, or simply show the test output in a different place (preferably in the JUnit view).

PR #1022, opened 2020-08-03
Format test output messages in the Executios view

donat:
### Context

Before this PR, the Executions view showed the test output events in the order they appeared and with a distracting `StdOut`/`StdErr` prefixes.

Screenshot 2020-08-03 at 10 20 37

This PR uses a red foreground color to distinguish messages that were printed on stderr. Also, the output events are now sorted: stdout messages always come before the ones on stderr.

Screenshot 2020-08-03 at 10 20 02

A follow-up PR will be to add new icons for the test output events.

### DIFF
--- a/org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/internal/util/color/ColorUtils.java
+++ b/org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/internal/util/color/ColorUtils.java
@@ -44,7 +44,6 @@ public final class ColorUtils {
      */
     public static ColorDescriptor getDecorationsColorDescriptorFromCurrentTheme() {
         ITheme theme = PlatformUI.getWorkbench().getThemeManager().getCurrentTheme();
-        System.out.println(theme.getColorRegistry().getKeySet());
         return Preconditions.checkNotNull(theme.getColorRegistry().getColorDescriptor(DECORATIONS_COLOR));
     }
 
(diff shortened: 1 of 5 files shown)


org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/internal/util/color/ColorUtils.java
```diff
@@ -43,7 +44,18 @@ public static Color getDecorationsColorFromCurrentTheme() {
      */
     public static ColorDescriptor getDecorationsColorDescriptorFromCurrentTheme() {
         ITheme theme = PlatformUI.getWorkbench().getThemeManager().getCurrentTheme();
+        System.out.println(theme.getColorRegistry().getKeySet());
```
big-guy, line 47: should this System.out be here?

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/internal/util/color/ColorUtils.java
```diff
@@ -43,7 +44,18 @@ public static Color getDecorationsColorFromCurrentTheme() {
...
     }
 
+    /**
+     * Returns the color descriptor for {@code ERROR_COLOR} from the current workbench theme.
+     *
+     * @return the theme color descriptor to decorate text
+     */
+    public static ColorDescriptor getErrorColorDescriptorFromCurrentTheme() {
+        ITheme theme = PlatformUI.getWorkbench().getThemeManager().getCurrentTheme();
+        System.out.println(theme.getColorRegistry().getKeySet());
```
big-guy, line 58: should this System.out be here?

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/internal/util/color/ColorUtils.java
donat: Of course, it's essential ;)
I've removed it.

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/internal/util/color/ColorUtils.java
donat: Removed

PR #1036, opened 2020-11-12
Do not attempt to import the same build twice

jjohannes:
If the same build is included several places in the build tree, we should not attempt to import it several times.

This PR also proposes to upgrade the TAPI to the lates 6.8 nightly. The "Can handle cyclic build includes" will only pass with a TAPI version that contains these fixes:
- https://github.com/gradle/gradle/pull/15148.
- https://github.com/gradle/gradle/pull/15039
- https://github.com/gradle/gradle/pull/14671
- https://github.com/gradle/gradle/pull/15183


jjohannes:
@donat this is good to go now with the latest 6.8 nightly. Before the next buildship release, we should update to the final 6.8 TAPI release.

donat (approved):
Looks good! Thanks for your contribution!

donat:
I think this justifies a service release. I'll take care of it once Gradle 6.8 is out.

PR #1040, opened 2020-11-30
GradleVersions: comply with XDG base spec

GovanifY:
### Context

This PR ensures compatible systems are compliant with the XDG base specification ( https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html ). If no XDG related environment variables are set then the folder used defaults to $HOME/.cache/tooling for the gradle versions cache, as per the standard.

### DIFF
--- a/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/util/gradle/PublishedGradleVersions.java
+++ b/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/util/gradle/PublishedGradleVersions.java
@@ -218,7 +218,12 @@ public final class PublishedGradleVersions {
     }
 
     private static File getCacheFile() {
-        return new File(System.getProperty("user.home"), ".tooling/gradle/versions.json");
+        // ensures compliance with XDG base spec: https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html
+        String xdgCache = System.getenv("XDG_CACHE_HOME");
+        if (xdgCache == null) {
+            xdgCache = System.getProperty("user.home") + "/.cache/";
+        }
+        return new File(xdgCache, "tooling/gradle/versions.json");
     }
 
     /**


org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/util/gradle/PublishedGradleVersions.java
```diff
@@ -218,7 +218,12 @@ private static URL createURL(String url) {
     }
 
     private static File getCacheFile() {
-        return new File(System.getProperty("user.home"), ".tooling/gradle/versions.json");
+        String xdgCache = System.getenv("XDG_CACHE_HOME");
```
donat, line 221:
```suggestion
        // ensures compliance with XDG base spec: https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html
        String xdgCache = System.getenv("XDG_CACHE_HOME");
```

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/util/gradle/PublishedGradleVersions.java
```diff
@@ -218,7 +218,12 @@ private static URL createURL(String url) {
...
+        if (xdgCache == null) {
+            // $HOME/.cache by default
```
donat, line 223: Redundant comment is redundant :)

donat (changes requested):
I'm fine with the change, I just have a couple of comment-related suggestions. 

Also, I need to ask you to sign the ECA: https://www.eclipse.org/projects/handbook/#resources-github.

GovanifY:
Thanks! Commited your suggestions, will get back once I've signed the ECA!

GovanifY:
This should be good now! Also added the Signed-Off required.

donat (approved):
lgtm

donat:
@GovanifY thank you for the contribution!

PR #1043, opened 2020-12-03
Allow running tasks from included builds

donat:
# Description

This PR enables the task execution functionality for included builds; a feature introduced by Gradle 6.8.

In order to properly distinguish between the root and the included builds, I needed to adjust how we query TAPI models (i.e `EclipseProject`). I needed to distinguish what model belongs to the root build and what model belongs to a particular included build. To do that, I've adjusted `CompositeModelQuery` to return `Map<String, MODEL>` instead of the list of models. 

The UI is adjusted to our expectations regarding the task execution. The task selectors are removed from the included builds from the tasks view. Here's a typical content:

image

When the user double-clicks on a task from an included build, the task executed via the identity path:

image

When executing "included tasks", the working directory is always set to the composite root dir. 
Also, the run configuration stores the identity path correctly, so that re-executing the same task works as expected:

image

When the user imports a composite build with an older Gradle version (<6.8) then the included tasks are disabled. The context menu has a clue on how to enable those tasks.

image

# Evaluation

### 1. Install the Eclipse installer

Go to https://www.eclipse.org/downloads/ and click the `Download 64` button and proceed with the installer installation.

### 2. Provision the Buildship IDE

- Open the Eclipse installer

- On the product page select Eclipse IDE for Eclipse Committers and change the product version to `Latest Release (2020-09)`

- On the variables page set the git branch to `donat/exec-tasks-from-included-builds`. Also, set the installation / workspace / git clone location to your liking.

- Click finish and wait for the provisioning to finish. Once done, you should see a fully configured IDE with no compilation errors

## Launch development version of Buildship to verify changes

- Import a composite build with Gradle 6.8-rc-1
- Open the task view and try to execute some tasks
- Open the preferences page, change the Gradle version to 6.7.1, refresh the task view and observe the disabled tasks.


org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/internal/view/task/BuildNode.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    public boolean isIncludedBuild() {
+        return this.includedBuildName != null;
+    }
+
+    public String getIncludedBuildName() {
+        return this.includedBuildName;
+    }
+
```
donat, line 49:
```suggestion
```

PR #1095, opened 2021-06-22
Check JDT.LS bundle when org.eclipse.platform is unavailable

### LINKED ISSUE #1094: Buildship cannot resolve classpath correctly in VS Code Java
The code here: https://github.com/eclipse/buildship/blob/master/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/util/eclipse/PlatformUtils.java#L22

```java
public static boolean supportsTestAttributes() {
    Bundle platformBundle = Platform.getBundle("org.eclipse.platform");
    if (platformBundle == null) {
        return false;
    }
    Version platform = platformBundle.getVersion();
    Version eclipseLuna = new Version(4, 8, 0);
    return platform.compareTo(eclipseLuna) >= 0;
    }
```
Tries to check the eclipse version via the bundle `org.eclipse.platform`. Since this bundle depends on some UI related components, so it's not included in JDT.LS

Then in https://github.com/eclipse/buildship/blob/master/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/workspace/GradleClasspathContainerRuntimeClasspathEntryResolver.java#L87-L91

it will fallback to old custom scope attributes, which is not correct enough.

jdneo:
Fixes #1094 

### Context
related with https://github.com/microsoft/vscode-java-debug/issues/1008

More details can be found in #1094.

Basically, we also check `org.eclipse.jdt.ls.core` when `org.eclipse.platform` is unavailable in the JDT.LS context.
Then the buildship can call `GradleClasspathContainerRuntimeClasspathEntryResolver.runtimeClasspathWithTestSources()` to get more accurate classpath in VS Code Java

### DIFF
--- a/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/util/eclipse/PlatformUtils.java
+++ b/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/util/eclipse/PlatformUtils.java
@@ -21,10 +21,20 @@ public class PlatformUtils {
     public static boolean supportsTestAttributes() {
         Bundle platformBundle = Platform.getBundle("org.eclipse.platform");
         if (platformBundle == null) {
-            return false;
+        	// the bundle "org.eclipse.platform" will be null when it's JDT.LS
+        	// in that case we check the JDT.LS bundle
+            return supportsTestAttrubutesInJdtLs();
         }
         Version platform = platformBundle.getVersion();
         Version eclipseLuna = new Version(4, 8, 0);
         return platform.compareTo(eclipseLuna) >= 0;
+    }
+    
+    private static boolean supportsTestAttrubutesInJdtLs() {
+    	Bundle lsBundle = Platform.getBundle("org.eclipse.jdt.ls.core");
+        if (lsBundle == null) {
+            return false;
+        }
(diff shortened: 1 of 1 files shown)


org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/util/eclipse/PlatformUtils.java
```diff
@@ -21,10 +21,20 @@
...
+        	// the bundle "org.eclipse.platform" will be null when it's JDT.LS
+        	// in that case we check the JDT.LS bundle
+            return supportsTestAttrubutesInJdtLs();
         }
         Version platform = platformBundle.getVersion();
         Version eclipseLuna = new Version(4, 8, 0);
         return platform.compareTo(eclipseLuna) >= 0;
     }
+    
+    private static boolean supportsTestAttrubutesInJdtLs() {
```
testforstephen, line 33: typos. 

supportsTestAttrubutesInJdtLs -> supportsTestAttributeInJdtLs

jdneo:
Hi @donat, 👋

This is my first time contributing code to Buildship. Would you mind to take a look at this? Also please let me know if there is anything I missed.

donat (approved):

donat:
@jdneo thank you for the contribution!

PR #1110, opened 2021-09-28
Do not attempt to import the root build multiple times

jjohannes:
If the root build is involved in an included build cycle, this was not detected correctly. Because the path of the root name is ':', while its name is something else.

This is a specific edge case that was covered by the changes in:
https://github.com/eclipse/buildship/pull/1036

Unfortunately no test cases for this specific edge case was added and it was broken in the follow up PR:
https://github.com/eclipse/buildship/pull/1044

This PR fixes the issue by specifically considering the root build name.
It also adds two tests to cover this case.


donat (approved):
LGTM. 
I'll run your changes through CI and merge after.

donat:
@jjohannes Thank you for the contribution!

PR #1114, opened 2021-10-14
Try to find the project first when load configuration

### LINKED ISSUE #1111: Buildship is not using Eclipse APIs to persist the .prefs files
Currently the Buildship uses [an own implementation](https://github.com/eclipse/buildship/blob/35d01caad3ab023fdd9fdf4a1b9e9c738bff7fd0/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/configuration/PreferenceStore.java#L40) to read/write preferences. But not using Eclipse API to do such jobs. Is it intended for some reasons? (See the **Context** section for more explanation about this request)

### Expected Behavior

Use APIs provided by Eclipse to read and write preferences to the `.settings/` folder

### Current Behavior

The current implementation hard-code the .prefs file's location and read/write it directly via Java IO APIs

### Context

In VS Code Java, there is an issue complained mostly about stop generating .project, .classpath, .settings files at the project root: https://github.com/redhat-developer/vscode-java/issues/618. After some investigation, we found that we can use the Eclipse File System Provider extension point to customize the file location for some specific files. But that requires other plugins use Eclipse standard APIs to read and write files.

We found that In Buildship, there is a customized implementation to deal with the preferences, and the file is persisted directly [via Java IO APIs](https://github.com/eclipse/buildship/blob/35d01caad3ab023fdd9fdf4a1b9e9c738bff7fd0/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/configuration/PreferenceStore.java#L246-L249).

This makes the extension point not working, so we'd like to know if this is done by some reasons? Or this is can be fixed to use the standard Eclipse API.

jdneo:
Signed-off-by: Sheng Chen <sheche@microsoft.com>

Fixes #1111

### Changed things
1. As other methods in `DefaultConfigurationManager`, It will first try to find the project by the given dir path, if it's found. use the platform API to deal with the preferences file. Otherwise fallback to IO based approach.
2. In import stage, I changed to task order to first import the project then persist the configurations.
   > Same procedures can be found in m2e, `org.eclipse.m2e.core.internal.project.ProjectConfigurationManager#importProjects() -> create() -> enableBasicMavenNature() -> saveResolverConfiguration()`
3. In test cases, because we will persist the preferences after project import, so the non-existence folder won't exist when starts import, so the result status is changed to warning. (Previously, the folder is created when persisting the preferences file via [Java IO API](https://github.com/eclipse/buildship/blob/0c31d23e4af982c3f8e61f31b70dda495203c888/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/configuration/PreferenceStore.java#L241-L244))

### Context

More context can be found in the context section of #1111 

_// cc @rgrunber @fbricon for awareness._

### DIFF
--- a/org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/SynchronizationInvalidLocationTest.groovy
+++ b/org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/SynchronizationInvalidLocationTest.groovy
@@ -16,7 +16,7 @@ import org.eclipse.buildship.core.internal.test.fixtures.ProjectSynchronizationS
 
 class SynchronizationInvalidLocationTest extends ProjectSynchronizationSpecification {
 
-   def "Can import a nonexistent location"() {
+   def "Cannot import a nonexistent location"() {
        setup:
        File location = new File('nonexistent')
 
@@ -24,7 +24,8 @@ class SynchronizationInvalidLocationTest extends ProjectSynchronizationSpecifica
        SynchronizationResult result = tryImportAndWait(location)
 
        then:
-       result.status.isOK()
+       result.status.severity == IStatus.WARNING
+       ToolingApiStatusType.IMPORT_ROOT_DIR_FAILED.matches(result.status)
    }
 
    def "Cannot import a plain file"() {
--- a/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/configuration/DefaultConfigurationManager.java
+++ b/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/configuration/DefaultConfigurationManager.java
@@ -115,7 +115,7 @@ public class DefaultConfigurationManager implements ConfigurationManager {
         if (project.isAccessible()) {
             try {
                 pathToRoot = this.buildConfigurationPersistence.readPathToRoot(project);
-            } catch (RuntimeException e) {
+            } catch (Exception e) {
                 // fallback to the file IO based preferences store.
             }   
         }
--- a/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/configuration/DefaultConfigurationManager.java
+++ b/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/configuration/DefaultConfigurationManager.java
@@ -111,7 +111,17 @@ public class DefaultConfigurationManager implements ConfigurationManager {
 
     @Override
     public ProjectConfiguration loadProjectConfiguration(IProject project) {
-        String pathToRoot = this.buildConfigurationPersistence.readPathToRoot(project.getLocation().toFile());
+        String pathToRoot = null;
+        if (project.isAccessible()) {
+            try {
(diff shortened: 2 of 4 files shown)


jdneo:
Hi @donat! 👋

Any chance to take a look at this PR?

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/configuration/DefaultConfigurationManager.java
```diff
@@ -111,7 +111,17 @@ public ProjectConfiguration createProjectConfiguration(BuildConfiguration config
 
     @Override
     public ProjectConfiguration loadProjectConfiguration(IProject project) {
-        String pathToRoot = this.buildConfigurationPersistence.readPathToRoot(project.getLocation().toFile());
+        String pathToRoot = null;
+        if (project.isAccessible()) {
+            try {
+                pathToRoot = this.buildConfigurationPersistence.readPathToRoot(project);
+            } catch (RuntimeException e) {
```
testforstephen, line 118: catch Exception instead of RuntimeException?

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/SynchronizationInvalidLocationTest.groovy
```diff
@@ -16,15 +16,16 @@ import org.eclipse.buildship.core.internal.test.fixtures.ProjectSynchronizationS
 
 class SynchronizationInvalidLocationTest extends ProjectSynchronizationSpecification {
 
-   def "Can import a nonexistent location"() {
+   def "Cannot import a nonexistent location"() {
```
donat, line 19: 👍

donat (approved):

donat:
Running through CI.

donat:
@jdneo thank you for your contribution!

jdneo:
Thank you @donat for the help😀

jdneo:
Hi @donat, would you mind to publish a new snapshot to the [Eclipse software repository](https://download.eclipse.org/buildship/updates/e416/snapshots/3.x/)?

donat:
Sure.
We have some difficulties at the moment as our release process relies on the Eclipse jar signing service at build.eclipse.org, which is no longer available.

rgrunber:
I believe they migrated the service away from build.eclipse.org. For most projects this required just updating their eclipse-jarsigner-plugin, but for those manually referencing the service :

For example:
```
-                  <signerUrl>http://build.eclipse.org:31338/api/sign/jce</signerUrl>
+                  <signerUrl>https://cbi.eclipse.org/api/sign/jce</signerUrl>
```

donat:
I'm trying to make use of https://cbi.eclipse.org/jarsigner/sign now. We use Gradle to build Buildship so I need to invoke the signing service programatically.

donat:
@jdneo a new snapshot is now available.

jdneo:
@donat Yes I've seen it. Thank you for the help 😀

Issue #1151, opened 2022-04-22
Prefer the usage of OSGi ds services instead of the CorePlugin service registration


vogella:
In #1148 a deadlock was identified and I looked into the org.eclipse.buildship.core.internal.CorePlugin which registers tons of OSGi services in its start method. You should move these to OSGi DS services.

For this:

1.) Activate the DS Preferences via Windows -> Preferences -> Plug-in Development  -> Ds Annotations 

![image](https://user-images.githubusercontent.com/139910/164699979-9d261ba3-62f6-4745-928c-924cf4f67cc4.png)

2.) This allows you to use the @Component annotation to define OSGi immediate components or OSGi services. The 1.) ensure that the IDE generates the corresponding meta-data for these services.

cc @laeubi and @HannesWell as they are working on this in platform and might be interested in helping buildship to update.

PR #1152, opened 2022-04-22
Moving PublishedGradleVersionsWrapper to a declarative OSGi service

vogella:
With this change the PublishedGradleVersionsWrapper is published as
declarative OSGi service instead registering it via the activator.

For #1151

Fixes #?

### DIFF
--- a/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/CorePlugin.java
+++ b/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/CorePlugin.java
@@ -12,6 +12,7 @@ package org.eclipse.buildship.core.internal;
 import java.util.Dictionary;
 import java.util.Hashtable;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 
 import org.osgi.framework.Bundle;
@@ -95,7 +96,6 @@ public final class CorePlugin extends Plugin {
     // service tracker for each service to allow to register other service implementations of the
     // same type but with higher prioritization, useful for testing
     private ServiceTracker loggerServiceTracker;
-    private ServiceTracker publishedGradleVersionsServiceTracker;
     private ServiceTracker workspaceOperationsServiceTracker;
     private ServiceTracker internalGradleWorkspaceServiceTracker;
     private ServiceTracker processStreamsProviderServiceTracker;
@@ -110,6 +110,9 @@ public final class CorePlugin extends Plugin {
     private DefaultExternalLaunchConfigurationManager externalLaunchConfigurationManager;
     private ToolingApiOperationManager operationManager;
     private ExtensionManager extensionManager;
+    
+    private static Map<Class<?>, ServiceTracker<?, ?>> trackers = new ConcurrentHashMap<>();
+
 
     @Override
     public void start(BundleContext bundleContext) throws Exception {
@@ -141,7 +144,6 @@ public final class CorePlugin extends Plugin {
 
         // initialize service trackers before the services are created
         this.loggerServiceTracker = createServiceTracker(context, Logger.class);
-        this.publishedGradleVersionsServiceTracker = createServiceTracker(context, PublishedGradleVersionsWrapper.class);
         this.workspaceOperationsServiceTracker = createServiceTracker(context, WorkspaceOperations.class);
         this.internalGradleWorkspaceServiceTracker = createServiceTracker(context, InternalGradleWorkspace.class);
         this.processStreamsProviderServiceTracker = createServiceTracker(context, ProcessStreamsProvider.class);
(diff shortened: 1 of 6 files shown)


org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/util/gradle/PublishedGradleVersionsWrapper.java
```diff
@@ -18,58 +18,60 @@
...
- * {@link PublishedGradleVersions} instance, default version information is provided. This handles,
- * for example, those scenarios where the versions cannot be retrieved because the user is behind a
- * proxy or offline.
+ * Wraps the {@link PublishedGradleVersions} functionality in a background job
+ * that handles all exceptions gracefully. If an exception occurs while calling
+ * the underlying {@link PublishedGradleVersions} instance, default version
+ * information is provided. This handles, for example, those scenarios where the
+ * versions cannot be retrieved because the user is behind a proxy or offline.
  */
+@Component(property = { "service.ranking:Integer=1" }, service = PublishedGradleVersionsWrapper.class)
```
laeubi, line 32: Do you really need a non default service rank here?

vogella: > Do you really need a non default service rank here?

If you look at the activator it also sets this ranking so I just replicated it.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/CorePlugin.java
```diff
@@ -263,7 +262,9 @@ public static Logger logger() {
     }
 
     public static PublishedGradleVersionsWrapper publishedGradleVersions() {
-        return (PublishedGradleVersionsWrapper) getInstance().publishedGradleVersionsServiceTracker.getService();
+    	BundleContext bundleContext = FrameworkUtil.getBundle(CorePlugin.class).getBundleContext();
```
laeubi, line 265: You should really use a servicetracker here see for example here how to handle such legacy cases:
https://github.com/eclipse-m2e/m2e-core/blob/25d6833f9cf78e0c423155c9d8381f0db8eaa9d0/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/MavenPluginActivator.java#L114

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/util/gradle/PublishedGradleVersionsWrapper.java
laeubi: This is a final class, it is very unlikley that there are more than one service and even if they seem to be totaly eqivialent so this is most probably obsolete.

Service ranking is only important if there are multiple service of the same type and you need to have some kind of ordering of them.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/CorePlugin.java
vogella: How is the tracker closed in your example?

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/util/gradle/PublishedGradleVersionsWrapper.java
vogella: > This is a final class, it is very unlikley that there are more than one service and even if they seem to be totaly eqivialent so this is most probably obsolete.
> 
> Service ranking is only important if there are multiple service of the same type and you need to have some kind of ordering of them.

Agree, but changing logic AND implementation is IMHO a bad way. So leaving the priority in should allow @donat to review this easier. And next step could be to remove with a follow-up commit.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/CorePlugin.java
laeubi: All services are automatically freed when the bundle stops, you can of course clear them mnaually if you like also see for a more complete example:

https://github.com/eclipse-platform/eclipse.platform.runtime/pull/28/files

laeubi: Your decisions of course, I usually try to avoid migrating useless/outdated stuff as one has to test this twice then or no one feels to change this later on.

vogella: Thanks, I update the PR with your suggestion.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/CorePlugin.java
```diff
@@ -309,4 +310,20 @@ public static ToolingApiOperationManager operationManager() {
     public static ExtensionManager extensionManager() {
         return getInstance().extensionManager;
     }
+
+    private static <T> T getService(Class<T> service) {
+        BundleContext context = plugin.getBundle().getBundleContext();
```
laeubi, line 315: you might want to check if the plugin !=null

PR #1156, opened 2022-05-06
Fixes oomph test configuration variables

fvclaus:
@donat `setup:VariableTask` was not doing what I expected. I changed it to add a system property to the `eclipse.ini`. After running Help > Perform setup task, Eclipse adds the system properties:

oomph-improvement-1

I also quoted the path to the JDK as you requested in #1154.

### DIFF
--- a/buildship.setup
+++ b/buildship.setup
@@ -52,6 +52,14 @@
         <description>Redirect setup model to check for changes in the local file instead in the url</description>
       </setupTask>
       <setupTask
+        xsi:type="setup:EclipseIniTask"
+        option="-Dbuildship.jre.location-1.8"
+        value="=${jre.location-1.8}"/>
+      <setupTask
+        xsi:type="setup:EclipseIniTask"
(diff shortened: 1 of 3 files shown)


donat (approved):
LGTM

fvclaus:
@donat I just found out by accident that it also possible to set string substitution variables through oomph:

```xml
      </setupTask>
        <setupTask
          xsi:type="setup:StringSubstitutionTask"
          name="git.clone.oomph"
          value="${git.clone.oomph.location}">
        <description>The location of Oomph's Git clone.</description>
      </setupTask>   
```

image

I was originally looking for them, before settling on system properties, because they seemed more appropriate for this use case. Do you want me to replace the system properties with string substitution variables?

donat:
For the JDK locations, I think using a custom system property is better as it will be universally available for all forked processes (i.e. the Buildship integration tests). The variable substitution will not work if a new launch configuration is created, for example when you launch a single test class/method from the editor context menu.

PR #1161, opened 2022-05-11
Support toolingAPI compatibility check

### LINKED ISSUE #1125: Discussion: Java Compatibility Issue report
For different Gradle versions, different range of Java versions are supported. Lots of users may encounter this issue when working on a new repository. It would be very helpful if we can report this issue to users and provide the corresponding ways to fix it.

Now we can find a [Compatibility Matrix](https://docs.gradle.org/current/userguide/compatibility.html) in the Gradle website, this matrix is correct in general, but there are some cases. For example, we can use Java 15 and Gradle 6.3 to configure build successfully but in that matrix the first Gradle version to support Java 15 is 6.7 (Maybe the complete support since 6.7?). So, if we use this matrix for checking in advance, can we revisit it and make it more precise? With the new matrix, we can check the compatibility issue before synchronization and report to users.

Another way is to improve the current **ToolingAPIStatusType**, now this enum type contains several types:
https://github.com/eclipse/buildship/blob/35d01caad3ab023fdd9fdf4a1b9e9c738bff7fd0/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/operation/ToolingApiStatus.java#L36-L45

Maybe we can expand these types and add something like `INCOMPATIBILITY_JAVA` to show that the current Java version is incompatible with Gradle.

CsCherrYY:
Fixes #1125

### Context

This PR enables compatibility check in `GradleBuild.synchronize()`, it will help buildship detect the TAPI compatibility issue in advance and return a specific status. The TAPI compatibility matrix can be found in `java-minimum-supported-toolingAPI.properties`. 

This PR also enables Teamcity CI to generate that properties file automatically, and we have only to specific the test JDKs in `Jdk.kt` and `Project.kt`. Since the matrix of published versions will not change, we can also make this process (generating `java-minimum-supported-toolingAPI.properties`) not a part of CI, but a Gradle task. When we need to add a new JDK version to the matrix we can manually run it and update the properties file, this way might save time when running CI. But the final solution will depend on @donat 's comments.

### What we should do in Teamcity server side
See `Jdk.kt`, we should have JDK 8~17 installed in the Teamcity server.


donat:
Hi! I had an initial look at the implementation and it's really promising. I'll add my full review soon.

FYI I'm currently working on modernizing the Buildship build scripts: https://github.com/eclipse/buildship/pull/1163
Once it's done, the wrapper will use the latest Gradle version. This also means that we can use toolchains, potentially making the cross-jvm test coverage implementation simper.

CsCherrYY:
@donat Sure! And thanks for the information, I'll take a look at #1163 and if there is any simpler way (e.g., use toolchain closure as you mentioned), I'm glad to change the implementation.

donat:
I still have a long way to go until #1163 is finished. The build is ancient and contains a large amount of technical debt.

donat:
@CsCherrYY the wrapper in the project has been updated to 7.4. Can you please check if we can use toolchains in this PR?

CsCherrYY:
@donat I added a simple closure in `org.eclipse.buildship.core.test/build.gradle`:
```
tasks.getByName("crossVersionEclipseTest") {
    javaLauncher = javaToolchains.launcherFor {
        languageVersion = JavaLanguageVersion.of(17)
    }
}
```
and running test via `gradlew crossVersionEclipsetest` in java 11 jvm, but it seems the test result is the same as that without this closure, it shows the compatiiblity of Java 11.

it there anything I'm missing?

donat:
I'm looking into this. Glancing through `buildSrc`, it seems the runtime Eclipse process created by the `eclipseTest` task always uses the same Java version as the build.

donat:
@CsCherrYY There was a bug in the `eclipseTest` implementation. The task always used the same JDK from the running build. 

Now you can pass `-Peclipse.test.java.version=17` to tell the build to launch the test Eclipse with Java 17: https://github.com/eclipse/buildship/blob/master/build.gradle#L966

CsCherrYY:
@donat thanks! I'll rebase and try again.

donat:
@CsCherrYY Cool. Let me know when I should take another look. Also, don't hesitate to reach out if you are stuck, I'm happy to help.

CsCherrYY:
hey @donat , thanks for reminding. I just rebased this branch and adopted the new properties. I did a test on my local machine with the following steps:

1. specify `eclipse.test.java.version` in gradle.properties file
2. execute `gradlew clean crossVersionEclipseTest`, it will successfully update the compatibility info for given JVM version in `src/main/resources/org/eclipse/buildship/core/test/gradle/java-minimum-supported-toolingAPI.properties`

However, I didn't test in the Teamcity CI server, could you please help me verify it?

donat:
Sure, I'll take a look soon.

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/DefaultGradleBuild.java
```diff
@@ -83,6 +128,29 @@ public DefaultGradleBuild(org.eclipse.buildship.core.internal.configuration.Buil
...
+            String javaVersion = javaHome != null ? new StandardVMType().readReleaseVersion(javaHome) : System.getProperty("java.version");
+            if (javaVersion == null) {
+                return DefaultSynchronizationResult.from(ToolingApiStatus.from("Project synchronization", new UnsupportedJavaVersionException("Can't determine Java version when synchronizing project.")));
+            }
+            JavaVersion javaVersionObject = JavaVersion.toVersion(javaVersion);
+            String minGradleVersion = compatibilityMap.get(javaVersionObject.getMajorVersion());
+            if (minGradleVersion != null && GradleVersion.version(gradleVersion).compareTo(GradleVersion.version(minGradleVersion)) < 0) {
+                return DefaultSynchronizationResult.from(ToolingApiStatus.from("Project synchronization", new UnsupportedJavaVersionException(String.format("Can't use Java %s and Gradle %s to synchronize project.", javaVersion, gradleVersion))));
+            }
+        }
```
donat, line 153: 💄 Please extract this to a `validateToolingApiCompatibility` method.

donat (changes requested):
Nice improvements, thanks for the contribution!
I have only one concern about CI feedback. I think if there's a compatibility breakage then the cross-version test should fail.

org.eclipse.buildship.core.test/build.gradle
```diff
@@ -13,3 +15,35 @@ eclipseTest {
     optionsFile rootProject.project(':org.eclipse.buildship.core').file('.options')
 }
 
+def compatibilityCheckClassName = 'org.eclipse.buildship.core.internal.workspace.ImportingProjectCrossVersionTest'
+def compatibilityCheckDescriptorPrefix = 'Can import a simple project build with Gradle '
+
+tasks.getByName("crossVersionEclipseTest") {
```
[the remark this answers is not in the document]
CsCherrYY: You're right, and IMO it depends on when we will update the compatibility matrix.

The existing matrix (for current version, the JVM 8~17) should be strictly "final" and is expected not to be edited any more, since it's for published JDK and Gradle versions. They can be checked for the current CI process as you mentioned.

But for a new JDK version, e.g., we're testing the lowest Gradle supported version for JDK 18,  If we return when a test case fails, how can we find the expected version?

org.eclipse.buildship.core.test/src/main/resources/org/eclipse/buildship/core/test/gradle/java-minimum-supported-toolingAPI.properties
```diff
@@ -0,0 +1,10 @@
+11=4.8
+12=4.8
+13=6.0
+14=6.3
+15=6.3
+16=7.0
+17=7.3
+8=2.9
+9=4.3
+10=4.5
```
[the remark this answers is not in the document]
CsCherrYY: Sure we can, if we have a better check mechanism from https://github.com/eclipse/buildship/pull/1161#discussion_r898800637, we can completely remove this file.

org.eclipse.buildship.core.test/build.gradle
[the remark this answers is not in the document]
CsCherrYY: A quick idea is we can update this matrix when all tests are finished, not per test case (this PR currently uses), in this way, we can choose to add a new property in the matrix or check an existing property. But this change could happen in the CI machine, how to update this matrix is still a problem.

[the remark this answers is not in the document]
donat: IIUC we need to keep the CI up-to-date with the JDK versions. I can also imagine to add an extra cross-version coverage for early access JDK releases.

donat:
@CsCherrYY I've added your changes to the `.teamcity` directory manually: https://github.com/eclipse/buildship/commit/c2de4a09f8a1c6f4777141aa94b6003e487c8a57 (hence the merge conflict)
This way it will be easier to test your changes.

org.eclipse.buildship.core.test/build.gradle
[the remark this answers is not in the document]
CsCherrYY: So I can imagine the workflow is:

- In this PR, we will initialize a compatibility matrix conatins JDK 7~18
- if the CI result does not match the current compatibility matrix, it will report the actual result and the expect result
- When JDK 19 is (early access) available, we will add related JDK info and cross-version coverage, at the first time, the compatibility information of JDK 19 has been generated in CI machine, we need to manually get it and update it to the master branch
- For the following CI runs, it will normally check JDK 7~19

does this make sense?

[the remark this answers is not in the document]
donat: Sorry for not replying earlier. Yes, that makes sense!

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/internal/workspace/ImportingProjectCrossVersionTest.groovy
```diff
@@ -102,6 +103,42 @@ class ImportingProjectCrossVersionTest extends ProjectSynchronizationSpecificati
...
+        when:
+        importAndWait(simpleProjectDir, distribution)
+
+        then:
+        allProjects().size() == 1
+        numOfGradleErrorMarkers == 0
+        getPlatformLogErrors().size() == 0
+
+        where:
+        distribution << getSupportedGradleDistributions()
```
donat, line 141: `getSupportedGradleDistributions()` should be changed to only return Gradle distributions that support the Java version running the test (i.e. `eclipse.test.java.version`). 

This test IMHO should use `getAllGradleDistributions()` (to be implemented), and should assert that
- import with compatible java versions succeed
- import with incompatible java versions fail.

org.eclipse.buildship.core/src/main/resources/org/eclipse/buildship/core/internal/gradle/java-minimum-supported-toolingAPI.properties
```diff
@@ -0,0 +1,10 @@
+11=4.8
```
donat, line 1: It seems to me that the data in this file almost matches the [compatibility matrix in the Gradle documentation](https://docs.gradle.org/current/userguide/compatibility.html). It seems like the docs are in fact a bit more restrictive. 

Shouldn't we start with the values in the docs and then adjust the values based on the test failures from the cross-version coverage? Or the numbers in this file is already a result of that?

org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/DefaultGradleBuild.java
```diff
@@ -83,9 +97,40 @@ public DefaultGradleBuild(org.eclipse.buildship.core.internal.configuration.Buil
...
+            }
+            File javaHome = this.buildConfig.getJavaHome();
+            String javaVersion = javaHome != null ? new StandardVMType().readReleaseVersion(javaHome) : System.getProperty("java.version");
+            if (javaVersion == null) {
+                return ToolingApiStatus.from("Project synchronization", new UnsupportedJavaVersionException("Can't determine Java version when synchronizing project."));
+            }
+            JavaVersion javaVersionObject = JavaVersion.toVersion(javaVersion);
+            String minGradleVersion = compatibilityMap.get(javaVersionObject.getMajorVersion());
+            if (minGradleVersion != null && GradleVersion.version(gradleVersion).compareTo(GradleVersion.version(minGradleVersion)) < 0) {
+                return ToolingApiStatus.from("Project synchronization", new UnsupportedJavaVersionException(String.format("Can't use Java %s and Gradle %s to synchronize project.", javaVersion, gradleVersion)));
```
donat, line 128: 💄 Let's make the exception message more actionable.

```suggestion
                return ToolingApiStatus.from("Project synchronization", new UnsupportedJavaVersionException(String.format("The current build uses Gradle %s running on Java %s which is not supported. Please consult the Gradle documentation to find the compatible combinations: https://docs.gradle.org/current/userguide/compatibility.html.", gradleVersion, javaVersion)));
```

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/internal/workspace/ImportingProjectCrossVersionTest.groovy
```diff
@@ -102,6 +103,54 @@ class ImportingProjectCrossVersionTest extends ProjectSynchronizationSpecificati
...
+        distribution << supportedGradleDistributions
+    }
+
+    @Unroll
+    def "Can not import a simple project build with Gradle #distribution.version"(GradleDistribution distribution) {
+        when:
+        def result = getImportAndWaitResult(simpleProjectDir, distribution)
+
+        then:
+        !result.status.isOK()
```
donat, line 150: Add assertion for the status message to verify a compatibility error.

donat:
> One thing to mention is that it seems not work with JDK 18 so I didn't put it in the compatibility matrix. Maybe related to the eclipse platform?

It's fine, let's worry about this later.

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/internal/test/fixtures/ProjectSynchronizationSpecification.groovy
```diff
@@ -75,6 +75,10 @@ abstract class ProjectSynchronizationSpecification extends WorkspaceSpecificatio
         assert result.status.isOK()
     }
 
+    protected SynchronizationResult getImportAndWaitResult(File location, GradleDistribution gradleDistribution = GradleDistribution.fromBuild(), File javaHome = null) {
+        tryImportAndWait(location, gradleDistribution, javaHome)
+    }
```
donat, line 80: ❌ This method seems redundant; you can just call `tryImportAndWait`.

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/internal/workspace/ImportingProjectCrossVersionTest.groovy
CsCherrYY: currently I just check the property `eclipse.pde.launch` to know if we are in testing. https://github.com/eclipse/buildship/blob/b3ebed60b964469e446af96af1d1d01be0cf9ef7/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/DefaultGradleBuild.java#L108-L109

In this way, we can't know it's a comatibility error. I put this check in order to get the real test result without the compatibility check.

If we just remove this check, we will get the test result under the compability limitation, so we can only test "we can limit the version with the matrix", but we can't test the correctness of the matrix itself.

WDYT?

donat: We have to be able to test both things:
- the Gradle-Java compatibility matrix itself
- the expected Buildship behavior when an incompatible combination is being used.

I think we should introduce a new `org.eclipse.buildship.integtest.bypassToolingApiCompatibilityChecks`system property:
- set the system propert on the `crossVersionEclipseTest` task: https://github.com/eclipse/buildship/pull/1161/files#diff-94f7e44129e98e0c4f448630d29f25ed8117e1554b6603e3f341200e9d6c2398R22
- Adjust the `EclipseTest` task implementation. `jvmArgs.addAll(testTask.getJvmArgs());` is missing here: https://github.com/eclipse/buildship/blob/master/buildSrc/src/main/groovy/eclipsebuild/testing/EclipseTestExecuter.java#L157

donat:
@CsCherrYY let me know if I can help with anything here.

CsCherrYY:
@donat sorry for replying late. I have pushed https://github.com/eclipse/buildship/pull/1161/commits/17ef0b7152fc345509ec109816c6144627588687 to address the comments and add tests to guard the scenarios we mentioned at https://github.com/eclipse/buildship/pull/1161#discussion_r911660408

donat (approved):
LGTM.
I've scheduled a CI build for this PR.

CsCherrYY:
@donat I just rebase this branch to the newest upstream/master and did crossversion tests in my local PC for all Java versions, feel free to trigger a new CI build. 

I just squash the previous changes and leave my newest change in https://github.com/eclipse/buildship/pull/1161/commits/21c4567d48f2598bcd65ef893d61f37d7daba9d2, this change mainly contains:

- move compatibility check to `CompatibilityChecker` to make `DefaultGradleBuild` clear
- guard some scenarios where we can't get environment, see: https://github.com/eclipse/buildship/blob/21c4567d48f2598bcd65ef893d61f37d7daba9d2/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/CompatibilityChecker.java#L55-L57

The only concern is about the current version values in the compatibility properties file. I'm not sure how to check if they are too "strict" or "loose".

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/internal/test/fixtures/GradleVersionConstraints.java
```diff
@@ -78,16 +78,6 @@ public boolean apply(GradleVersion element) {
                 }
             };
         }
-        if (trimmed.startsWith(NOT_EQUALS)) {
```
CsCherrYY, line 81: remove this part since it will ignore https://github.com/eclipse/buildship/blob/117332457a6763c21a8ea7d602ed25b39ba8e92e/org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/internal/test/fixtures/GradleVersionConstraints.java#L86-L94 which can also handle multiple `!=`.

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/internal/test/fixtures/GradleVersionParameterization.java
```diff
@@ -118,22 +120,16 @@ public GradleDistribution apply(GradleVersion input) {
...
      * @param gradleVersionPattern the pattern of Gradle versions to match, must not be null
+     * @param ignoreJavaConstraint whether to ignore the java version constraint or not
      * @return the matching Gradle versions falling into the configured range, never null
      */
-    ImmutableList<GradleVersion> getGradleVersions(String gradleVersionPattern, boolean negateJavaConstraints) {
+    ImmutableList<GradleVersion> getGradleVersions(String gradleVersionPattern, boolean ignoreJavaConstraint) {
         Preconditions.checkNotNull(gradleVersionPattern);
 
         Predicate<GradleVersion> versionConstraint = GradleVersionConstraints.toPredicate(gradleVersionPattern);
         Predicate<GradleVersion> toolingApiConstraint = GradleVersionConstraints.toPredicate(">=2.9");
```
CsCherrYY, line 130: Currently the `toolingApiConstraint` limits version >= 2.9. I tried to change it to 2.6 but the following test cases in Java 8 failed:

- Natures and build commands are updated for Gradle #version
- Custom project output location is not updated for #version

So I didn't change it here thus the versions in unsupported gradle properties file start from 2.9. It would be necessary to check if the version is under the limitation of TAPI before we check the properties file, but I have no idea if we should set the limitation to 2.6 or 2.9.

donat: Sorry for the late follow-up. 

The minimum tested Gradle version should be 2.6 to be in sync with [what Gradle supports](https://github.com/gradle/gradle/blob/de63e27780bef54829e2cd0b96dcc6bca599cb59/subprojects/tooling-api/src/testFixtures/groovy/org/gradle/integtests/tooling/fixture/ToolingApiSpecification.groovy#L65). I don't know why it was set to 2.9 in the past.

I'll take a look at this.

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/internal/test/fixtures/GradleVersionConstraints.java
donat: 👍

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/internal/test/fixtures/GradleVersionParameterization.java
donat: The failure is expected: the DSL for custom eclipse projects/natures was added in Gradle 2.9. I've created a PR to adjust the the cross-version test range: https://github.com/eclipse/buildship/pull/1198

CsCherrYY: Get it. I'll rebase this PR once #1198 got merged.

donat: I've merged #1198

CsCherrYY: I have pushed https://github.com/eclipse/buildship/pull/1161/commits/c2cba1da69822e5898ed810960b16b3ef15b3370 to extend the matrix to include versions from 2.6

donat:
🎉 🎉 🎉

donat:
Thanks, @CsCherrYY for the contribution!

PR #1170, opened 2022-05-18
Restore lib/ folder for core.tests and oomph.test

fvclaus:
@donat , I looked into `BundlePlugin` as you pointed out in https://github.com/eclipse/buildship/pull/1154#issuecomment-1121091278

I did not want to complicate the `BundlePlugin` implementation by making it aware of the shared test dependencies. Which would require reading dependencies between projects and figuring out what is shared. Instead, I took a second look at the duplicate classpath issue of https://github.com/eclipse/buildship/pull/1154 and came up with a better solution: I made the `oomph.test` dependency a fragment host of `oomph` not of `core`.


donat:
So we had the duplicate classpath issue only because of `oomph.test` and `core.test` specified the same fragment host? That makes sense.

donat:
I've tested it locally and looks good. Are you planning to update the lib folder in `ui.test`?

fvclaus:
> So we had the duplicate classpath issue only because of `oomph.test` and `core.test` specified the same fragment host? That makes sense.

Exactly. When I provided the first fix, I thought thats the way its supposed to be. This is way I made up the "testdependencies" project.

> Are you planning to update the lib folder in ui.test?
Done. The `bundled` dependency for Spock was missing in `build.gradle`

donat:
Thank you @fvclaus!

PR #1179, opened 2022-06-15
Fixes usage of reserved keywords in New Project wizard

### LINKED ISSUE #1085: New project wizard should escape keywords in generated package name
### Expected Behavior

Java keywords should be escaped when generating a new project

### Current Behavior

If the project name `gradle-new-project` then the generated project will define an invalid `gradle.new.project` package.

### Context

https://bugs.eclipse.org/bugs/show_bug.cgi?id=573398

"New Gradle Project" wizard generates Library.java and puts in a package whose name comes from splitting the project name by hyphens.

For instance, if the desired project name is "test-new-gradle", the contents of Library.java are the following:

    package test.new.gradle;
    
    public class Library {
        public boolean someLibraryMethod() {
            return true;
        }
    }

"new" is a reserved word.

I expect such invalid java identifiers to be pre-processed by whatever facility Eclipse has for that. In this case, prefixing with an underscore would be enough:

    package test._new.gradle;

I also noticed that leading digits are currently removed from package names. Not sure it's what the users want.

### Steps to Reproduce 

Create a new Gradle project in eclipse with the following name: `gradle-new-project`.

fvclaus:
Fixes #1085

### DIFF
--- a/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/workspace/InitializeNewProjectOperation.java
+++ b/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/workspace/InitializeNewProjectOperation.java
@@ -52,8 +52,7 @@ public class InitializeNewProjectOperation extends BaseToolingApiOperation {
 
     }
 
-    private static void initProjectIfNotExists(BuildConfiguration buildConfig, CancellationTokenSource tokenSource,
-            IProgressMonitor monitor) {
+    private static void initProjectIfNotExists(BuildConfiguration buildConfig, CancellationTokenSource tokenSource, IProgressMonitor monitor) {
         File projectDir = buildConfig.getRootProjectDirectory().getAbsoluteFile();
         if (!projectDir.exists()) {
             if (projectDir.mkdir()) {
--- a/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/workspace/InitializeNewProjectOperation.java
+++ b/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/workspace/InitializeNewProjectOperation.java
@@ -58,12 +58,12 @@ public class InitializeNewProjectOperation extends BaseToolingApiOperation {
         if (!projectDir.exists()) {
             if (projectDir.mkdir()) {
                 InternalGradleBuild gradleBuild = CorePlugin.internalGradleWorkspace().getGradleBuild(buildConfig);
-                RunConfiguration runConfiguration = CorePlugin.configurationManager()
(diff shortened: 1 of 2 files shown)


org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/workspace/InitializeNewProjectOperation.java
```diff
@@ -46,22 +52,40 @@ public void runInToolingApi(CancellationTokenSource tokenSource, IProgressMonito
...
+                RunConfiguration runConfiguration = CorePlugin.configurationManager()
+                        .createDefaultRunConfiguration(buildConfig);
                 GradleProgressAttributes progressAttributes = GradleProgressAttributes.builder(tokenSource, monitor)
-                        .forNonInteractiveBackgroundProcess()
-                        .withFilteredProgress()
-                        .build();
-                gradleBuild.newBuildLauncher(runConfiguration, progressAttributes).forTasks(tasks.toArray(new String[tasks.size()])).run();
+                        .forNonInteractiveBackgroundProcess().withFilteredProgress().build();
+                gradleBuild.newBuildLauncher(runConfiguration, progressAttributes)
+                        .forTasks(createInitJavaLibraryTask(runConfiguration.getProjectConfiguration())).run();
```
donat, line 66: Please restore the formatting.

org.eclipse.buildship.ui.test/src/main/groovy/org/eclipse/buildship/ui/internal/wizard/project/ProjectCreationWizardUiTest.groovy
```diff
@@ -149,18 +171,17 @@ class ProjectCreationWizardUiTest extends SwtBotSpecification {
...
+                        IWorkingSetManager workingSetManager = PlatformUI.getWorkbench().getWorkingSetManager()
+                        IWorkingSet gradleWorkingSet = workingSetManager.getWorkingSet("Gradle")
+                        return gradleWorkingSet.getElements().length > 0
+                    }
+
+                    @Override
+                    public String getFailureMessage() {
+                        return "The Gradle workingset has not been added to the IWorkingSetManager, yet."
+                    }
+                }, 60000)
```
donat, line 185: Same as above

donat (approved):
LGTM.

donat:
The [build](https://builds.gradle.org/buildConfiguration/Tooling_Buildship_Checkpoint_Cross_Version_Test_Coverage_Phase_34/53174543?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true) is green, merging...

donat:
@fvclaus thank you for the contribution!

Issue #1184, opened 2022-07-03
Oomph-based setup fails with exception


FieteO:
### Expected Behavior
I am following the [Setup instructions](https://github.com/eclipse/buildship/blob/master/docs/development/Setup.md) for the Oomph-based setup.

### Current Behavior
The installation via the Eclipse Installer fails with the following exception:
```
Fetching content.xml.xz from https://download.eclipse.org/releases/2022-06/202206151000/
Calculating requirements and dependencies.
Cannot complete the request.  Generating details.
ERROR: org.eclipse.equinox.p2.director code=10053 Cannot complete the install because one or more required items could not be found.
  at org.eclipse.oomph.util.OomphPlugin.coreException(OomphPlugin.java:296)
  at org.eclipse.oomph.p2.internal.core.ProfileTransactionImpl.resolve(ProfileTransactionImpl.java:431)
  at org.eclipse.oomph.p2.internal.core.ProfileTransactionImpl.commit(ProfileTransactionImpl.java:341)
  at org.eclipse.oomph.setup.p2.impl.P2TaskImpl.perform(P2TaskImpl.java:904)
  at org.eclipse.oomph.setup.internal.core.SetupTaskPerformer.doPerformNeededSetupTasks(SetupTaskPerformer.java:3864)
  at org.eclipse.oomph.setup.internal.core.SetupTaskPerformer.performNeededSetupTasks(SetupTaskPerformer.java:3792)
  at org.eclipse.oomph.setup.internal.core.SetupTaskPerformer.performTriggeredSetupTasks(SetupTaskPerformer.java:3773)
  at org.eclipse.oomph.setup.internal.core.SetupTaskPerformer.perform(SetupTaskPerformer.java:3651)
  at org.eclipse.oomph.setup.ui.wizards.ProgressPage$9.run(ProgressPage.java:592)
  at org.eclipse.oomph.setup.ui.wizards.ProgressPage$11$1.run(ProgressPage.java:721)
  at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
  ERROR: org.eclipse.equinox.p2.director code=0 Software being installed: artificial_root 1.0.0.v1656836586776
[...]

PR #1193, opened 2022-07-28
Add Oomph setup compatibility with more than one Eclipse version

fvclaus:
Currently the groovy-eclipse repository targets one specific Eclipse version. If a different Eclipse version is selected in the Eclipse installer, the dependency resolution fails with a cryptic error (something about not being able to resolve artifical root). This has come up in a number of issues. For example #1184.

I don't think it is a good idea to limit the Oomph setup to one specific Eclipse Platform Version.

### DIFF
--- a/buildship.setup
+++ b/buildship.setup
@@ -30,6 +30,12 @@
   <!-- See https://github.com/groovy/groovy-eclipse/wiki#releases -->
   <setupTask
       xsi:type="setup.p2:P2Task"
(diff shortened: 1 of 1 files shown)


fvclaus:
@donat I have also added the snapshot repository to support 2022-09. Without it setting up the project on 2022-09 fails with the same cryptic error message like all the other versions.

donat:
This is an awesome contribution, thanks for doing that!
(Also, sorry for the late reply)

PR #1197, opened 2022-08-11
Allow non-exist source folder when it's optional

### LINKED ISSUE #1196: Allow optional source directory to be added into project classpath
### Expected Behavior
Allow to add optional source folder classpath entry even the folder does not exist on disk.

### Current Behavior
folder that does not exist on disk will be ignored.

### Context

Currently, in https://github.com/eclipse/buildship/blob/35d01caad3ab023fdd9fdf4a1b9e9c738bff7fd0/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/workspace/SourceFolderUpdater.java#L138

The source folder entry will be ignored if it does not exist.

There is one exception when the source folder has the optional attribute:

```
<attribute name="optional" value="true"/>
```

In JDT, optional classpath entry is allowed to be not existed - if I understood the JDT code correctly:

https://github.com/eclipse-jdt/eclipse.jdt.core/blob/f71a39afb62a1436cd094f6274ad9a91b96484f7/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/ClasspathEntry.java#L2219-L2224 

### Steps to Reproduce 

Try to add a non-exist source folder to classpath, sth like:

```groovy
EclipseModel model = project.getExtensions().findByType(EclipseModel)
model.classpath.file.whenMerged { Classpath cp ->
  SourceFolder entry = new SourceFolder("<a path that does not exist on disk>", null)
  entry.entryAttributes.put("optional", "true")
  entry.entryAttributes.put("ignore_optional_problems", "true")
  cp.entries.add(entry)
}
```

### More information

Another example is m2e, for example, if we import the [spring-petclinic](https://github.com/spring-projects/spring-petclinic) as a Maven project, m2e will generate classpath entries like:

```
<classpathentry kind="src" path="target/generated-sources/annotations">
	<attributes>
		<attribute name="optional" value="true"/>
		<attribute name="maven.pomderived" value="true"/>
		<attribute name="ignore_optional_problems" value="true"/>
		<attribute name="m2e-apt" value="true"/>
	</attributes>
</classpathentry>
```

while that generated folder does not exist on disk until build.

jdneo:
Signed-off-by: sheche <sheche@microsoft.com>

Fixes #1196

### DIFF
--- a/org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/internal/workspace/SourceFolderUpdaterTest.groovy
+++ b/org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/internal/workspace/SourceFolderUpdaterTest.groovy
@@ -80,10 +80,7 @@ class SourceFolderUpdaterTest extends WorkspaceSpecification {
 	
 	def "Optional source folders that don't physically exist are allowed"() {
 		given:
-		def newModelSourceFolders = gradleSourceFoldersWithoutAttributes(['src-not-there'])
-		for (EclipseSourceDirectory sd : newModelSourceFolders) {
-			sd.classpathAttributes = ModelUtils.asDomainObjectSet(gradleClasspathAttributes(attributes(["optional" : "true"])))
-		}
+		def newModelSourceFolders = gradleSourceFolders(['src-not-there'], [], [], ["optional" : "true"], null)
 
 		when:
 		SourceFolderUpdater.update(javaProject, newModelSourceFolders, null)
@@ -91,8 +88,8 @@ class SourceFolderUpdaterTest extends WorkspaceSpecification {
 		then:
 		javaProject.rawClasspath.length == 1
 		javaProject.rawClasspath[0].entryKind == IClasspathEntry.CPE_SOURCE
-                javaProject.rawClasspath[0].path.toPortableString() == "/project-name/src-not-there"
-                javaProject.rawClasspath[0].extraAttributes as List == attributes(['optional': 'true']) as List
+        javaProject.rawClasspath[0].path.toPortableString() == "/project-name/src-not-there"
+        javaProject.rawClasspath[0].extraAttributes as List == attributes(['optional': 'true']) as List
 	}
 
     def "Previous source folders are removed if they no longer exist in the Gradle model"() {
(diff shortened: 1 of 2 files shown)


org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/internal/workspace/SourceFolderUpdaterTest.groovy
```diff
@@ -77,6 +77,20 @@ class SourceFolderUpdaterTest extends WorkspaceSpecification {
         then:
         javaProject.rawClasspath.length == 0
     }
+	
+	def "Optional source folders that don't physically exist are allowed."() {
```
donat, line 81: Nit: please remove the trailing dot.

```suggestion
	def "Optional source folders that don't physically exist are allowed"() {
```

org.eclipse.buildship.core.test/src/main/groovy/org/eclipse/buildship/core/internal/workspace/SourceFolderUpdaterTest.groovy
```diff
@@ -77,6 +77,20 @@ class SourceFolderUpdaterTest extends WorkspaceSpecification {
...
+		def newModelSourceFolders = gradleSourceFoldersWithoutAttributes(['src-not-there'])
+		for (EclipseSourceDirectory sd : newModelSourceFolders) {
+			sd.classpathAttributes = ModelUtils.asDomainObjectSet(gradleClasspathAttributes(attributes(["optional" : "true"])))
+		}
+
+		when:
+		SourceFolderUpdater.update(javaProject, newModelSourceFolders, null)
+
+		then:
+		javaProject.rawClasspath.length == 1
```
donat, line 92: Let's make the expectation a bit more concrete.

```suggestion
		javaProject.rawClasspath.length == 1
		javaProject.rawClasspath[0].entryKind == IClasspathEntry.CPE_SOURCE
                javaProject.rawClasspath[0].path.toPortableString() == "/project-name/src-not-there"
                javaProject.rawClasspath[0].extraAttributes as List == attributes(['optional': 'true']) as List
```

donat (changes requested):
Looks good; I only have a couple of comments on the test coverage.

donat:
@jdneo Thank you for the contribution!

jdneo:
@donat Thank you too for reviewing and providing help 🎉

PR #1212, opened 2022-11-26
Save versions cache file in plugin's state location by default

### LINKED ISSUE #1178: Move `.cache\tooling\gradle\versions.json` to a folder in Gradle's or Eclipse's
In my case, the entire `.cache` folder is only used for that single file, and it's annoying to have it there knowing it's only used for that.

I'd expect Buildship to instead use a folder instead inside Gradle's `.gradle` folder or at least in some Eclipse folder. Or maybe the system's temp folder, given the file is only valid for a day.

The file path is decided here:
https://github.com/eclipse/buildship/blob/35d01caad3ab023fdd9fdf4a1b9e9c738bff7fd0/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/util/gradle/PublishedGradleVersions.java#L224

altrisi:
Fixes #1178.

This PR changes the folder where the versions cache is stored to be inside buildship's state location. It keeps the xdgcache thing though and will use it if it's set.

### Context
In my case, the `.cache` folder where it was being saved previously was only used for this one file, and I doubt I'm the only one. Not only that, I think it makes sense for buildship's files to be in buildship's folder, according to [this Eclipse wiki page](https://wiki.eclipse.org/FAQ_Where_do_plug-ins_store_their_state%3F) it's the folder that should be used for "cached information".

Only testing I ran was from the set up Github Actions. Weirdly the default build failed because Gradle was unable to find dependencies for the `buildSrc` project. I changed the repos (in another branch) just for the purposes of running the build, and it then passed.

### DIFF
--- a/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/util/gradle/PublishedGradleVersions.java
+++ b/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/util/gradle/PublishedGradleVersions.java
@@ -37,6 +37,8 @@ import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.google.gson.reflect.TypeToken;
 
+import org.eclipse.buildship.core.internal.CorePlugin;
+
 /**
  * Provides information about the Gradle versions available from services.gradle.org. The version information can optionally be cached on the local file system.
  *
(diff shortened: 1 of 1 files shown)


altrisi:
Do I really need to give my address for this tiny change?

Here's the action run btw: https://github.com/altrisi/buildship/actions/runs/3553609398/jobs/5969212032

altrisi:
Any input on this? It's been over 2 months without response.

donat:
Hi! Sorry for the late response. I'm fine with the changes. The only thing blocking this PR is that you need to sign the Eclipse Contributor Agreement: [eclipse.org/projects/handbook/#resources-github](https://www.eclipse.org/projects/handbook/#resources-github)

donat:
I got no response so I assume the PR is abandoned. I still think there's value here, so I implemented my variant. 

Superseded by https://github.com/eclipse/buildship/pull/1312

altrisi:
Completely missed that comment sorry, but thanks for considering and implementing the change!

Issue #356, opened 2017-02-08
Provide smart import support


donat:
Original post: https://bugs.eclipse.org/bugs/show_bug.cgi?id=469636

armandomejiam:
+1

donat:
A few tips on how to implement smart import support is provided by @mickaelistria at #751.

donat:
Another request for smart import in Bugzilla: https://bugs.eclipse.org/bugs/show_bug.cgi?id=549760

thokuest:
Yet another request for smart import in Bugzilla: https://bugs.eclipse.org/bugs/show_bug.cgi?id=559074

maxandersen:
I opened https://bugs.eclipse.org/bugs/show_bug.cgi?id=559074 as https://projects.eclipse.org/projects/tools.buildship/developer points to bugzilla :)

But since then I found this issue is open for smart import for gradle.

I've read through the various comments I honestly don't follow what the remaining blockers are for providing a basic 'if build.gradle file available, mark it as a gradle project and trigger import" feature ?  vscode-java support does auto-import and it seem to work quite well (maybe @fbricon can comment on how they did it).

it is literally the one thing that makes eclipse gradle experience stand out from from vscode and intellij that with both of those plain eclipse, maven and gradle projects "just opens" but for eclipse it is everything but Gradle that is handled.

Would be great to see if we could get this enabled for gradle - willing to submit patches/test it out if can get an update on what are the knowns blocker if any.

donat:
@maxandersen 

> projects.eclipse.org/projects/tools.buildship/developer points to bugzilla :)

Unfortunately, there's no way to configure the bug tracker on the project page. 

> I've read through the various comments I honestly don't follow what the remaining blockers are for providing a basic 'if build.gradle file available, mark it as a gradle project and trigger import. 

I think there should not be any blockers to implement this feature, at least not big ones. We simply had bigger issues/missing features at our hands to work on. 

> Would be great to see if we could get this enabled for gradle - willing to submit patches/test it out if can get an update on what are the knowns blocker if any.

Awesome, thanks for volunteering. Well, for started I'd recommend reading through the [initial conversation](https://bugs.eclipse.org/bugs/show_bug.cgi?id=469636) on the topic why we didn't implement this feature right away. I think the problems are still valid now: the importer works per project folder whereas Gradle can import one build (containing multiple projects) at a time.

[...]

PR #1219, opened 2023-01-05
Add GradleProjectConfigurator to support smart-import

laeubi:
Fixes https://github.com/eclipse/buildship/issues/356

### Context
Currently gradle projects are not detected when using smart import. This adds the necessary `ProjectConfigurator` to support importing gradle projects this way.

### DIFF
--- a/org.eclipse.buildship.ui/plugin.xml
+++ b/org.eclipse.buildship.ui/plugin.xml
@@ -650,7 +650,7 @@
        <activeWhen>
            <or>
                <hasFileRecursively filename="gradle.properties"/>
-               <hasFileRecursively filename="settings.gradles"/>
+               <hasFileRecursively filename="settings.gradle"/>
                <hasFileRecursively filename="settings.gradle.kts"/>
            </or>
        </activeWhen>
--- a/org.eclipse.buildship.ui/plugin.xml
+++ b/org.eclipse.buildship.ui/plugin.xml
@@ -642,4 +642,18 @@
           </themeid>
        </stylesheet>
     </extension>
+    <extension
+          point="org.eclipse.ui.ide.projectConfigurators">
+       <projectConfigurator
+               class="org.eclipse.buildship.ui.internal.project.GradleProjectConfigurator"
+               label="Gradle">
+       <activeWhen>
+           <or>
+               <hasFileRecursively filename="gradle.properties"/>
+               <hasFileRecursively filename="settings.gradles"/>
+               <hasFileRecursively filename="settings.gradle.kts"/>
+           </or>
+       </activeWhen>
+       </projectConfigurator>
+    </extension>
 </plugin>
--- a/org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/internal/project/GradleProjectConfigurator.java
+++ b/org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/internal/project/GradleProjectConfigurator.java
@@ -1,0 +1,104 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Gradle Inc. and others.
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ ******************************************************************************/
+package org.eclipse.buildship.ui.internal.project;
+
+import java.io.File;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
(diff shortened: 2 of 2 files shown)


laeubi:
Build complain about 

>  Line does not match expected header line of '^ \* Copyright \(c\) \d\d\d\d Gradle Inc\.$'. [RegexpHeader]

but I do not have anything to do with `Gradle Inc` so why should they hold the copyright of my file?

laeubi:
@donat can you probably take a look at this? This will really improve user experience and preventing false positive detection in m2eclipse.

donat:
Sorry for the late reply. We'll take a look shortly.

reinsch82:
@laeubi Would you mind rebasing your PR and changing the copyright line to 

  Gradle Inc. and others
  
This seems to be the recommended way of leaving the Copyright with it's original creators.

Feel free to add specific copyright notices somewhere down in the file, this was done in the past too.

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/internal/project/GradleProjectConfigurator.java
```diff
@@ -0,0 +1,101 @@
...
+    if(!gradleRoots.isEmpty()) {
+      for(File file : nonGradleRoots) {
+        proposals.remove(file);
+      }
+    }
+  }
+
+  private boolean isGradleRoot(File root) {
+    if(root != null) {
+      if(new File(root, "gradle.properties").isFile()) {
```
reinsch82, line 65: its recommend best practice to use a [`settings` file](https://docs.gradle.org/current/userguide/organizing_gradle_projects.html#sec:settings_file). 
`gradle.properties` is not required, although it is a reasonable indicator for a project.
I would suggest to at least also check for `settings.gradle` and `settings.gradle.kts`.

laeubi:
Done.

org.eclipse.buildship.ui/plugin.xml
```diff
@@ -642,4 +642,14 @@
           </themeid>
        </stylesheet>
     </extension>
+    <extension
+          point="org.eclipse.ui.ide.projectConfigurators">
+       <projectConfigurator
+		        class="org.eclipse.buildship.ui.internal.project.GradleProjectConfigurator"
+		        label="Gradle">
+			<activeWhen>
+				<hasFileRecursively filename="gradle.properties"/>
```
reinsch82, line 651: see comment on `isGradleRoot`

reinsch82 (changes requested):
Thanks for your contribution

reinsch82:
@laeubi you'll need to rebase again shortly, we had to change our copyrights too to match the new rule.
For some reason the PR would let it through
thx for your help

laeubi:
@reinsch82 I have adjusted the "marker files", also feel free to enhance this as needed afterwards, this is more a first start to make it work, but you as the gradel experts might have some ideas how to further improve that, I'm only a (seldom) gradle user so most of the time :-)

org.eclipse.buildship.ui/plugin.xml
```diff
@@ -642,4 +642,18 @@
...
     </extension>
+    <extension
+          point="org.eclipse.ui.ide.projectConfigurators">
+       <projectConfigurator
+               class="org.eclipse.buildship.ui.internal.project.GradleProjectConfigurator"
+               label="Gradle">
+       <activeWhen>
+           <or>
+               <hasFileRecursively filename="gradle.properties"/>
+               <hasFileRecursively filename="settings.gradles"/>
```
donat, line 653: ❌ 

```suggestion
               <hasFileRecursively filename="settings.gradle"/>
```

donat (approved):

org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/internal/project/GradleProjectConfigurator.java
```diff
@@ -0,0 +1,104 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Gradle Inc. and others.
```
donat, line 2:
```suggestion
 * Copyright (c) 2019 Gradle Inc. and others
```

donat (approved):

donat:
Thank you for the contribution

jabby:
Thank you @laeubi

laeubi:
Great to see this integrated, will this be part of the next Eclipse Release (2023-06) or are there usually release in between?

donat:
We're planning to do a release in the upcoming weeks, so you can expect your changes to appear in Eclipse 2023-06 by default.

PR #1259, opened 2023-08-23
Support of Kotlin DSL files [google summer of code]

D0zee:
This PR adds support for `.build.kts` files inside the Eclipse IDE. This plugin is a wrapper that checks whether the user has the required Java version and launches the KTS language server.

**About implementation of language server.**
 [Kotlin-language-server](https://github.com/fwcd/kotlin-language-server) was used as a basis. My fixes and modifications are located [here](https://github.com/D0zee/Eclipse-Gradle-KTS-editing/tree/language-server). The following work has been done:
 - Dependencies and implicit imports of the project are calculated by invoking the Gradle Tooling API instead of custom build scripts.
- If the build configuration is broken, the user receives a detailed error in all opened Kotlin build files, and subsequent compilation is impossible.
- The compilation system has been completely changed.
- Indexing and searching all types of files except `.gradle.kts` have been removed to increase the performance of the language server.
- Included builds are supported via a modified version of [CompositeModelQuery](https://github.com/eclipse/buildship/blob/master/org.eclipse.buildship.compat/src/main/java/org/eclipse/buildship/core/internal/workspace/CompositeModelQuery.java). 

**How Language Server Works?**

Each Kotlin build file has its own compilation environment to provide accurate diagnostics to the user. I decided to create a `Map< File, Set<Path> >` where the set of paths represents the classpath. This is necessary because if a user adds plugins, the classpath changes, and therefore the language server must update the compilation environment.

An important point to note is that the `updating of the compilation environment happens when the user saves the file.` In this scenario, the server checks two invariants to avoid unnecessary Tooling API (TAPI) calls:
- If the build configuration contains an error (TAPI call failed during initialization or the last attempt to update the compilation environment), the server updates ALL compilation environments.
- If the plugin block has been modified (the user added or removed some plugins), the server updates the compilation environment only for the current build file.

The Tooling API is invoked on each project and includes compilation information about subprojects and included builds. The language server decides whether to make the call based on two criteria:
1) If a subproject has a parent or is the root, the language server makes a TAPI call on the root project.
2) if 1st statement is false, but the directory contains `settings.gradle.kts` then language server invokes TAPI on this subproject.

**How are Errors Processed?**
If an error occurs during the initialization stage of the language server, the user receives a detailed description of the issue on each opened build file. Compilation doesn't proceed until the user fixes the errors and saves the files with problems. To obtain the most up-to-date and accurate diagnostics, the user must re-lint the file by either saving or editing it.

### DIFF
--- a/org.eclipse.buildship.kotlindsl.provider/.classpath
+++ b/org.eclipse.buildship.kotlindsl.provider/.classpath
@@ -1,7 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <classpath>
-	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER"/>
+	<classpathentry kind="src" path="src/main/java"/>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER">
+		<attributes>
+			<attribute name="module" value="true"/>
+		</attributes>
+	</classpathentry>
 	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
-	<classpathentry kind="src" path="src"/>
 	<classpathentry kind="output" path="bin"/>
 </classpath>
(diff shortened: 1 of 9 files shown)


donat:
Excellent contribution, thank you @D0zee for making this happen!

PR #1284, opened 2023-12-17
Fix editor bindings for gradle.properties and Kotlin DSL files

oleosterhagen:
In order for the editor content type bindings configured in the `plugin.xml` to work correctly, the missing dependencies for the _Generic Text Editor_ have to be added.

For Gradle Properties the content type configuration should use the full file name `gradle.properties` without a separate entry for the file extension, so that these files are opened with the _Generic Text Editor_ and other property files are still handled by the _Properties File Editor_.


donat:
I've tested your changes and does fix the editor association issue. Also, I'd like to mention that there was a similar problems with .gradle.kts files, which this PR (maybe unintentionally) fixed.

donat:

Issue #1275, opened 2023-10-20
deadlock during startup


martinlippert:
A colleague reported a problem during startup, the thread dump reveals a deadlock with Buildship involved:

```
Found one Java-level deadlock:
=============================
"Worker-3: Load persistent model for all projects":
  waiting for ownable synchronizer 0x000000070072dac8, (a java.util.concurrent.locks.ReentrantLock$NonfairSync),
  which is held by "Worker-4: Refreshing workspace"

"Worker-4: Refreshing workspace":
  waiting to lock monitor 0x00006000000568a0 (object 0x000000073ba67d30, a java.lang.Object),
  which is held by "Worker-3: Load persistent model for all projects"

Java stack information for the threads listed above:
===================================================
"Worker-3: Load persistent model for all projects":
	at jdk.internal.misc.Unsafe.park(java.base@21/Native Method)
	- parking to wait for  <0x000000070072dac8> (a java.util.concurrent.locks.ReentrantLock$NonfairSync)
	at java.util.concurrent.locks.LockSupport.parkNanos(java.base@21/LockSupport.java:269)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquire(java.base@21/AbstractQueuedSynchronizer.java:756)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer.tryAcquireNanos(java.base@21/AbstractQueuedSynchronizer.java:1038)
	at java.util.concurrent.locks.ReentrantLock$Sync.tryLockNanos(java.base@21/ReentrantLock.java:168)
	at java.util.concurrent.locks.ReentrantLock.tryLock(java.base@21/ReentrantLock.java:479)
	at org.eclipse.osgi.container.Module.lockStateChange(Module.java:309)
	at org.eclipse.osgi.container.Module.start(Module.java:419)
	at org.eclipse.osgi.framework.util.SecureAction.start(SecureAction.java:513)
	at org.eclipse.osgi.internal.hooks.EclipseLazyStarter.postFindLocalClass(EclipseLazyStarter.java:117)
	at org.eclipse.osgi.internal.loader.classpath.ClasspathManager.findLocalClass(ClasspathManager.java:570)
	at org.eclipse.osgi.internal.loader.ModuleClassLoader.findLocalClass(ModuleClassLoader.java:335)
	at org.eclipse.osgi.internal.loader.BundleLoader.findLocalClass(BundleLoader.java:397)
[...]

PR #1289, opened 2024-01-03
Move model creation outside of synchronized block

reinsch82:
Fixes https://github.com/eclipse/buildship/issues/1275

The deadlock seems to happen because `DefaultModelPersistence.this.lock` is still locked when `PersistentModelConverter.toModel` is called. Since the purpose of the lock is to synchronise access to the content of the property file this is not necessary.
This change moves the call to `PersistentModelConverter.toModel` outside of the lock and should therefore avoid the deadlock situation.


donat (approved):

donat:
The changes [passed](https://builds.gradle.org/buildConfiguration/Tooling_Buildship_Checkpoint_Full_Test_Coverage_Fork_Phase_33/76454332?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true&expandBuildDeploymentsSection=false) CI; merging.

PR #1293, opened 2024-01-14
Prevent execution of the Java Resource Import Configurator #1288

### LINKED ISSUE #1288: Importing java-library builds via "Projects from Folder or Archive" does not work correctly
### Expected Behavior
The Java source folders `src/main/java` and `src/test/java` from subproject `lib` are imported only once.

### Current Behavior
The source folders are imported two times:
1. As source folders of subproject `lib` (this is right)
2. As source folders of the root project (this is wrong)

The incorrectly created folders are selected in the following screenshot:

grafik

### Context
Only the import with automatic project detection (introduced with https://github.com/eclipse/buildship/pull/1219) is affected. Import via _Gradle > Existing Gradle Project_ works correctly.

### Steps to Reproduce 
1. Generate a _java-library_ build: `~/gradle-8.5/bin/gradle init --type java-library`
4. _File > Import: General > Projects from Folder or Archive_
5. Enter _Import source_ directory and press _Finish_

### Your Environment
Platform: Linux
Eclipse: 2023-12
Buildship: 3.1.8

oleosterhagen:
Fixes #1288

When importing Gradle projects from a folder with the generic project configurator, other configurators like the `ProjectWithJavaResourcesImportConfigurator` must not run. Otherwise resources in subprojects could be imported twice.

To prevent the execution of secondary configurators the method `shouldBeAnEclipseProject(...)` has to return `true` for the root of Gradle projects.

Note: In the implementation of the integration test I have to wait for the `SmartImportJob` to finish. Unfortunately the job family of this job is the non-API job class. This is why I used `Class.forName(...)`.

### DIFF
--- a/org.eclipse.buildship.ui.test/Launch Buildship UI tests.launch
+++ b/org.eclipse.buildship.ui.test/Launch Buildship UI tests.launch
@@ -26,7 +26,7 @@
     <stringAttribute key="org.eclipse.jdt.junit.CONTAINER" value="=org.eclipse.buildship.ui.test/src\/main\/groovy"/>
     <booleanAttribute key="org.eclipse.jdt.junit.KEEPRUNNING_ATTR" value="false"/>
     <stringAttribute key="org.eclipse.jdt.junit.TESTNAME" value=""/>
-    <stringAttribute key="org.eclipse.jdt.junit.TEST_KIND" value="org.eclipse.jdt.junit.loader.junit4"/>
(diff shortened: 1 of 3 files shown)


donat (approved):
Reviewed and tested; LGTM!

oleosterhagen:
Thank you!

donat:
@oleosterhagen I've mentioned you in the 3.1.9 [release notes](https://discuss.gradle.org/t/buildship-3-1-9-is-now-available/47442). Once again, thanks for all the work! We do appreciate it.

Issue #1294, opened 2024-01-17
Marker Property value is too long


jdneo:
When import failure happens, Buildship will add a marker with the error stacktrace: https://github.com/eclipse/buildship/blob/455bf33f32d133f7028377d1b000e044194b0bd4/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/marker/GradleErrorMarker.java#L62

Sometimes, the stacktrace could be extremely long depending on the actual error. This will cause an exception when Eclipse checking the marker property, see:

https://github.com/eclipse-platform/eclipse.platform/blob/97d555a8b563dcb3a32bd43ad58ba452fa027a73/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/MarkerInfo.java#L56-L60

Unfortunately, Eclipse does not expose this magic number, so I don't think hardcode the length check at Buildship side will be a good idea. What about only append the stacktrace from the root cause exception to the marker?

PR #1295, opened 2024-01-29
Trim long marker properties

donat:
Fixes https://github.com/eclipse/buildship/issues/1294


jdneo (approved):
LGTM. I built a Buildship locally with this change and not observe the marker property too long error on a project which has that error before.

reinsch82 (approved):
LGTM.
good catch on the string length vs byte array length!

donat:
I merely followed the pattern from https://github.com/eclipse-platform/eclipse.platform/blob/97d555a8b563dcb3a32bd43ad58ba452fa027a73/resources/bundles/org.eclipse.core.resources/src/org/eclipse/core/internal/resources/MarkerInfo.java#L56-L60. Anyway, thanks!

PR #1317, opened 2024-09-04
Github Actions build with multiple OSes

ribafish:
This set up Github Actions build with multiple OSes. An example of how this would look like can be seen [here](https://github.com/ribafish/buildship/actions/runs/10699868215).

This also updates to latest versions of actions used and added the `setup-gradle` action. To get buildScan publishing a secret (`DEVELOCITY_ACCESS_KEY`) will need to be added to this repository.


donat:
LGTM

PR #1326, opened 2025-03-08
Don't set focus on preference page creation

### LINKED ISSUE #1204: GradleWorkbenchPreferencePage steals focus when first created
2022-09 RC2

While navigating the preference page tree by keyboard, when the GradleWorkbenchPreferencePage is first created, focus moves from the tree into the page's "Gradle distribution" group. It should not do that.

nitind:
Fixes #1204, and properly refers to BREE JRE


org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/internal/util/widget/GradleDistributionGroup.java
```diff
@@ -214,7 +214,6 @@ private void updateSelectedState(Button selectedRadioButton, List<Button> allRad
         for (Button radioButton : allRadioButtons) {
             if (radioButton.equals(selectedRadioButton)) {
                 radioButton.setSelection(true);
-                radioButton.setFocus();
```
donat, line 217: 👍

org.eclipse.buildship.ui/.classpath
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <classpath>
-	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER"/>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-11"/>
```
donat, line 3: The intent here with not specifying the version was to allow the active target definition to select the Java version. That way, we can always select the lowest required java version for a given Eclipse platform version. 

❌ This change should not be part of this PR.

donat (changes requested):
Thank you for your contribution! 
I have one request before I can merge your changes. I'd be happy discuss that separately.

org.eclipse.buildship.ui/.classpath
nitind: I disagree with the reasoning, but I've removed this change.

donat (approved):

donat:
@nitind Thank you for the contribution!

PR #1327, opened 2025-03-11
Configured the correct secret in GHA for DV key

ribafish:
Configured the correct secret in GHA for DV key

Fixes build scans not publishing from Github Actions workflow.


ribafish:
@donat ☝️

PR #1338, opened 2025-05-04
Update and simplify the buildship.setup

merks:
- Given the migration to the eclipse-buildship GitHub organization, the clone URI and the redirection URIs should be updated accordingly.
- Install a newer version of Groovy and just list a bunch of possible repositories because the version with an applicable/matching fragment will be used.
- Update the target platform to the most recent one in the repository

### Context

The setup only worked by installing a very old version of Eclipse.

### DIFF
--- a/buildship.setup
+++ b/buildship.setup
@@ -20,66 +20,17 @@
     <requirement
         name="org.codehaus.groovy.eclipse.feature.feature.group"/>
     <requirement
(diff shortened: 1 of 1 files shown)


donat (approved):

donat:

merks:
I've changed the URI in the catalog now too.  I will retest soon...

merks:
FYI, I tested the setup today and it's working well, including the redirection to the local clone's setup resource.

PR #1339, opened 2025-05-04
Include about.html in org.eclipse.buildship.compat

merks:
The required license file is missing which I noticed analyzing the SimRel content.

---

I would suggest checking all your bundles that this file is included and in mentioned in the build.properties too.


donat (approved):

donat:
@merks Thanks for the contribution!

PR #1345, opened 2025-06-28
Ensure minimum width for tree columns

### LINKED ISSUE #1322: Gradle Task Headings Disappear after Update
When I did the update for the STS in my existing workspace the Gradle task headings Disappear

image

image

Please Also note that this is not occurring in new Workspace. Only on existing workspace. Even on existing workspace, when I delete all my projects and re-import them in they work fine.

# System Details Report
---

## Hardware Information:
- **Hardware Model:**                              Hewlett-Packard HP 240 14 inch G9 Notebook PC
- **Memory:**                                      32.0 GiB
- **Processor:**                                   12th Gen Intel® Core™ i5-1235U × 12
- **Graphics:**                                    Intel® Graphics (ADL GT2)
- **Disk Capacity:**                               512.1 GB

## Software Information:
- **Firmware Version:**                            F.05
- **OS Name:**                                     Ubuntu 24.04.1 LTS
- **OS Build:**                                    (null)
- **OS Type:**                                     64-bit
- **GNOME Version:**                               46
- **Windowing System:**                            X11
- **Kernel Version:**                              Linux 6.8.0-51-generic

oleosterhagen:
When restoring the width of a tree column from the preferences, ensure a minimum width of 1, so that the column is not completely hidden and can be resized by the user.

Fixes #1322

### DIFF
--- a/org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/internal/util/layout/LayoutUtils.java
+++ b/org.eclipse.buildship.ui/src/main/java/org/eclipse/buildship/ui/internal/util/layout/LayoutUtils.java
@@ -31,4 +31,16 @@ public final class LayoutUtils {
         return layout;
     }
 
+    /**
+     * When restoring the width of a tree column from the preferences, ensure a
+     * minimum width of 1, so that the column is not completely hidden and can be
+     * resized by the user.
+     * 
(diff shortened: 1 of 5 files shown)


donat:
@oleosterhagen thank you for the contribution!

PR #1356, opened 2025-11-25
Let TAPI connection carry contextual system properties

eskatos:
`buildship.active=true` for all TAPI connections
`buildship.sync.active=true` for IDE Synchronization

Fixes #?

### Context

Just like IntelliJ based IDE set `idea.active` and `idea.sync.active` system properties on the TAPI connection, this PR let Buildship do the same.


donat (approved):
The code looks good, only  a request for clarification.

donat (approved):

donat:
LGTM!

PR #1361, opened 2026-01-21
remove project search on isNeeded in customGradleImport

### LINKED ISSUE #1319: Exception thrown using Gradle Import via oomph for a fresh install requiring a clone
I have created an Oomph product and project to install Eclipse and configure it as required.

For a user doing a fresh install of this oomph config, the STARTUP tasks should firstly clone the git repository to a given location, and once cloned, the `Gradle Import` task should import all projects in that location.

### Expected Behavior
After initial install of the product, Oomph setup tasks begins shortly after Eclipse starts up, and works through all the STARTUP tasks, which includes cloning the Git repo to the machine because its not already there, then Gradle Imports the projects within that cloned repo.

### Current Behavior
The Oomph setup tasks dialog never appears. There is an error in the Eclipse .log file:

```
!ENTRY org.eclipse.oomph.setup.ui 4 0 2024-10-01 16:33:35.753
!MESSAGE Failed to compare project paths for equality
!STACK 0
java.io.UncheckedIOException: Failed to compare project paths for equality
	at org.eclipse.buildship.oomph.impl.CustomGradleImportTaskImpl.calculateProjectsToImport(CustomGradleImportTaskImpl.java:104)
	at org.eclipse.buildship.oomph.impl.CustomGradleImportTaskImpl.isNeeded(CustomGradleImportTaskImpl.java:54)
	at org.eclipse.oomph.setup.internal.core.SetupTaskPerformer.initNeededSetupTasks(SetupTaskPerformer.java:2285)
	at org.eclipse.oomph.setup.ui.SetupUIPlugin.performStartup(SetupUIPlugin.java:811)
	at org.eclipse.oomph.setup.ui.SetupUIPlugin$1$1.run(SetupUIPlugin.java:266)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
Caused by: java.nio.file.NoSuchFileException: /Users/marklawrence/gitRepos/<org>/<repo name>
	at java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:92)
	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)
	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:116)
	at java.base/sun.nio.fs.UnixFileSystemProvider.isSameFile(UnixFileSystemProvider.java:343)
	at java.base/java.nio.file.Files.isSameFile(Files.java:1531)
	at org.eclipse.buildship.oomph.impl.CustomGradleImportTaskImpl.calculateProjectsToImport(CustomGradleImportTaskImpl.java:99)
	... 5 more
```

This is clearly happening because `isNeeded()` is looking for projects to import to decide if it needs to run, however it is looking at a folder location that doesn't exist yet as the Git clone task hasn't ran yet. Could it assume that it 'is needed' if no folder is currently found?

### Context
I can't really use the gradle import task as part of STARTUP tasks after first install, combined with needing to clone the repository containing the projects first.

I've tried using the ResourceCreationTask to create a dummy file in the location, which does cause the Oomph setup tasks dialog to appear, however the Git clone task then fails with an error as it finds a folder it doesn't recognise as a git repository.

### Steps to Reproduce 
Self-explanatory

### Your Environment
Eclipse 4.23
Installing buildship oomph plugins from https://download.eclipse.org/buildship/updates/e411/releases/3.x/3.1.5.v20210113-0909/

Mark-J-Lawrence:
Fixes #1319 

### Context

When installing a fresh Eclipse install via Oomph, users will likely "git clone" then afterwards "gradle import" as part of the `STARTUP` stage of the install.

`isNeeded()` is called as part of a preliminary check before running any `STARTUP` tasks. The `isNeeded` for the Gradle Import task is looking for projects to import on the filesystem before continuing, however, this is always going to fail when its a fresh install, because the repository containing the projects hasn't been cloned yet - oomph will clone as part of the STARTUP tasks, then it gradle imports directly afterwards.

Is there any need to even do this check in `isNeeded`? I don't think so. If this task is configured to be part of a STARTUP task, then its almost always needed, we don't need to check.
the `perform` task re-runs `calculateProjectsToImport` so it will either fail as it currently does at this point later if it cannot still find the project post-clone, or an empty list is returned, meaning it gracefully doesn't import anything

### DIFF
--- a/org.eclipse.buildship.oomph/src/main/java/org/eclipse/buildship/oomph/impl/CustomGradleImportTaskImpl.java
+++ b/org.eclipse.buildship.oomph/src/main/java/org/eclipse/buildship/oomph/impl/CustomGradleImportTaskImpl.java
@@ -50,9 +50,6 @@ public class CustomGradleImportTaskImpl extends GradleImportTaskImpl {
         if (sourceLocators.isEmpty()) {
             return false;
         }
-        if (context.getTrigger() != Trigger.MANUAL) {
-            return !calculateProjectsToImport().isEmpty();
-        }
         return true;
     }
 


Mark-J-Lawrence:
@fraenkelc - I don't suppose you know any tricks to get this working seeing you wrote it and provided examples? - how do you use `Gradle Import` in the `STARTUP` so that it immediately imports projects after cloning the repo containing the projects?

fraenkelc:
I haven't been using eclipse or buildship / oomph for years now. I don't recall the details of the implementation of the task.
Regarding the sample: After updating the git repo the example references it still works for me:
* I downloaded and started the current eclipse installer
* picked "Eclipse IDE for Java Developers" on jre 21
* added my adjusted setup file as a user project
[BuildshipImportTaskExample.setup.xml](https://github.com/user-attachments/files/24780321/BuildshipImportTaskExample.setup.xml) (rename to *.setup)
  image
* The installer created and launched product
* The git clone and import automatically run on start  
  image

Mark-J-Lawrence:
Thank you for clarifying on this @fraenkelc and proving it works.

I think the problem likely most lies with the amount of customising we do in our Oomph projects - we don't place the git repos we clone into the workspace folder, we keep them outside of the Eclipse install, so that if the IDE needs reinstalling, any in-progress work won't be affected if we deleted all the folders etc.

On looking at the code again, it looks like its designed to expect the user to clone into the workspace - so its our niche problem, and I've since managed to work around it.

I'll close this PR for this reason.

